PR #8, opened 2015-01-01
Fixes for Linux and various other small fixes

Hiradur:
With these changes RoR compiles fine with GCC 4.8.2 and Clang 3.4.


only-a-ptr:
Excellent work, thanks a lot.

Issue #10, opened 2015-01-06
Assertion fails for debug build (Linux)


Hiradur:
RoR: rigsofrods-next-stable/source/main/utils/Singleton.h:83: static T* RoRSingletonNoCreation<T>::getSingletonPtr() [with T = GUI_MainMenu]: Assertion `_instance' failed.

Hiradur:
Cause: Console::resize() requests GUI_MainMenu before it got instantiated. 
Solution: Initialize GUI_MainMenu before Console.

Issue #9, opened 2015-01-06
Segfault when entering any vehicle (Linux)


Hiradur:
snippet from RoR.log:

== Validating vehicle: Bus RVI Agora S
== Validating done OK
== Spawning vehicle: Bus RVI Agora S
ResourceTrueTypeFont 'DefaultDashFont' using texture size 256 x 256
ResourceTrueTypeFont 'DefaultDashFont' using real height 19 pixels
Freetype returned nullptr for character 1025 in font DefaultDashFont
Freetype returned nullptr for character 1040 in font DefaultDashFont
Freetype returned nullptr for character 1041 in font DefaultDashFont
Freetype returned nullptr for character 1042 in font DefaultDashFont
Freetype returned nullptr for character 1043 in font DefaultDashFont
Freetype returned nullptr for character 1044 in font DefaultDashFont

\<snip\>

Freetype returned nullptr for character 1101 in font DefaultDashFont
Freetype returned nullptr for character 1102 in font DefaultDashFont
Freetype returned nullptr for character 1103 in font DefaultDashFont
Freetype returned nullptr for character 1105 in font DefaultDashFont
Freetype returned nullptr for character 8470 in font DefaultDashFont
Segmentation fault (core dumped)

Hiradur:
It's coming from RigSpawner.cpp line 6741:

```
node.id_str = def.id.Str();

(gdb) print def.id
$3 = {static INVALID_ID_VALUE = 4294967295, m_id_num = 0, m_id_str = {static npos = <optimized out>, _M_dataplus = {<std::allocator<char>> = {<__gnu_cxx::new_allocator<char>> = {<No data fields>}, <No data fields>}, 
      _M_p = 0xa99618 <std::string::_Rep::_S_empty_rep_storage@@GLIBCXX_3.4+24> ""}}}
```

Aperion:
get rid of node.id_str, it is not used. the problem is using the string with the zero allocator, it causes the internal string to point to null.

PR #12, opened 2015-01-09
Linux fixes

Aperion:
Ffixes for issue #9 and #10.


only-a-ptr:
Thanks for the contributions, Aperion.
The node_t::id_str is actually my leftover debug code. Thanks for cleaning it up.

Aperion:
@hiradur not sure, perhaps it has to do with the ogre implementation or the zero allocator ?

PR #23, opened 2015-01-10
2 small fixes

Hiradur:


Hiradur:
Added a commit that fixes linking errors on Windows that Max98 reported.

Hiradur:
Actually, now that I think about it, it should be _MSC_VER instead of _WIN32 because it's probably specific to Visual Studio, isn't it?

Max98:
I don't know, i might need to setup an other Windows IDE to test it out.

Hiradur:
Well, at least MinGW should work with it.

only-a-ptr:
Just a nitpick - The pullrequest is titled "fixes", but the boat GUI commit isn't a fix by any means :-P

Compilation: _WIN32 is preferable to _MSC_VER, because it's intended for the purpose of platform recognition, but _MSC_VER is intended for MSVC version identification.
http://stackoverflow.com/a/662543
http://msdn.microsoft.com/en-us/library/b0084kay.aspx

Hiradur:
GUI: I got confused by the thrust track files I had lying around and used the wrong one in my first merge request which is what I "fixed" here. :-P

The problem with _WIN32 is that it presumably prevents other compilers than VS to work with the code when compiling for Windows. E.g. MinGW defines _WIN32 but since it's using the gcc compiler it will probably complain about the code. Didn't try it though.

PR #36, opened 2015-01-30
Update OutProtocol.h

zappadoc:
Rigs Of Rods is now supported by our game devices manager SLIMax Manager, here is an addition to extend the outgauge API with backward compatibility, some important data are missing in the current API like the max RPM of current engine, the distance traveled, the ignition/starter status, etc. so here is my 2 cents to extend it smoothly just by altered the OutProtocol header and cpp.
To activate the new OutGauge V2 structure you just have to change the OutGauge Mode to 3 (see OutProtocol.cpp).
This contribution includes a few additional data just to see if you agree with the change after that I or someone else will be able to add other telemetry data.
Cheers,
zappadoc

### DIFF
--- a/source/main/gameplay/OutProtocol.h
+++ b/source/main/gameplay/OutProtocol.h
@@ -106,6 +106,14 @@ private:
 		char           Display2[16]; // Usually Settings
 		int			   ID;           // optional - only if OutGauge ID is specified
 	});
+		
+	// Zappadoc addition 2015-01, to extend outgauge API with backward compatibility
+	PACK (struct OutGaugePackV2 : public OutGaugePack // include v1
+	{
+		unsigned cha	MaxGears;	 // max forward gears
+		unsigned char	IgnitionStarter; // 0=0ff; 1=ignition; 2=ignition+starter (running)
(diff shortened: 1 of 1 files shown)


Hiradur:
line 113 
```unsigned cha MaxGears; = unsigned char?```

Please make sure it works before submitting a pull request.

zappadoc:
sorry again, obviously it was an unsignised char... this was more a "spontaneous" suggested change quickly done but ok next time I will review my code before submit it...
Regards

Hiradur:
I like the idea of adding more telemetry to it but if we merged this without checking it the project wouldn't compile anymore.
Since the changes to OutProtocol.h and OutProtocol.cpp refer to the same feature you should work with them on one branch and only start one pull request.

zappadoc:
ok I understand. I just reviewed and pushed the right code for OutProtocol.cpp.

Aperion:
@zappadoc I'm going to close both of these pull requests for a couple of reasons.

1. There are compilation errors present in the merges
2. Request #37 cannot compile without this pull request and they need to be merged.

PR #37, opened 2015-01-30
Update OutProtocol.cpp

zappadoc:
Rigs Of Rods is now supported by our game devices manager SLIMax Manager, here is an addition to extend the outgauge API with backward compatibility, some important data are missing in the current API like the max RPM of current engine, the distance traveled, the ignition/starter status, etc. so here is my 2 cents to extend it smoothly just by altered the OutProtocol header and cpp.
To activate the new OutGauge V2 structure you just have to change the OutGauge Mode to 3 (see OutProtocol.cpp).
This contribution includes a few additional data just to see if you agree with the change after that I or someone else will be able to add other telemetry data.
Cheers,
zappadoc

### DIFF
--- a/source/main/gameplay/OutProtocol.cpp
+++ b/source/main/gameplay/OutProtocol.cpp
@@ -125,11 +125,7 @@ bool OutProtocol::update(float dt)
 	timer = 0;
 
 	// send a package
-	if( mode == 3) // OutGauge Packet V2
-		OutGaugePackV2 gd;
-	else
-		OutGaugePack gd;
-	
+	OutGaugePackV2 gd; // declare V2 as it is wider and backward compatible with v1
 	memset(&gd, 0, sizeof(gd));
 
 	// set some common things
@@ -208,7 +204,11 @@ bool OutProtocol::update(float dt)
 		}
 	}
 	// send the package
-	send(sockfd, (const char*)&gd, sizeof(gd), NULL);
+	// set the right size of the buffer depending the OutGauge mode
+	int sz = sizeof(OutGaugePack);
+	if(mode == 3)
+		sz = sizeof(OutGaugePackV2);
+	send(sockfd, (const char*)&gd, sz, NULL);
 
 	return true;
 #else
--- a/source/main/gameplay/OutProtocol.cpp
+++ b/source/main/gameplay/OutProtocol.cpp
@@ -125,7 +125,11 @@ bool OutProtocol::update(float dt)
 	timer = 0;
 
 	// send a package
-	OutGaugePack gd;
+	if( mode == 3) // OutGauge Packet V2
+		OutGaugePackV2 gd;
+	else
+		OutGaugePack gd;
+	
 	memset(&gd, 0, sizeof(gd));
 
 	// set some common things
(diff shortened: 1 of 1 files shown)


Aperion:
Is this some kind of github spam? this commit doesn't even look like it it will compile.

Max98:
I don't really see why it won't compile.
The other half is here:
https://github.com/RigsOfRods/rigs-of-rods/pull/36

Aperion:
Because the variable declarations for gd are in the scope of the if else blocks, and immediately go out of scope before any of the calls to gd were made.

    if( mode == 3) // OutGauge Packet V2
        OutGaugePackV2 gd;
    else
        OutGaugePack gd;

    memset(&gd, 0, sizeof(gd)); // <-- gd is undefined
 
is the same as

    if( mode == 3) // OutGauge Packet V2
    {
        OutGaugePackV2 gd;
    } // <- gd goes out of scope
    else
    {
        OutGaugePack gd;
    } // <- gd goes out of scope

    memset(&gd, 0, sizeof(gd)); // <-- gd is undefined

Hiradur:
Maybe it's Visual Studio being tolerant again? Will compile tomorrow.

zappadoc:
oupppsss! sorry I pushed the code too quickly! :) Obviously the var gd is out of the scope as stated by  Aperion. Sorry again

zappadoc:
my suggestion is to declare the  biggest struct as it is backward compatible with V1:

 OutGaugePackV2 gd;
 memset(&gd, 0, sizeof(gd));

...
// and send the right buffer (packet) depending the mode:

int sz = sizeof(OutGaugePack);
if( mode == 3) // OutGauge Packet V2
       sz = sizeof(OutGaugePackV2);

// send the package
send(sockfd, (const char*)&gd, sz, NULL);

Hiradur:
This also was my first idea. Please test it on your system since you probably got everything set up already.

btw, you can put code in \`\`\` <code> \`\`\`

zappadoc:
well, as I said it was a quick suggestion to improve the OutGauge API and match my minimum requirement for my game devices manager SLIMax Manager (you can see the manager in action with several hardware configurations here: http://www.eksimracing.com/media ).

>>Please test it on your system since you probably got everything set up already.

unfortunately I do not have the full RoR project ready...I just download the source and mod the OutGauge part... if you have an archive with the complete project (ideally for VS 2010) I will be pleased to test it and improve the API, no problem.

Hiradur:
> unfortunately I do not have the full RoR project ready...I just download the source and mod the OutGauge part... if you have an archive with the complete project (ideally for VS 2010) I will be pleased to test it and improve the API, no problem.

I don't have an archive ready since I use Linux but it shouldn't take to long to set it up. Somewhere in these instruction is a download link to precompiled dependencies which should get you set up in no time: http://www.rigsofrods.com/wiki/pages/Compiling_Sources_under_Windows
It's also VS2010

zappadoc:
I checked the project with VS2010, CMakeList should be fixed with the correct path of dependencies (removed source/ ).
SET(ROR_DEPENDENCY_DIR       "${RoR_SOURCE_DIR}/dependencies")

also these resource was missing in full install of 0.4 but required by the new compiled build:
gui_menu_wallpapers.zip
gui_startup_screen.zip

I created the archive for both and add them to resources directory
the project compiled correctly but crash just after the startup screen... the CMake didn't create a DEBUG build so I cannot easily start debugging the app to see what's going on... sorry no time for instance for that.
Regards

Hiradur:
Alright, best to wait for our Windows experts @Max98 and @only-a-ptr then. Thanks for testing.

Max98:
@zappadoc 
I see that you downloaded a version then compiled over it.
No, that's not how it works. You fork this repo and compile, it will work directly because all the needed files are in there.

zappadoc:
@Max98 
>About the DebugBuild, use RelWithDebInfo instead.

I will do it as soon as I find the time, no problem. Fortunately my change is really minor and can be tested easily.

Max98:
As far as i know, we don't have one. And before we can merge the edits, you have to compile it, test it and see if it's working. :p

Aperion:
@zappadoc I'm going to close both of these pull requests for a couple of reasons.

1. There are compilation errors present in the merges
2. This request won't compile without #36

PR #46, opened 2015-02-11
GUI Overhaul merge

Max98:
Let's hope that this can be merged. 
While syncing with the upstream, I had to fix few conflicts.

So this includes ONLY the GuiOverhaul and all it's new components like the new Console, Notification box, new dashboard, new ingame settings, etc..


Hiradur:
Did you fix the font problem on Linux?

Max98:
I can't debug the problem on Linux simply because I don't have experience with GCC. It would be great if you could.

only-a-ptr:
Excellent work, Max98!
I'll merge it as soon as I get home

Max98:
I think that we should merge it as soon as we fix the Linux problem. I'll try to compile on Linux and see what I can do.

Hiradur:
It crashes on startup, same as #10 you probably reverted the fix somehow and maybe others as well.
I don't think that gdb will be of much help to find the cause of the font problems. I probably won't be able to take an in-depth look before the weekend.

Max98:
@Hiradur https://github.com/RigsOfRods/rigs-of-rods/issues/10 Yup, looks like it so. I made it so that when you enter ingame, the GUI_Menu gets init'd and when you go back to the menu, it's destroyed. For some reasons, making it invisible doesn't work (It just don't want to go invisible)

Hiradur:
On shutdown I get this sometimes (random, for obvious reasons). I've never seen this with upstream:
```
RoR: ../nptl/pthread_mutex_lock.c:350: __pthread_mutex_lock_full: Assertion `(-(e)) != 3 || !robust' failed.
```

Hiradur:
Alright, I found the cause for the font problem. The properties are missing. With this MyGUI_FontsEnglish.xml it's working but needs fine tuning, characters are thin and hard to read (and I think it still looks different to Windows):
```
<?xml version="1.0" encoding="UTF-8"?>
<MyGUI type="Resource" version="1.1">
	<Resource type="ResourceTrueTypeFont" name="Default">
		<Property key="Source" value="DejaVuSans.ttf"/>
		<Property key="Size" value="12"/>
		<Property key="Resolution" value="50"/>
		<Property key="Antialias" value="true"/>
		<Property key="SpaceWidth" value="4"/>
		<Property key="TabWidth" value="8"/>
		<Property key="CursorWidth" value="2"/>
		<Property key="Distance" value="7"/>
		<Property key="OffsetHeight" value="0"/>
		<Codes>
			<Code range="33 126"/>
			<Code range="1025 1105"/>
			<Code range="8470"/>
			<Code hide="1026 1039"/>
			<Code hide="1104"/>
			<Property key="Resolution" value="50"/>
		</Codes>
	</Resource>
	<Resource type="ResourceTrueTypeFont" name="DefaultBig">
		<Property key="Source" value="DejaVuSans.ttf"/>
		<Property key="Size" value="15"/>
		<Property key="Resolution" value="50"/>
		<Property key="Antialias" value="true"/>
		<Property key="SpaceWidth" value="4"/>
		<Property key="TabWidth" value="8"/>
		<Property key="CursorWidth" value="2"/>
		<Property key="Distance" value="7"/>
		<Property key="OffsetHeight" value="0"/>
		<Codes>
			<Code range="33 126"/>
			<Code range="1025 1105"/>
			<Code range="8470"/>
			<Code hide="1026 1039"/>
			<Code hide="1104"/>
		</Codes>
	</Resource>
	<Resource type="ResourceTrueTypeFont" name="DefaultSmall">
		<Property key="Source" value="DejaVuSans.ttf"/>
		<Property key="Size" value="8"/>
		<Property key="Resolution" value="50"/>
		<Property key="Antialias" value="true"/>
		<Property key="SpaceWidth" value="4"/>
		<Property key="TabWidth" value="8"/>
		<Property key="CursorWidth" value="2"/>
		<Property key="Distance" value="7"/>
		<Property key="OffsetHeight" value="0"/>
		<Codes>
			<Code range="33 126"/>
			<Code range="1025 1105"/>
			<Code range="8470"/>
			<Code hide="128"/>
			<Code hide="1026 1039"/>
			<Code hide="1104"/>
		</Codes>
	</Resource>
</MyGUI>
```

Max98:
They got screw'd up on windows ;(
I'll take a look why tonight.

only-a-ptr:
@Hiradur: I'm also getting random crash on shutdown, caused by collecting joystick input from OIS. Seems like OIS is destroyed too early. However, the root problem is that shutdown code is still left in RoRFrameListener and called from various places.
I'm already tired of this system and the crashes it gives everybody, so here's what I'll do:
1. Create an abstract interface class for MainThread, containing RequestShutdown() function. This will also reduce compile-time dependencies.
2. I'll port all the shutdown code to MainThread.
3. I'll replace all calls to RoRFrameListener->*shutdown*(). 

However, to perform this refactor, I absolutely need Max98's changes already merged.

Hiradur:
> However, to perform this refactor, I absolutely need Max98's changes already merged.

I'm fine with this, the only problem is that without the properties mentioned above the font is super tiny, glitched and missing white spaces on Linux.
@Max98: I realized that there was a <Property key> in the code section of the first <Resource type> in my fix. Must have happened when I copy&pasted it. I removed it from my fix. Just so you know.

Max98:
@Hiradur Tried again and still the same result.

screenshot_8

I'll take a look :p

Hiradur:
It works. Font is still a little small for my taste but at least it looks correct now.

Max98:
Oh, thanks for the idea. I'll try to make the font size changeable ingame. :p²

Max98:
I think that we are ready for the merge.

Max98:
So guys? @Hiradur @only-a-ptr

Hiradur:
Push the button.

Max98:
Done. :)

PR #59, opened 2015-02-15
Cleanup calc beam forces euler

Aperion:


Aperion:
Closed this since it had more reviion history than I intended

PR #62, opened 2015-02-15
Separate calcForcesEulerCompute into smaller functions

Aperion:
This separates calcForcesEulerCompute into seperate functions. I   used the BES_START/BES_STOP as guide for new  function sections.


Hiradur:
Love these cleanups but do the function calls decrease performance? Maybe they should be inlined.

PR #63, opened 2015-02-15
Separate BeamData.h into smaller header files

Aperion:
This separates the most commonly used parts of BeamData.h into separate headers.  I've included the new headers in BeamData.h to avoid breaking existing code.


Max98:
Did you sync your local with upstream before doing this? Because this define is missing:

>//skidmarks v2
>bool isSkiding;

Aperion:
OMG. I merged that about 3+ times tringt o get the commits sorted out. can't beleive I missed  that in the one I pushed :( I will fix real quick

Max98:
hehe :P
Also, it would be great if next time you could use the Commit style, so in the Upstream, all commits will be clear: http://www.rigsofrods.com/wiki/pages/Commit_style

Aperion:
I'll try to remember. Go ahead and close, I'll redo tomorrow. Night!

Max98:
Alright, Goodluck! :P

PR #70, opened 2015-02-16
Workaround for FF and legal stuff

Hiradur:
1. Disabled Force feedback for Linux since it only causes crashes with a controller connected.
2. Added a bunch of legal notices that I feel are necessary  
Please review the legal stuff and let me know what you think.


Max98:
It seems like we need those IMO.

Hiradur:
Yeah, I forgot you :P 
No, I figured since you would have to complete your entries anyway I don't have to include you.
Does somebody happen to know the license for the assets?

Aperion:
Good idea, althouh to be picky shouldn't  the FF work aruond be done in a different pull request?

Yes, you missed me for both the previous authors, and the current authors. And Box5Diesel should be added to the asset list.

Hiradur:
FF: Yeah that's true, I already created a branch for the legal stuff but forgot to reset the FF commit. I realised that after merging with upstream and didn't feel like doing it again since the download was big due to the changes to binary files.

I thought you would rather want to add yourself. But if you send me the required information (forum pm maybe) I can add all of you right away so everything is bundled in one pull request.

Aperion:
I'm fine with adding my self. One thing that was never talked about
previously was copyright. I know pricordes name is always up to date but
hasn't contributed in ages. I've seen my name removed, or not attributed,
and I've never seen a conversation, or notice about it.

One more thing, Thomas and pricordes copyright dates should match with
their contribution dates. Unless something else wants to be done with
copyright, like transfer to another entity or... Not sure what else.

Hiradur:
I got the feeling that copyright is an issue here before (e.g. assets). There probably is no way to make this completely correct now without replacing all the old stuff which isn't what we intend to do.
Another way would be to go through the SVN and Hg commits but I don't think that anybody would be willing to do this.
I started this legal stuff mainly because the longer we wait the more problematic the whole copyright thing will become. Some open source gurus don't like to work on projects with uncertain licenses.
A recent example would be Cube2:Sauerbraten. While the source code is zlib/libpng-licensed they suddenly realized they don't know what most of their assets are licensed as leading to decline in development.

Aperion:
I'll poke around the invites board and see if I can dig up any discussions
about content licensing. I know they happened I just never paid too much
attention.

PR #79, opened 2015-02-17
[Cleanup] Separate calcForcesEulerCompute into smaller functions

Aperion:
[Prepare] This is some cleanup used to integrate the new drivetrain changes

This separates calcForcesEulerCompute into seperate functions. I used the BES_START/BES_STOP as guide for new function sections.


Aperion:
@Hiradur from #62 
> Love these cleanups but do the function calls decrease performance? Maybe they should be inlined.

Technicaly adding function calls where non existed will always produce slower code.

Realistically it can lead to faster more mainainable code by understanding how it works, or even introducing different mechanisms. For example jump tables could be used for different beam types instead of nested branch code. 

Worrying about th performance of function calls is pre-optimizaton(i.e. the root of all evil), make it work, then make it fast.

Hiradur:
:+1:

> Realistically it can lead to faster more mainainable code by understanding how it works, or even introducing different mechanisms. For example jump tables could be used for different beam types instead of nested branch code.

I mean inlining with the 'inline' keyword. This wouldn't change readability at all. However, this would require more space in cache when working with threads which could lead to decreased performance which is why I said 'maybe' (ie profile first).

Aperion:
Ah. I think the inline keyword is more of a suggestion than a requirement
when used, so it's not guaranteed to be inline.

Hiradur:
This is correct. Only way to know for sure if the compiler inlined or not is to look at the assembler output. But it increases the probability that the compiler inlines the functions which may be a good thing since it may not know that these functions are the hotspot of the program.

Max98:
So when will this get merged?

Max98:
I see few mistakes in here, will fix them.

Max98:
Vehicles are exploding now, will try to see why.

Hiradur:
Looks similar to the OpenGL problem now, maybe it's a timing issue?

Max98:
It's timing, or calculations order.
Using OpenGL, when you spawn the Agora Bug, do the shocks explode?

Hiradur:
Before or after the cleanup? Before they sometimes broke but this didn't happen very often.

Max98:
Well, now they always break.

Hiradur:
I once replaced all approxsqrt/fastinvsqrt calls with std::sqrt() to see if it has an impact in performance and it created some anomalies in physics. Probably not related to this but I thought I'd share it anyway. I think we should use sqrt() everywhere, it decrease FPS a bit but its behaviour is deterministic.

Hiradur:
I think I found the problem: calcBeams is called twice: One time in calcForcesEulerCompute and once in calcForcesEulerPrepare. Commenting out in calcForcesEulerPrepare fixed it for me.

Max98:
Then you should merge the fix with upstream.

Hiradur:
I'd like Aperion to do it since there seems to be more duplicates like slidenodes etc.

Max98:
Yeah but Upstream is not usable at its current state and I believe the Aperion is busy for the moment.

only-a-ptr:
@Aperion : I like the idea of cleaning up the physics code, but please DO NOT merge it with upstream yet. At least not until we release 0.4.5. The current milestone, dubbed "NextStable" is aimed to stabilize the codebase after my changes to truck parsing and spawning. RoR hasn't had a stable release in ages and I'd really like to deliver one soon. Refactorings and breaking cleanups may come later. Please keep your work in your repository and release a test build on the forums if you want to do a proof-of-concept for something.

I'm also doing experimental cleanups and keeping them in my repo, see http://www.rigsofrods.com/wiki/pages/Programmers_%26_Projects#only_a_ptr

Current physics code is fragile. A slightest change in computation accuracy or timing may have fatal results. Myself, I'm carefully walking around it until I get better understanding. You may know the codebase better than me since you've been a contributor before, but the discussion after this merge suggests you've pushed in breaking changes. This puts the NextStable project at risk.

Max98:
So any fixes about this yet?

Aperion:
@only-a-ptr This should have only been a refactor, only a change to structure and not execution order.If the problem is calling calcBeams twice then it should be easy to find. 

Refactoring will always be uncertain without unit testing.

only-a-ptr:
@Aperion : OK, but my point stands: There's no reason to hurry with pushing refactors into the upstream. Please keep them in your branches until later time.
Right now, I'm using the BeamData.h as a pseudo unit test: I load a vehicle using both v0.4.0.7 codebase and upstream, and I print all Beam data using RigInspector class [/tools]. Doing changes to BeamData classes would break my workflow and cause the debugging to become utterly hard. Yes, I know you've only moved the code so far, but still.
Debugging vehicle support is difficult: old truck parser processed sections in undeterministic order, allowing many unexpected combinations. My parser uses deterministic order, thus breaking all the quirks which old content depended on. Example: half-baked support for named nodes, where half of the sections referred to named nodes by generated numbers.

PR #93, opened 2015-02-23
Rearrangement of the files

Max98:
One could say that this way, the source is much more organized. There are still few things that I want to move there and here but I need to know what you guys think of this.

I suggest you to browse the repository rather than taking a look at the commits.


Aperion:
why do you seam tto always have such huge commits? they really need to be broken down.

Max98:
Yes, I'm aware of that, it's for the simple reason that I was switching too much between branches for some reasons and I was lazy to stash.. :P

PR #96, opened 2015-02-25
[CodeChange] Cleared out few things

Max98:
Renamed CAELUM_ stuffs to SKY_..


Max98:
Actually, this shouldn't be merged right now.

PR #107, opened 2015-03-04
Progress

Max98:


Max98:
Here's the new input.map 
https://gist.github.com/Max98/003e491d6662ee8c4e63

Issue #112, opened 2015-03-06
Change version number?


Max98:
I know that this is useless but shouldn't we change the version number to 0.4.5.0? :P

Hiradur:
More like 0.4.5-dev, this will prevent confusion in case you want to look at the code of an older release in the future.

Max98:
You didn't get me, or I didn't get you. What i mean is simply to increase the version numbers here: https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/RoRVersion.h

How will we put the -dev tag thing? Unless we put a boolean variable in RoRVersion.h and make it do the work for the -dev tag.

Hiradur:
I got you.
IMO it should be 0.4.5 only for the release tag and 0.4.5-dev before the tag to differentiate between release and dev build. Everytime a release gets tagged we remove the -dev suffix for the tag and afterwards increase the version number and attach -dev again.

Max98:
Done.

PR #118, opened 2015-03-07
Increased version to 0.4.5 and added the -dev tag

Max98:
As said here: https://github.com/RigsOfRods/rigs-of-rods/issues/112

Will not merge until you guys agree.


Hiradur:
I think ROR_DEVMODE is not a good name for this. I suggest to use ROR_INDEV + a comment to explain.  
Or we include it directly in the string:
```
#define ROR_VERSION_STRING "0.4.5.0-dev"
```

Max98:
Will do that then :P

Max98:
@Hiradur Should I merge it like that? :P

Max98:
I'll go on and merge it, as because I did the exact thing that Hiradur said. :P

PR #165, opened 2015-04-24
Compilation fixes

skybon:
This fixes compilation broken by previous merge.


Max98:
Don't you think that and0 is a little bit weird? Change it to b_and and add the tag [Fix] to the commit and it will get merged.

PR #172, opened 2015-05-07
Remove stencil shadows

skybon:
Stencil shadows were scrapped in Ogre 2.0 and are not maintained in RoR. It is time for us to say goodbye, cut them and move on.


Max98:
Good work done there! :)

only-a-ptr:
Nice work, thanks.

PR #176, opened 2015-05-15
Travis CI integration

Hiradur:
So far it's only using gcc to build RoR. Test build took ~43min which is close to the time limit of 50min set by Travis. It can be speed up by prebuilding the dependencies and having Travis download them from a server which is not included in this PR.


masterkorp:
:+1:

PR #177, opened 2015-05-16
Add #pragma once to all headers to speed up build process

Hiradur:
Left old include guards intact, for compatibility.


Hiradur:
Shaved off 8 minutes of Travis build time even though gcc is optimized for include guards. Nice.

Max98:
Then what's the point of leave these?
> #ifndef UTILS_H__
> #define UTILS_H__

Hiradur:
I didn't touch external sources.

PR #219, opened 2015-06-01
More progress on GUIOverhaul

Max98:


Hiradur:
Nice and calm tune but IMO .wav is a bad idea due to its size. What about encoding it using Opus (the new kickass codec which destroys mp3 and vorbis)? I propose to make opus the standard codec for compressed audio. Have a listen to compare: https://opus-codec.org/examples/

I think it doesn't play on Linux in main menu, too. At least I didn't hear anything and didn't see any setting in the configuration menu. There should be a button to disable music in case somebody wants to listen to his own music.

Bob-Z:
Main menu music is playing well here. Using ubuntu 15.04.

vido89:
Can also confirm that audio is playing in main menu, my vote goes to open-source standard for audio enc/dec, and why not use ogg audio when its already well supported across platform, rather then adding one more dependency

Hiradur:
Still no playback on my Laptop. Gotta check on my PC later.

@vido89 ogg is a container format which can carry multiple video/audio codecs. I assume you are talking about the Vorbis codec. Regardless which codec we use it would be a dependency. Only because it's preinstalled on Ubuntu doesn't mean it's available on every other system. Opus is a FOSS codec and may replace Vorbis in the future. 
For RoR it has to be a FOSS codec. Vorbis and Opus are probably our best options. Vorbis is more widespread and widely supported but Opus is a new codec making use of the power of modern machines and knowledge gathered by the development of older codecs. Opus is more futureproof and better than Vorbis in almost every aspect (not just quality but also latency etc.). Opus is also pushed by giants such as Google who want to make it the standard codec for audio applications on the internet.

btw: Opus comes with Ubuntu if I'm not mistaken (at least I don't remember installing it)

PR #265, opened 2015-06-19
Updated Readme, deleted LICENSE.txt

Hiradur:
Anything I should change or add?


DirtGamer301:
To the readme you could add that also many mods can be found on the forum (http://www.rigsofrods.com/forums/103-Showrooms-and-WIP) as many mods simply aren't released on the repo anymore and only on the forums. 

(Sorry for being annoying of saying the same all over)

only-a-ptr:
@Hiradur  I think it's fine this way.
@DirtGamer301 You have a point, although this needs wider discussion. The repo needs a revamp, from both technical and organizational perspective.

PR #269, opened 2015-06-20
[Update] Some cleanups

Max98:


Max98:
Oh, I'll have to merge Hiradur's work into mine first lol.

Hiradur:
I don't think it's necessary, if you check the Travis details you can see that it builds fine, it just appears as failed because cppcheck detected defects in the code. This will make all builds look like fails until the defects are fixed. I admit it's inconvenient, I'll change the behavior.

Max98:
"a bunch of odefs from meshes" 
Didn't happen.
I just removed all the unused resource folders, and::
- Renamed "meshes" to "models"
- created "mygui_layouts" where all the mygui layouts will be stored there.

Also, I'm still not done with this yet. I just didn't have enough time these days to work on RoR.

Hiradur:
@Max98 I didn't see the last points you mentioned because GitHub doesn't display them (too many changes).
I have nothing against these changes, I just thought some insight would be good.

PR #297, opened 2015-07-11
New AngelScript functions && some small fixes 

AnotherFoxGuy:
trying to add some more functions to AngelScript
you can test it using this map 
http://anotherfoxguy.github.io/Files/AITestMaps.zip


Max98:
Nice work done there, love the AI thing, we should develop it more to do other things like races and stuff. 
It should be possible with your code tho.

only-a-ptr:
Hi. I only noticed this change now.

First, thanks for your contribution, greatly appreciated.

However, I have one huge complaint - The  REPO/source/AA_DoxygenMainpage/ - WTF? That's awful! This thing totally belongs to REPO/doc - or better to REPO/doc/doxygen since there's also Sphinx. Please update it.

PR #299, opened 2015-07-14
Alphabetically Sort menu's categories and entries.

Bob-Z:
This alphabetically sorts maps and vehicules categories and entries.
It leaves "All" and "Fresh" categories at the top of the list.


Max98:
This is indeed a nice feature to have, but you didn't respect something: http://www.rigsofrods.com/wiki/pages/Commit_style

If you can't fix it, we will merge it anyway, but it would be better to follow the commit style.

Hiradur:
I think we can merge it, the missing prefix is not the end of the world. 
@Bob-Z Just conform to the commit style rules next time. Anyway, thanks for the contribution!

Bob-Z:
Sorry, for the coding style. I hope it is OK now, let me know

Max98:
Great work done there! :)

PR #337, opened 2015-07-28
RigParser fault tolerance ++

### LINKED ISSUE #335: RigParser / *wheels*: Tolerate both comma and space as separators.
On behalf of this forum post: http://www.rigsofrods.com/threads/119110-Test-Build-Rigs-of-rods-0-4-5-0-dev?p=1384499&viewfull=1#post1384499

The current parser requires strictly using comma or space depending on examples in the documentation, even though old parser accepted both and as a result, lots of vehicles don't load correctly.

However, for stability, it's not desirable to change all parser rules at once. A lot of elements have already been resolved, other may become unstable.

Let's only loosen the rules for all wheel definitions this time.

only-a-ptr:
Closes #335


Max98:
Since this merge, when adding a new map/vehicle, my cache crashes. Will post log later.

PR #386, opened 2015-08-12
[Cleanup] Replaced goto, style improvements

skybon:
This pull requests replaces goto statements with proper function calls.


skybon:
Third party library, never mind.

PR #398, opened 2015-08-13
shadow support for roads and runways

Speciesx:


Max98:
I was too lazy to do all of this. x)

Hiradur:
:+1:

PR #414, opened 2015-08-19
Bug fixes

Max98:


Max98:
Doesn't need travis, not much to break linux

PR #416, opened 2015-08-19
Added friction settings to some default meshes and more shadow support

Speciesx:


Max98:
Travis failed because of the renaming I believe.

PR #429, opened 2015-08-29
dashboard corrections

Speciesx:
Corrected needle and Turbo gauge position, reduced dashboard3500 to same size as dashboard7000.


TechGiga:
I actually liked the large tachometer with the digital speedometer. It made RoR feel like Need For Speed while still being a realistic simulator.

PR #490, opened 2015-12-19
[Codechange] CMake flag update (UNIX)

ulteq:
* Suppresses the excessive "Deprecated Ogre Function" warnings
* Enables -o3 for the release build

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -93,8 +93,8 @@ if(WIN32)
 
 ELSEIF(UNIX)
   set(CMAKE_CXX_FLAGS                       "-std=c++11 ${CMAKE_CXX_FLAGS}")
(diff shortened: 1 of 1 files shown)


Hiradur:
I've actually been thinking about dropping the gcc flags entirely since CMake already has a set of standard flags per ```CMAKE_BUILD_TYPE``` usually set by the distribution/package maintainers. In very rare cases interfacing software built with different optimization levels could crash (this actually does happen on the Ogre-2.0 branch of RoR if you build the dependencies with -O2 and RoR with ```CMAKE_BUILD_TYPE=DEBUG```).
```-O3``` was also known to cause crashes and thus was considered unsafe but this may have changed.

mikadou:
In *modern CMake* compile flags are typically set per target using the `target_compile_options` command. Another possibility which could be used is to apply compile flags to specific files with `set_source_files_properties`. E.g. more agressive O3 optimization could be selectively activated for physics calculations.

ulteq:
@Hiradur, @mikadou agreed, I'll close this for now.

PR #496, opened 2015-12-22
[Cleanup] truckCollision / PointColDetector

ulteq:
**Beam.cpp:**
* Removed redundant intraPointCD objects

* Combined the Prepare, Compute and Final methods

Positive side effect: The interPointCD->update() method is now executed in the thread pool

**PointColDetector:**
* Simplified the bounding box calculation
* Removed redundant bounding box intersection checks


ulteq:
> it's sparse for no reason

It's sparse for easy detection of new (colliding) trucks:
https://github.com/ulteq/rigs-of-rods/blob/pointColBBox/source/main/physics/collision/PointColDetector.cpp#L74

only-a-ptr:
@ulteq OK then.

PR #511, opened 2015-12-30
[Bugfix] Fixes incorrect 'reset in place' alignment on some trailers/loads

### LINKED ISSUE #503: 'reset in place' aligns trailer and loads to SE
After reset (Backspace key) trailer and loads should keep their previous alignment.

ulteq:
Fixes #503

### DIFF
--- a/source/main/physics/Beam.cpp
+++ b/source/main/physics/Beam.cpp
@@ -1066,7 +1066,7 @@ void Beam::resetAngle(float rot)
 
 	// Set up matrix for yaw rotation
 	Matrix3 matrix;
-	matrix.FromEulerAnglesXYZ(Radian(0), Radian(-rot - Math::HALF_PI), Radian(0));
+	matrix.FromEulerAnglesXYZ(Radian(0), Radian(-rot + m_spawn_rotation), Radian(0));
 
 	for (int i = 0; i < free_node; i++)
 	{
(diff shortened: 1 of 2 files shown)


AnotherFoxGuy:
Compiling this with vs10 throws the error 
```Beam.cpp(1311): error C2039: 'round' : is not a member of 'std'   ```
```Beam.cpp(1311): error C3861: 'round': identifier not found  ```

only-a-ptr:
@AnotherFoxGuy : add ```#include <math>```
http://en.cppreference.com/w/cpp/numeric/math/round

AnotherFoxGuy:
That trows the error 
```error C1083: Cannot open include file: 'math': No such file or directory```

Hiradur:
try ```#include <cmath>```
but since round() is a c++11 feature it may be missing in VS 2010

AnotherFoxGuy:
Nope, still throws the error 
```'round' : is not a member of 'std'   ```

Hiradur:
It's probably missing in VS 2010 then, similar to #494 
Instead of rewriting the C++11 standard library for VS2010 you Windows guys should update your VS. :P

mikadou:
> Instead of rewriting the C++11 standard library for VS2010 you Windows guys should update your VS. :P

:+1: +1

only-a-ptr:
> Instead of rewriting the C++11 standard library for VS2010 you Windows guys should update your VS. :P
:+1: :smile:

PR #516, opened 2016-01-02
[Cleanup] Simplifies the overly complex threading logic

ulteq:
This only affects the **interTruckCollision**:

* Instead of having a fixed number of thread tasks and skipping trucks based on the thread index, we now have one thread task per simulated truck.
* The workload is distributed dynamically by the operating system.

Positive side effect: A better (more uniform) CPU usage when multiple vehicles collide.

**Edit 1:**
* Moved parts of the multi-threading logic from the Beam class into the BeamFactory class.

**Edit 2:**
* Fixes a bug where only one truck would be simulated when the thread pool is deactivated.


only-a-ptr:
Yay! :smiley:

Issue #174, opened 2015-05-09
Angelscript not working correctly on Linux x86_64


tritonas00:
Loading this server: cifpo modded ror.ezzg.be:14005 nHelens (http://www.rigsofrods.com/repository/index/Search.keywords:nHelens)

In 0.38.66 looks ok:
http://s10.postimg.org/5tfhzrdop/2015_05_09_120708_1920x1080_scrot.png

Now (latest git), colors are different and missing elements: http://s10.postimg.org/6rvv21r0p/2015_05_09_120802_1920x1080_scrot.png

[RoR.log](https://www.dropbox.com/s/qrld4bk618kf1nq/RoR-multi.log?dl=0)

Also in single player the water is above the map:
http://s10.postimg.org/tj4xogc21/2015_05_09_120906_1920x1080_scrot.png

[RoR.log](https://www.dropbox.com/s/7bg243fquco8gwj/RoR-single.log?dl=0)

Disabling waves fixes the issue. (Thanks Skybon)

Max98:
The coloring problem is just a conversion mistake. So it's not our fault. But for me, going on the same server, all the objects load. The only problem I see is the water about the ground with enabled waves. I'll take a look at your logs.

tritonas00:
> But for me, going on the same server, all the objects load.

In Windows?

Max98:
Yup, so it can be a problem from AngelScript.

tritonas00:
> Yup, so it can be a problem from AngelScript.

Is there anyway to debug this? Its a bit annoying :P

Also this map has two spawn places, and because the one isnt loading it spawn me in the other...

tritonas00:
Maybe those angelscript logs are usefull

0.38.66
http://pastebin.com/epcwnyVX

latest git
http://pastebin.com/6SCjQvWA

Hiradur:
Which AngelScript version do you use?

tritonas00:
I use 2.20.3 in 0.38.66 and 2.22.1 in git

tritonas00:
Ok i have more problems on this map in multiplayer

Sometimes the game logins fine but sometimes (most) simply crashes.

gdb:

    Program received signal SIGSEGV, Segmentation fault.
    #0  0x00007ffff7490959 in Ogre::SceneNode::setVisible(bool, bool) () from /opt/ror-next/lib/libOgreMain.so.1.8.1

dmesg:

[...]

PR #518, opened 2016-01-03
[Bugfix] Fixes #459

### LINKED ISSUE #459: Crash in MP when another player spawns a vehicle
Compiled latest git in Archlinux and downloaded 0.4 High Quality Pack.

Tried to login in Awesome_Gaming_Mit (f4afUID-flat_mitcity) server. (http://www.rigsofrods.com/repository/download/4932)

First attempt:

>Program received signal SIGSEGV, Segmentation fault.
0x00007ffff749ed99 in Ogre::SceneNode::setVisible(bool, bool) () from /usr/lib/libOgreMain.so.1.9.0

Removed everything from packs folder, left only the map and tried again, game logins nicely.

Restored packs folder, cleaned and regen cache, logged in, but most of the players are out. While playing, one other player came in and spawned a vehicle (gavrilzeta). Game crashes.

gdb output->http://pastebin.com/qmc1WpqH
gdb bt full->http://pastebin.com/wHKaGPLZ

Bob-Z:
This seems to fix the multi-player crash for me.
The "access element" paragraph of this page http://www.cplusplus.com/reference/vector/vector/operator%5B%5D/ inspired me.

### DIFF
--- a/source/main/datatypes/rig_t.h
+++ b/source/main/datatypes/rig_t.h
@@ -207,7 +207,7 @@ struct rig_t
 	std::vector<Ogre::Entity*> deletion_Entities; //!< For unloading vehicle; filled at spawn.
 	std::vector<Ogre::MovableObject *> deletion_Objects; //!< For unloading vehicle; filled at spawn.
 	std::vector<Ogre::SceneNode*> deletion_sceneNodes; //!< For unloading vehicle; filled at spawn.
-	int netCustomLightArray[4];
+	unsigned int netCustomLightArray[4];
 	unsigned char netCustomLightArray_counter;
 	MaterialFunctionMapper *materialFunctionMapper;
 	bool ispolice;
--- a/source/main/physics/Beam.cpp
+++ b/source/main/physics/Beam.cpp
@@ -6632,8 +6632,11 @@ void Beam::setCustomLightVisible(int number, bool visible)
 
 	try
 	{
-		if (flares[netCustomLightArray[number]].snode)
-			flares[netCustomLightArray[number]].controltoggle_status = visible;
+		if (flares.at(netCustomLightArray[number]).snode)
+			flares.at(netCustomLightArray[number]).controltoggle_status = visible;
+	}
+	catch (std::out_of_range oor)
+	{
 	}
 	catch (Exception ex)
 	{
--- a/source/main/physics/Beam.cpp
+++ b/source/main/physics/Beam.cpp
@@ -6624,25 +6624,18 @@ bool Beam::getCustomLightVisible(int number)
 
 void Beam::setCustomLightVisible(int number, bool visible)
 {
-	if (number >= 5)
+	if (number < 0 || number > 4)
 	{
-		LOG("AngelScript: Light ID (" + TOSTRING(number) + ") overflow, max: 4...");
+		LOG("AngelScript: Invalid Light ID (" + TOSTRING(number) + "), allowed range is (0 - 4)");
 		return;
 	}
 
-	try
+	unsigned int flareID = netCustomLightArray[number];
+
+	if (flareID < flares.size() && flares[flareID].snode)
 	{
-		if (flares.at(netCustomLightArray[number]).snode)
-			flares.at(netCustomLightArray[number]).controltoggle_status = visible;
+		flares[flareID].controltoggle_status = visible;
 	}
-	catch (std::out_of_range oor)
-	{
-	}
-	catch (Exception ex)
-	{
-	}
-	/*else
-		LOG("AngelScript: Light ID (" + TOSTRING(number) + ") doesn't exist, ignored...");*/
 }
 
 bool Beam::getBeaconMode() // Angelscript export
--- a/source/main/physics/Beam.cpp
+++ b/source/main/physics/Beam.cpp
@@ -6618,7 +6618,7 @@ bool Beam::getBrakeLightVisible()
 
 bool Beam::getCustomLightVisible(int number)
 {
-	return netCustomLightArray[number] != -1
+	return netCustomLightArray[number] != UINT_MAX
 			&& flares[netCustomLightArray[number]].controltoggle_status;
 }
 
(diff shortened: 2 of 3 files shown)


only-a-ptr:
Thanks a lot for the contribution! Yes, the code seems suspicious.
However, I added 2 notes to the code - please attend to them.

Hiradur:
Thanks, however for the next commit please use a commit headline like ```[Fix] Crash in Multiplayer due to OOB access``` because nobody knows of the bat what issue number 459 is. You can link to the issue in the message field of the commit.

Also, if I understood this correctly (https://github.com/RigsOfRods/rigs-of-rods/issues/174#issuecomment-167164294 ) then it's always trying to access the element with index -1 which doesn't exist of course. The question is, why does it do that and can it be prevented?

Bob-Z:
Sorry for the commit headline. I just copied another commit message in the log. You are totally right that this is not a relevant message.

Accesing a non existent index of a std::vector with the [] operator results in undefined behaviour. In linux 64 bits it seems that undefined behavior is SIGSEGV. (Note that I have seen sometimes that, even with index set to -1, the call to flares[-1] do not crash and everything works as expected. This is the definition of undefined behavior !)

I don't like the idea of using -1 as an index in this case because -1 is cast to unsigned int in calls to [] operator or at() method. This means that accessing index -1 is accessing elements 0xFFFFFFFF, which fails the bound checking of at() method. Anyway I have seen this use of -1 in other projects so maybe it is a good practice for some reasons, I don't know...

There may be the same bug elsewhere in the code where std::vector::operator[] is used inside a try/catch block. But since it seems to happen only on 64 bits platforms, they will be slower to emerge.

Bob-Z:
In fact for each use of std::vector::operator[] (even outside a try/catch block) we should either make sure index is in-bounds or use at() method instead (which should be a bit slower).

ulteq:
My suggestion:

```
void Beam::setCustomLightVisible(int number, bool visible)
{
	if (number < 0)
	{
		LOG("AngelScript: Light ID (" + TOSTRING(number) + ") underflow");
		return;
	}

	if (number >= 5)
	{
		LOG("AngelScript: Light ID (" + TOSTRING(number) + ") overflow, max: 4...");
		return;
	}

	int flareID = netCustomLightArray[number];

	if (flareID >= 0 && flareID < flares.size() && flares[flareID].snode)
	{
		flares[flareID].controltoggle_status = visible;
	}
}
```

only-a-ptr:
@Bob-Z Oops, I didn't know the ```index``` parameter receives negative input, too. Anyway that means the function has been unsafe all along.
@ulteq: That way, it still doesn't check array overflow.

My suggestion:
```
void Beam::setCustomLightVisible(int number, bool visible)
{
    if (number < 0 || number > 4)
    {
        LOG("AngelScript: Invalid Light ID (" + TOSTRING(number) + "), allowed range is (0 - 4)");
        return;
    }

    if ()
    {
        LOG("AngelScript: Light ID (" + TOSTRING(number) 
            + ") overflow, array size is: " + TOSTRING(netCustomLightArray.size()) );
        return;
    }

    int flareID = netCustomLightArray[number];

    if (flareID >= 0 && flareID < flares.size() && flares[flareID].snode)
    {
        flares[flareID].controltoggle_status = visible;
    }
}
```

ulteq:
`netCustomLightArray` is not a std::vector.
https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/datatypes/rig_t.h#L210

only-a-ptr:
@ulteq What?? Then how did it throw std::out_of_range for BobZ?

Bob-Z:
Thanks for your comments. I have updated my pull request accordingly.

Anyway I have 2 more questions:
 - Is the AngelScript reference in the LOG text relevant ?
 - Why custom light number is limited to 4 ?

mikadou:
@only-a-ptr Just curious, how is this related to c++11?

ulteq:
@Bob-Z, you must also change this line: https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/physics/Beam.cpp#L6612

ulteq:
Since `getCustomLightVisible` can also be called from AngleScript, why not do it like that:

```
bool Beam::getCustomLightVisible(int number)
{
    if (number < 0 || number > 4)
    {
        LOG("AngelScript: Invalid Light ID (" + TOSTRING(number) + "), allowed range is (0 - 4)");
        return false;
    }

    unsigned int flareID = netCustomLightArray[number];

    return flareID < flares.size() && flares[flareID].controltoggle_status;
}
```

Hiradur:
@tritonas00 told me he hasn't had any crash since he downgraded to Ogre 1.8. I know @only-a-ptr was unable to reproduce this crash and he's also been using Ogre 1.8. 
I tested with the 0.4.5.1 release on Windows and it crashed for me (it's using Ogre 1.8, isn't it?)
Any more input from people using Ogre 1.8?

only-a-ptr:
@Hiradur I have nothing to add. However, this pull request definitely fixes an unsafe function, even if it doesn't fix the MP crash completely.

@Bob-Z The custom light limit of 4 is defined by RoRNet: https://github.com/RigsOfRods/rigs-of-rods/blob/86b6bc5f95b9abd937b270747b2c7e0137b65fe8/source/main/network/protocol/rornet.h#L128-L131

@mikadou BobZ's code relies on using ```size()``` and ```at()``` functions with a plain C-array. This is only possible in C++11 and further: http://www.cplusplus.com/reference/array/array/size/ http://www.cplusplus.com/reference/array/array/at/

Bob-Z:
Please do not merge now, I will add @ulteq  modif on getCustomLightVisible for consitency.

only-a-ptr:
@Bob-Z OK. However, that would be 4 commits for one simple change. For cleaner commit history, please create a new branch at https://github.com/Bob-Z/rigs-of-rods/commit/b7b9810f189b3837c14bd46cd6052b92c6eb3f12 and prepare a single clean commit to be merged.

PR #520, opened 2016-01-04
[Fix] CMake/UNIX: Build with Ogre 1.8

Hiradur:
- allows to build RoR on older Linux distributions with only Ogre 1.8 in the repositories
- allows Intel GPU users to opt for Ogre 1.8 instead of Ogre 1.9, @tritonas00 told me Ogre 1.9 doesn't work as well with the Intel graphics driver


Hiradur:
Ugh, just realized that this created a new branch on upstream. Sorry, I used the webinterface...

PR #524, opened 2016-01-05
[Fix] Crash in Multiplayer due to out of bounds access

Bob-Z:
Fix flares out of bounds access detection.
Fixed in Beam::setCustomLightVisible and Beam::getCustomLightVisible.
This fix bug #459


tritonas00:
:+1:

Hiradur:
:+1:

PR #532, opened 2016-01-07
[Bugfix] Fixes incorrect 'reset in place' alignment on most trailers/loads

### LINKED ISSUE #503: 'reset in place' aligns trailer and loads to SE

ulteq:
Nothing sexy, but it works on all trailers/loads I have tested so far.

* Should fix #503


Hiradur:
:+1:

Issue #534, opened 2016-01-07
Build without Angelscript is broken


Hubbitus:
Building 
```
[  0%] Building CXX object source/main/main_sim/CMakeFiles/RoR.dir/MainThread.cpp.o
/home/pasha/SOFT/rpmbuild/BUILD/rigs-of-rods-c10f9acd5ccc6147a507ec2baed8f2e564d94bd0/source/main/main_sim/MainThread.cpp:63:26: fatal error: ScriptEngine.h: No such file or directory
 #include "ScriptEngine.h"
```
In MainThread.cpp:63 I have changed
```
#include "ScriptEngine.h"
```
to
```
#include "scripting/ScriptEngine.h"
```
then got:
```
/home/pasha/SOFT/rpmbuild/BUILD/rigs-of-rods-c10f9acd5ccc6147a507ec2baed8f2e564d94bd0/source/main/scripting/ScriptEngine.h:33:47: fatal error: scriptdictionary/scriptdictionary.h: No such file or directory
 #include "scriptdictionary/scriptdictionary.h"
```
such file I found in ./angelscript_addons/scriptdictionary/scriptdictionary.h, but I configure it without angelscript (-DROR_USE_ANGELSCRIPT="FALSE") due to the rely RoR on outdated API - https://github.com/RigsOfRods/rigs-of-rods/issues/530

Please help.

Hiradur:
You are probably the first to build without Angelscript in years.

```
diff --git a/source/main/main_sim/MainThread.cpp b/source/main/main_sim/MainThread.cpp
index 5aa93ff..d325997 100644
--- a/source/main/main_sim/MainThread.cpp
+++ b/source/main/main_sim/MainThread.cpp
@@ -60,7 +60,9 @@
 #include "RigEditor_Config.h"
 #include "RigEditor_Main.h"
 #include "RoRFrameListener.h"
-#include "ScriptEngine.h"
+#ifdef USE_ANGELSCRIPT
+  #include "ScriptEngine.h"
+#endif
 #include "Scripting.h"
 #include "Settings.h"
 #include "Skin.h"
@@ -237,7 +239,9 @@ void MainThread::Go()
 	// Create legacy RoRFrameListener
 
 	gEnv->frameListener = new RoRFrameListener();
+#ifdef USE_ANGELSCRIPT
 	ScriptEngine::getSingleton().SetFrameListener(gEnv->frameListener);
+#endif
 
[...]

PR #537, opened 2016-01-07
[Fix] Build without Angelscript

Hiradur:
* Fixes compiler errors when building without Angelscript (#534)
* Additionally, to avoid crashing when loading terrains, in resources/particles/water.particle this needs to be removed:
```
	affector FireExtinguisher
	{
		effectiveness 	1
	}
```

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -18,6 +18,7 @@ core requirements:
 optional but recommended:
 * angelscript: 2.22.1 
   * required for scripting (AI, racing, server mods...)
+  * when building without AS this has to be removed in resources/particles/water.particle: "affector FireExtinguisher {	effectiveness 	1 }"
 * caelum: >= 0.6.2, compatible with the OGRE version you chose 
   * sky plugin: provides dynamic sky with time of day, weather and clouds
 * mysocketw: latest from git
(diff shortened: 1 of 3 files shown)


Hubbitus:
It should be removed on build time only when building without angelscript? Why it does not touched in your commit?

Hiradur:
> It should be removed on build time only when building without angelscript? Why it does not touched in your commit?

Not sure how to do this since it's a custom file format used by Ogre, missing e.g. pre-processor macros

Hubbitus:
No-no. I ask it can't be deleted and commited? It should be deleted only when angelscript disabled (and then require some scripting)?

Hiradur:
> No-no. I ask it can't be deleted and commited? It should be deleted only when angelscript disabled (and then require some scripting)?

Maybe CMake could do that but I wouldn't know how. Research in that area would be welcome.

Hubbitus:
In other words, that's lines needed if angelscript support enabled?

Hiradur:
> In other words, that's lines needed if angelscript support enabled?

Probably not but a graphical effect would be missing.

Hubbitus:
Thanks for the explanation.

PR #543, opened 2016-01-08
Ogre 1.9.0 cmake fix

Speciesx:


Speciesx:
Doesn't work in this way. :(

PR #545, opened 2016-01-09
[Bugfix] Workaround to prevent crash during terrain loading

ulteq:
> How to reproduce this crash?

```mTerrainGroup->saveAllTerrains(false);``` tries to save the  ```mapbin``` file next to the RoR binary.

It will fail if the user that runs RoR does not have write access on that folder.

### DIFF
--- a/source/main/terrain/TerrainGeometryManager.cpp
+++ b/source/main/terrain/TerrainGeometryManager.cpp
@@ -189,7 +189,12 @@ void TerrainGeometryManager::initTerrain()
 		if (!disableCaching)
 		{
 			LoadingWindow::getSingleton().setProgress(23, _L("saving all terrain pages ..."));
-			mTerrainGroup->saveAllTerrains(false);
+			try
+			{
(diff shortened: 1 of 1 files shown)


Hiradur:
How to reproduce this crash?

only-a-ptr:
This isn't ```[bugfix]```, this is ```[uglyhack]``` :grimacing:

IIRC I saw ```mapbin``` in ```ROR_HOME/cache``` on windows - Is the described scenario linux-only? It's definitely a bug, ```mapbin``` belongs to cache.

ulteq:
> Is the described scenario linux-only?
Yes

only-a-ptr:
> <Ulteq> Yes

Then it's a linux mixbehavior. By nature, the mapbin is a cache file, so it belongs to /cache.
Anyway, IIRC you said in the chat the file isn't actually ever read from, just written down on every map load?

ulteq:
See: https://github.com/RigsOfRods/rigs-of-rods/pull/566

PR #551, opened 2016-01-09
Add system include and library directories for Paged Geometry

jeremybennett:
There seems to be an assumption that Paged Geometry will be built from
	source and installed in `/usr/local`.  However it can now be installed
	from standard repositories.  So we need to ensure that it is also
	searched for in `/usr`.

### Steps to reproduce
1. On Fedora 23 Install ogre-pagedgeometry-devel from the standard repositories
2. Build RoR following the instructions on http://www.rigsofrods.com/wiki/pages/Compiling_Sources_under_Linux#Building_the_source_code up to the point where ''cmake'' should be run.
3. Run `cmake .`

### Expected behaviour
''cmake'' configures RoR with Paged Geometry support.

### Actual behaviour
''cmake'' fails to detect that the Paged Geometry library is installed.

### System configuration
Fedora 23 64bit with Radeon R9 R390 and default open source device driver. Output of `uname -a`:

    Linux wotan.home.gateway 4.2.8-300.fc23.x86_64 #1 SMP Tue Dec 15 16:49:06 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux

### Additional information, logs and screenshots (optional)

The problem is a failure to look for headers in `/usr/include` and libraries in either `/usr/lib` (for 32-bit) or `/usr/lib64` (for 64-bit). I have suggested adding these as candidate paths for ''cmake''.  However I am a ''cmake'' novice, and there may be much better ways of solving the problem.

ChangeLog entry

	* CMakeDependenciesConfig.txt: Add system include path and library
	search paths for Paged Geometry.

### DIFF
--- a/CMakeDependenciesConfig.txt
+++ b/CMakeDependenciesConfig.txt
@@ -20,7 +20,7 @@ IF(WIN32)
 macro(importLib libName dirName)
 	message(STATUS "adding imported lib: ${libName} : ${ROR_DEPENDENCIES_DIR}/libs/${ARCH_DIR}/${dirName}/Release/${libName}.lib")
 	add_library(${libName} STATIC IMPORTED)
-	
+
 	# default is release mode:
 	SET_PROPERTY(TARGET ${libName} PROPERTY  IMPORTED_LOCATION ${ROR_DEPENDENCIES_DIR}/libs/${ARCH_DIR}/${dirName}/Release/${libName}.lib)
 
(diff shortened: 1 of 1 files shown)


Hiradur:
I don't know of any better way to do this but maybe @mikadou does.

mikadou:
It would be much nicer to handle the detection of third party libs with separate *find modules*. I'm currently experimenting with this. I hope to handle dependecies on Windows and Linux in a unified way, currently it is just two completely separate procedures.

For now, I think the provided patch is a good solution. Probably a little better would be the following, which I guess should also work (not testet though):

    find_path(PAGED_INCLUDE_DIRS "PagedGeometry/PagedGeometry.h" PATH_SUFFIXES OGRE)

    find_library(PAGED_LIBRARIES "PagedGeometry" PATH_SUFFIXES OGRE)

However, I think the patch in its current state is good enough.

PR #553, opened 2016-01-10
[Codechange] Clearing out and reordering of the node_t members

ulteq:
Edit: I'll do separate pull requests for the ```beam_t``` and ```node_t``` structs.

**Before**: sizeof(node_t) = **240**
**After**: sizeof(node_t') = **128**

The ```node_t``` struct now fits into 2 cache lines.

### DIFF
--- a/source/main/datatypes/beam_t.h
+++ b/source/main/datatypes/beam_t.h
@@ -14,18 +14,17 @@
 */
 struct beam_t
 {
-	Ogre::SceneNode *mSceneNode; //!< visual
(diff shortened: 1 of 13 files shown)


Hiradur:
How does it affect performance? I think they were more or less optimized for cache locality before.

ulteq:
I wasn't able to measure any performance difference so far.

ulteq:
Edit: I'll do separate pull requests for the beam_t and node_t structs.

PR #556, opened 2016-01-11
[Codechange] Clearing out and reordering of the beam_t members 

ulteq:
**Before**: sizeof(beam_t) = **208**
**After**: sizeof(beam_t) = **164**

All frequently accessed data is within the first 64 Bytes of the struct.

### DIFF
--- a/source/main/datatypes/beam_t.h
+++ b/source/main/datatypes/beam_t.h
@@ -18,54 +18,55 @@ struct beam_t
 {
 	node_t *p1;
 	node_t *p2;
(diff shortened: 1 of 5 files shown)


ulteq:
This can't happen, because ```shorter != 0``` implies ```shtruck != 0```.
https://github.com/ulteq/rigs-of-rods/blob/beam_t_Reordering/source/main/physics/Beam.cpp#L4363

But you are right, I will change it into ```it->beam->p2truck = (shtruck != 0);```. It's much cleaner that way.

PR #557, opened 2016-01-11
[Codechange] Clearing out and reordering of the node_t members

ulteq:
**Before**: sizeof(node_t) = **240**
**After**: sizeof(node_t) = **128**

All frequently accessed data is within the first 64 Bytes of the struct.


ulteq:
> was step&7 a typo here?

No, it was a speed-optimized variant of ```!(i%8)```.

> Can this be removed?

Probably yes, the performance impact is small. But I'm open to suggestions.

PR #561, opened 2016-01-12
[Bugfix] Fixes an input.map conflict (boat steering)

### LINKED ISSUE #472: Throttling backwards with boats resets steering direction
The steering wheel image and the steering direction are reset when pressing the backwards throttle button in boats. Makes manoeuvring backwards really cumbersome.

ulteq:
Fixes #472


Hiradur:
:+1:

only-a-ptr:
:boat:

PR #562, opened 2016-01-12
[Bugfix] Prevents boats from spawning underwater (#147)

### LINKED ISSUE #147: Boats spawn underwater
Boats that are spawned via the port building are spawned underwater, the same thing happens with boats that are placed on terrain. Boats that are underwater get deformed, some can be fixed with backspace, some cannot.
Tried with North St. Helens, Island, Starling Island.

ulteq:
Fixes #147


Hiradur:
:+1:

tritonas00:
Nice!

only-a-ptr:
:+1:

PR #563, opened 2016-01-12
[Bugfix] Wheel axes orientation no longer depends on the spawn rotation.

### LINKED ISSUE #33: Wheels And Gears Backwards On trucks Spawned in Truckshops Rotated Over 90° On The Y-Axis
Posted by Aviar in the forum's improvised bug tracker thread:
http://www.rigsofrods.com/threads/111967-Bugtracker-users-we-need-your-help-to-collect-all-bugs!?p=1363187&viewfull=1#post1363187

I have also had this issue happen once in a blue moon when spawning using the drop-down spawner, but I can't reproduce it.

ulteq:
Fixes #33


only-a-ptr:
:+1:

Issue #447, opened 2015-11-05
Otter.airplane spawn error


only-a-ptr:
http://www.rigsofrods.com/threads/114853-Install-scripts-for-latest-snapshot-for-Debian-Ubuntu-Mint?p=1388123&viewfull=1#post1388123

only-a-ptr:
It spawns well in the current upstream code.

Hiradur:
Spawns under the surface on Auriga Proving Grounds with OpenGL for me. Heavy physic glitches afterwards. Works fine on simple2 terrain.

only-a-ptr:
@Hiradur Thanks, I'll check it out

Hiradur:
I think the problem is that when you spawn the plane on a mesh it seems to spawn in the ground because only the heightmap is taken into account for spawn height.

PR #565, opened 2016-01-13
[Bugfix] Fixes Otter.airplane spawn error (#447)

ulteq:
The truck creation from the in-game menu now respects the lowest node.


only-a-ptr:
:+1:

PR #566, opened 2016-01-13
[Codechange] Disables terrain caching on linux

ulteq:
The mapbin file should be written into the ~/.rigsofrods/cache folder, but it's not

The ResourceGroupManager is not able to find the cache file


only-a-ptr:
Good.

PR #578, opened 2016-01-14
[Bugfix] Fixes crash when entering the skeletonview on a coupled train

ulteq:
Prevents asynchronous calls to updateSimpleSkeleton()


ulteq:
> the skeletonview could use some cleanup

Agreed, but in a separate pull request.

only-a-ptr:
> Agreed, but in a separate pull request.
OK.

PR #579, opened 2016-01-14
[Feature] Sort GUI Selector search results by relevance

ulteq:
The search results are now arranged according to the position of the first search hit.

And additionally ranked by relevance in the following way:
* name
* filename
* description
* author name
* author email
* alphabetically

Example search: **Tra**

**Before:**
1. Muddy Russian Tracks
2. Train Valley

**After:**
1. Train Valley
2. Muddy Russian Tracks

### DIFF
--- a/source/main/gui/panels/GUI_MainSelector.cpp
+++ b/source/main/gui/panels/GUI_MainSelector.cpp
@@ -39,7 +39,6 @@
 
 #include <MyGUI.h>
 
-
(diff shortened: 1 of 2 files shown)


disloyalpick:
Would be also nice for there to be a section on the selection menu for terrains/vehicles that shows the most frequently used item. Just a thought

only-a-ptr:
@ulteq :+1:

PR #585, opened 2016-01-15
[Bugfix] Fixes cabs parsing on IHwood trailer

ulteq:
Allows ```texcoords``` and ```cabs``` section parsing outside of ```submesh``` sections.


only-a-ptr:
The more compatibility, the better :+1:

PR #595, opened 2016-01-16
[Bugfix] Fixes shock stability active suspension for right side

### LINKED ISSUE #442: Air_Purge triggered from Shock stability option L or R
I've tried using the stability options (l/r) for Shocks in the past and I have always gotten this same issue. If the vehicle has a sound to be played by the air_purge trigger, it plays whenever the vehicle rocks left to right.
I'm not sure if this is a glitch, unfinished feature, or what, but I don't think this is the intended operation of that sound trigger.
Plus stability seems to have no impact on handling.
https://www.youtube.com/watch?v=FWEDlsFNySs&feature=youtu.be

I have thought of putting a creak sound as air_purge to make it sound like the suspension is creaking, but the trigger stutters way too much for that to work.

ulteq:
Implements this: https://github.com/RigsOfRods/rigs-of-rods/issues/442#issuecomment-153393226

Fixes #442


only-a-ptr:
:smiley:

PR #600, opened 2016-01-17
[Cleanup] Indentation (tabs, spaces) cleanup

ulteq:


Hiradur:
I thought we decided to use 4 spaces instead of tabs or was it the other way around? :blush:

ulteq:
Yes, but before we pull the lever it should at least be consistent.

Issue #171, opened 2015-05-05
Misc. spawn issues


TheUnknownGuyFromRoR:
![screenshot_1](https://cloud.githubusercontent.com/assets/10696803/7474871/6328ddd2-f344-11e4-8477-54ecca3cae60.png)
![screenshot_2](https://cloud.githubusercontent.com/assets/10696803/7474872/63319c56-f344-11e4-8379-ed3f94a4653f.png)

Max98:
It would be great if you gave us a link to the car.

TheUnknownGuyFromRoR:
http://www.rigsofrods.com/threads/98617-Lotus-Esprit-S1-With-rather-special-accessories-*DO-NOT-MOD*

only-a-ptr:
Oops, my bad. Will fix soon.

only-a-ptr:
This car is tricky. It combines buoyancy properties with truck properties, which is a dangerous mixture. I spent quite a while trying to fix it, but it's still not working.

Nonetheless, my debug tools keep improving. Eventually I'll figure this out.

tritonas00:
Link for testing: http://www.mediafire.com/file/ka4uruepf2ko4uo/LotusEsprit_JB.zip

Michael10055:
This vehicle crashes on Windows due to #1058:
`18:28:41: FLEXBODY ERROR on mesh LotusEsprit_rudderL.mesh: VY node not found`

ulteq:
It seems to be a parser issue. The new parser handles flexbody forsets differently.

The old code looked like this: https://github.com/only-a-ptr/ror-legacy-svn-trunk/blob/master/source/main/physics/flex/FlexBody.cpp#L52-L76

This affects all trucks which use `nodes2` and flexbody forset.

Michael10055:
Solution found, see this post: https://forum.rigsofrods.org/thread-1602-post-12866.html#pid12866
I'm going to leave this open though since there might be a way to change how the parser handles this .

tritonas00:
http://www.austingratzer.com/rigs/downloads.php?do=file&id=303

not anymore

![ep](https://user-images.githubusercontent.com/2660424/50435897-f1e27e00-08eb-11e9-9b24-55a52d8a93fe.png)

tritonas00:
@only-a-ptr is there anything that can be done in the new parser to fix this? (it worked in the old one)

if not just close the issue i guess

only-a-ptr:
[...]

PR #605, opened 2016-01-17
[Codechange] Changes parsing of very large numbers

ulteq:
Almost fixes: #171

Before we merge this, we should take a detailed look at what the old parser did in this case.

**Edit:** I think we should not merge this. It's an ugly workaround.


ulteq:
I think we should not merge this. It's an ugly workaround.

Issue #468, opened 2015-12-06
Error creating "mini" picture


Bob-Z:
After clearing cache, I launch RoR and I see the following traces

```C++
Adding archive /home/bobz/.rigsofrods/packs/Boats/HouseBoat.zip (hash: B6C5B59014316784A3D16F59E62433E864A5E87B) (No Compression)
Loading /home/bobz/.rigsofrods/packs/Boats/HouseBoat.zip
Creating resource group General-21
Added resource location '/home/bobz/.rigsofrods/packs/Boats/HouseBoat.zip' of type 'Zip' to resource group 'General-21'
Initialising resource group General-21
Parsing scripts for resource group General-21
Parsing script 4161UID-houseboat.material
Finished parsing scripts for resource group General-21
Creating resources for group General-21
All done
Preparing to add 4161UID-houseboat.boat
 -4161UID-houseboat.boat ...
error deleting file '4161UID-houseboat-mini.dds'
Unloading /home/bobz/.rigsofrods/packs/Boats/HouseBoat.zip
Removed resource location /home/bobz/.rigsofrods/packs/Boats/HouseBoat.zip
Clearing resource group General-21
Finished clearing resource group General-21
Unloading resource group General-21
Finished unloading resource group General-21
Destroying resource group General-21
Unloading resource group General-21
Finished unloading resource group General-21
```

Note the "error deleting file '4161UID-houseboat-mini.dds' " log.
When RoR has finished loading, I can't see any mini pics (either for maps or vehicule). And, indeed, the "cache" directory only contains "mods.cache" file.

After a bit of investigation I found that the following code cause this issue:

file: main/resources/CacheSystem.cpp
method: void CacheSystem::generateFileCache(CacheEntry &entry, Ogre::String directory)

```C++
                 FileInfoListPtr files = ResourceGroupManager::getSingleton().findResourceFileInfo(group, minifn);
                 if (files->empty())
                 {
                         deleteFileCache(const_cast<char*>(dst.c_str()));
                         return;
                 }
```

"files->empty" seems to be always true so the function returns before creating the mini.
Commenting out these lines fixes the issue.

I didn't make a pull request on this because I don't understand why "files->empty" is always true.

Some more info:
Built with Hiradur's latest scripts.
64 bits system

Hiradur:
@only-a-ptr The lovely Ogre resource management system is the problem here, can you comment on this?

Bob-Z:
Maybe related, maybe not:

[...]

PR #606, opened 2016-01-17
[Bugfix] 'Fixes' the mini picture generation on Ogre 1.9

ulteq:
Fixes: #468

Is there any reason not to fix it like that?


Hiradur:
We need @tritonas00 to test if this works with Ogre 1.8

tritonas00:
> We need @tritonas00 to test if this works with Ogre 1.8

Works fine with/without this.

Issue #608, opened 2016-01-17
Does not compile on ARM


Hubbitus:
```
/builddir/build/BUILD/rigs-of-rods-23c619a9010e375ea7357ef190ff683f2beac5d4/source/main/physics/BeamFactory.cpp:71:23: error: impossible constraint in 'asm'
    : "a" (i), "c" (0));
                       ^
source/main/main_sim/CMakeFiles/RoR.dir/build.make:3902: recipe for target 'source/main/main_sim/CMakeFiles/RoR.dir/__/physics/BeamFactory.cpp.o' failed
make[2]: *** [source/main/main_sim/CMakeFiles/RoR.dir/__/physics/BeamFactory.cpp.o] Error 1
```

Full log: https://kojipkgs.fedoraproject.org//work/tasks/6906/12586906/build.log

Hiradur:
I compiled RoR on ARM before and it worked, this is a problem with the CPU core cound detection recently introduced by @ulteq 

Code here: https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/physics/BeamFactory.cpp#L66

Instead of ```#else```  we need ```#elif defined(_X86_) || defined(__i386__) || defined(__386__)```and another ```#else``` path for non-x86 CPUs.

Hubbitus:
Indeed it compiles now: http://koji.fedoraproject.org/koji/taskinfo?taskID=12667538
Thank you.

PR #610, opened 2016-01-18
[Codechange] Attempt to fix compilation on ARM

ulteq:
Should fix: #608


Hiradur:
I see that ```__i386``` should be enough for non-VS compilers:
http://nadeausoftware.com/articles/2012/02/c_c_tip_how_detect_processor_type_using_compiler_predefined_macros#x86andx8664

PR #612, opened 2016-01-18
[Bugfix] Make proper use of the 'set_beam_defaults' visual beam diameter

ulteq:
The old parser has always used the visual beam diameter set via the 'set_beam_defaults' command, not the default value.


Hiradur:
:+1:

PR #618, opened 2016-01-19
[Discussion] main loop: mainthread vs. simthread updates.

only-a-ptr:
*Discussion PR, do not merge yet!*

While working on [LuaPowertrain](http://www.rigsofrods.com/threads/120811-Dev-LuaPowertrain-project), I got blocked out by RoR's architecture - the engine/brakes/otherstuff are updated from both mainthread (gameplay/LandVehicleSimulation.cpp) and simthread (gameplay/BeamEngine.cpp) without proper synchronization. By coincidence, it always worked fine, but it can't stay that way. For one thing, it blows up when any scripting is added.

I refactored the code towards this architecture:
 * Mainthread reads and buffers all user inputs. Also updates GUI+visuals from vehicle state snapshot.
 * Simthread reads the input buffer and performs updates. At the end, it takes a snapshot of vehicle state.
 * [sync] Input buffer is passed to simthread, snapshot is passed to mainthread.

Now there's a question: what next?
 * Stabilize the code as it is and implement Lua scripting of the powertrain.
 * Continue the redesign and make simthread exclusively responsible for ALL vehicle updates. Powertrain took care of most, but there's also steering, commands, hooks...

@Speciesx I need you to build and test this. I think the vehicles behave differently.
@ulteq I need you to review this.

Branch: https://github.com/only-a-ptr/rigs-of-rods/tree/simulation-powertrain-MT-rev2

### DIFF
--- a/source/main/gameplay/Powertrain.h
+++ b/source/main/gameplay/Powertrain.h
@@ -8,31 +8,12 @@
 #pragma once
 
 #include "ForwardDeclarations.h"
+#include "transmission.h"
 
 #include <vector>
 
 namespace RoR
 {
-
-struct Gearbox
-{
-    enum shiftmodes {
-        SHIFTMODE_AUTOMATIC,
-        SHIFTMODE_SEMIAUTO,
-        SHIFTMODE_MANUAL,
-        SHIFTMODE_MANUAL_STICK,
-        SHIFTMODE_MANUAL_RANGES
-    };
-
-    enum autoswitch {
-        AUTOSWITCH_REAR,
-        AUTOSWITCH_NEUTRAL,
-        AUTOSWITCH_DRIVE,
-        AUTOSWITCH_TWO,
-        AUTOSWITCH_ONE,
-        AUTOSWITCH_MANUALMODE
-    };
-};
 
 /// Powertrain state
 /// Each member replaces a member of legacy class BeamEngine (referenced in doc comments).
@@ -99,6 +80,13 @@ struct PowertrainState
 
     /// Precomputed ```int getNumGearsRanges() { return getNumGears() / 6 + 1; };```
     int conf_transmission_num_gears_ranges;
+
+    // Simthread -> mainthread updates
+
+    bool transmission_shiftmodes_toggled;
+    int  transmission_gear_range_selected; // 0=no change, 1=lowrange, 2=midrange, 3=highrange
+    int  transmission_axlelock_toggled; // 0 = no change, -1 = `no differentials` console message, 1 = toggled, show console message.
+
 };
 
 struct PowertrainCommand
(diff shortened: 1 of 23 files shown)


source/main/gameplay/Powertrain.h
```diff
@@ -0,0 +1,198 @@
...
+/** 
+    @file
+    @date   01/2016
+    @author Petr Ohlidal
+*/
+
+#pragma once
+
+#include "ForwardDeclarations.h"
+#include "transmission.h"
```
Speciesx: #include "datatypes/transmission.h"

shouldn't be this?

only-a-ptr: @Speciesx Could be, but it's a detail.

Speciesx:
I've noticed so far:
- In automatic shift gearbox vehicles start without starter, switch on ignition and press accelerator 
- Arcade controls are active when start the engine in non automatic shift gearbox
- Switch between 4k and 8k Speedometer doesn't work

only-a-ptr:
NOTE TO ALL:
Do not merge this. It's only here to attract discussion (it's not working, anyway...)

When we decide what to do about it next, I'll create a new branch from current upstream and rebase+polish the commits.

only-a-ptr:
Decided in dev chat:
now that there's been so much progress lately based around the current upstream, I agree with @mikadou : Let's first stabilize upstream, make a release and then start restructuring hard.

>  There exist better means for navigating code
@mikadou Show me.

mikadou:
@only-a-ptr I'm using rtags on Linux. I might be wrong, but I'm pretty sure  visualstudio also has some kind of 'find usages' feature built-in

only-a-ptr:
@mikadou VisualStudio's "find all references" sucks - it's very slow and it's unreliable - it's practically just a dumb fulltext searcher (it shows all symbols of the same name) but some results get left out. Unusable.

I agree the repetitive function names aren't elegant, but they're so damn useful! Any other way of reference searching is technically super-complex and requires extra utilities. Personally I think the very reason for programming languages still being text-based is text indexing and plaintext searching. In the company I work for, the most valuable tool is [OpenGrok](https://opengrok.github.io/OpenGrok/) server.

mikadou:
I see. I wasn't aware that VisualStudio's search is not that sophisticated. Thanks for telling me about the OpenGrok tool. Looks very interesting. I'm curious, does OpenGrok provide context-aware search? Or is it also basically also a dumb fulltext searcher?

only-a-ptr:
@mikadou OpenGrok understands code and uses the data to cross-link the documets and to annotate results, but not for searching itself.
These are supported search fields: 
Full Search - fulltext	
Definition - find struct/variable/function definition by symbol name
Symbol 	- basically case-sensitive fulltext
File Path 	- filters results
History 	- dunno what this does

PR #576, opened 2016-01-14
Moved MainPage.h

AnotherFoxGuy:
Moved the MainPage file to its correct place :smiley:
Also fixed some links in the file

PR #621, opened 2016-01-20
[Add] New Doxygen config. Integrated with CMake build system.

mikadou:
The file location `doc/doxygen/MainPage.h` has already been added to inputs in the config file (See https://github.com/RigsOfRods/rigs-of-rods/pull/576)


only-a-ptr:
:+1:

PR #626, opened 2016-01-20
[Feature] New RoR.cfg option for high resolution (wheel-)node collisions

ulteq:
Greatly improves the overall behavior of the wheel <-> ground surface interaction.
* This only affects wheel nodes
* You can enable this by appending ```HighResWheelNodeCollisions=Yes``` to your ```RoR.cfg```.

This could become a new parameter for truck files.


Hiradur:
:+1:

PR #628, opened 2016-01-21
[Codechange] Revamped the braking code

ulteq:
Brakes in RoR apply torque against the direction of rotation of each wheel:
https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/physics/BeamForcesEuler.cpp#L729

However the rotary speed has no effect on the amount of torque which is applied. This is changed by this PR. It will fix the following issue:

* Greatly reduces brake induced skidding in parking positions.
Best example: [Bobcat Skid-Steer](http://www.rigsofrods.com/repository/view/1962)

* Prevents brakes from making wheels spin backwards.
Best example: [1974 Lotus 76/1](http://www.rigsofrods.com/threads/118984-Lotus-76-1-for-current-and-future-ror-versions-(Updated-for-test-build-4))

* Removes the (now unnecessary) [slopebrake](http://www.rigsofrods.com/wiki/pages/Truck_Description_File/#SlopeBrake).

* Fixes the anti-lock brake and traction control pulse modulation.

### DIFF
--- a/source/main/datatypes/rig_t.h
+++ b/source/main/datatypes/rig_t.h
@@ -146,7 +146,7 @@ struct rig_t
 	float alb_ratio;        //!< Anti-lock brake attribute: Regulating force
 	float alb_minspeed;     //!< Anti-lock brake attribute;
 	int alb_mode;           //!< Anti-lock brake status; Enabled? {1/0}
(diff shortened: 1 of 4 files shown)


Hiradur:
:+1:

only-a-ptr:
:+1:

PR #633, opened 2016-01-21
[Codechange] Revamped the auto clutch code

ulteq:
* Fixes cruise control conflicting with the auto transmission (```autocurAcc``` was not set correctly)
* Moves the ```is_Eletric``` engine type check outside of the actual shifting code

* Makes the auto clutch release much more realistic:
It now depends on the speed difference, clutch force and current engine power


Hiradur:
:+1:

ulteq:
I have the commits restructured in order to make clear what was actually changed.

Issue #634, opened 2016-01-22
Latest master compile error with ogre 1.8


tritonas00:
Two days ago compiled fine but today i get http://pastebin.com/HLkY165X

Speciesx:
Compiling fine on Windows.

Hiradur:
no problems here
```gcc version 5.2.1 20151010 (Ubuntu 5.2.1-22ubuntu2) ```

Travis (gcc 4.8) also had no problem

**Comment by tritonas00** [comment]
I tried from aur which uses ogre 1.9 and it builds fine. So an ogre 1.8 thing?

**Comment by tritonas00** [comment]
Ok the problem is in this commit https://github.com/RigsOfRods/rigs-of-rods/commit/d963b6d417e6adc00f928d863e0d8b0b822064b0

Removed lines from 679 to 720 here https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/gameplay/RoRFrameListener.cpp#L679 and compiles again.

**Comment by Hiradur** [comment]
try ```String ssmsg = _L("New simulation speed: ") + TOSTRING(static_cast<float>(floor(mLastSimulationSpeed * 1000.0f) / 10.0f)) + "%";```

**Comment by tritonas00** [comment]
Again with this https://github.com/RigsOfRods/rigs-of-rods/commit/6d7bdee946784158a36ab1af761366a502421604 i get

> /home/babis/Downloads/rigsofrods-next-build/rigsofrods/src/rigs-of-rods/source/configurator/Configurator.cpp: In function ‘void initLanguage(wxString, wxString)’:
/home/babis/Downloads/rigsofrods-next-build/rigsofrods/src/rigs-of-rods/source/configurator/Configurator.cpp:555:58: error: no matching function for call to ‘Ogre::ImprovedConfigFile::getSetting(const char [9])’
   wxString langSavedName = conv(cfg.getSetting("Language"));

**Comment by AnotherFoxGuy** [comment]
Also hapens on windows 
```
D:\Jenkins\workspace\rigs-of-rods\source\configurator\Configurator.cpp(555): error C2661: 'RoR::ConfigFile::getSetting' : no overloaded function takes 1 arguments
```
full log: http://pastebin.com/raw/iG8TzCUQ

tritonas00:
fixed in https://github.com/RigsOfRods/rigs-of-rods/pull/704

PR #641, opened 2016-01-25
[Bugfix] Fixes calling toString(double) being called on Ogre 1.8

ulteq:
Made use of the Round function in the Utils.cpp

https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/utils/Utils.cpp#L319

Should fix: #634


ulteq:
@tritonas00 Does that fix your issue?

tritonas00:
Tried just now, yes, and also simulation speed control works fine! Thank you :+1:

PR #644, opened 2016-01-26
[Bugfix] Prevents brakes from unlocking if you step on the gas

ulteq:
The brakes are already unlocked in the LandVehicleSimulation logic.


Hiradur:
:+1:

PR #647, opened 2016-01-26
[Bugfix] Attempt to fix the (high) pitch of the turbo sound

ulteq:


ulteq:
It's more likely an issue with the turbo sound file or config.

Issue #631, opened 2016-01-21
Duplicate sections in vehicle / load selector


Hotrod55:
Using the latest master, the "unknown" section in the vehicle / load selector is duplicated:
![2016-01-21_15-12-28](https://cloud.githubusercontent.com/assets/11002490/12494027/0f4c5952-c056-11e5-9541-c3a109581b12.png)
Clearing the cache does not help.

PR #648, opened 2016-01-27
[Bugfix] Removes duplicate 'unknown' categories from the MainSelector

ulteq:
Those 'unknown' cache entries will now show up in the 'Unsorted' category

Fixes: #631


Hotrod55:
:+1:

PR #652, opened 2016-01-27
[Bugfix] Prevents infinite selector window loop

ulteq:
```game.clearEventCache()``` is no longer needed: https://github.com/RigsOfRods/rigs-of-rods/commit/1319fd755431f6623fa97baec223ba47b3bb4c42


ulteq:
Does not work reliably on all maps.

Closing the selector window causes the character to jump due north for whatever reason.
Which can make the selector window pop up again.

Issue #643, opened 2016-01-26
Garage spawn points triggering


tritonas00:
Load nHelens (happens in all maps). Go to garage door and will open the vehicles menu. Press cancel, go in again. Now the menu is not opening. Go to another spawn point do the same and return to the first which now can be triggered again.

It happens with all spawners (vehicles/loads/planes/boats).

Example video: https://www.dropbox.com/s/k2jqtm178ts652i/garage.mp4?dl=0

Shouldn't this be triggered every time you get in?

Sorry for bad english

DirtGamer301:
IIRC that already happaned with 0.38, if that helps. I can't test it though,

tritonas00:
> IIRC that already happaned with 0.38, if that helps. I can't test it though

I think so, but i cant test because 0.38's angelsctipt is broken in linux amd64

DirtGamer301:
I can't test it either because my 0.38 is broken, it shoots me an error message when starting. Might be a mod, might be something else. But as I didn't use it for ages and kinda replaced it with 0.4 I didn't really bother fixing it.

PR #653, opened 2016-01-28
[Bugfix] Fixes event cache not being cleared on exit

ulteq:
More robust solution for #643

Reverts: https://github.com/ulteq/rigs-of-rods/commit/1319fd755431f6623fa97baec223ba47b3bb4c42


only-a-ptr:
Looks good.

PR #655, opened 2016-01-28
[Codechange] 'RShift + Space' disables both camera zoom and rotation limits

ulteq:
Rather than only allowing truck mesh clipping, we now also allow terrain clipping.

* ```RShift + Space``` is mainly used for taking screenshots (where you want arbitrary camera movement)


DirtGamer301:
That's awesome, I have been hoping for something like this since I used 0.4.0.7 :)

Hotrod55:
:+1:

Hotrod55:
Well, after testing this, the normal camera can clip the terrain but freecam doesn't.

PR #659, opened 2016-01-29
[Cleanup] Made the trigger blocker parsing more explicit

ulteq:


ulteq:
We better leave it exactly as it was: https://github.com/only-a-ptr/ror-legacy-svn-trunk/blob/master/source/main/physics/input_output/SerializedRig.cpp#L1919

only-a-ptr:
@ulteq Sorry for being slow to respond,
> We better leave it exactly as it was: https://github.com/only-a-ptr/ror-legacy-svn-trunk/blob/master/source/main/physics/input_output/SerializedRig.cpp#L1919

I'm afraid that link is obsolete. The last state of SerializedRig.cpp (v0.4.0.7) is this: We better leave it exactly as it was: https://github.com/only-a-ptr/ror-legacy-svn-trunk/blob/master/source/main/physics/input_output/SerializedRig.cpp#L1919

PR #676, opened 2016-01-31
[Bugfix] Prevents segfault on exit

ulteq:
Needs to be tested on Windows.


Hiradur:
I think the line can be removed, I tried to delete the overlay system at earlier places in the code and even directly after the initialization of the overlay system it causes a segfault with the same stacktrace. Not sure if we are supposed to call it manually, it doesn't show up in tutorials I've seen so far.

PR #680, opened 2016-01-31
[Codechange] lockNodes should not be changed outside of the truck parser

ulteq:
This might fix Bug #: 1013: http://www.rigsofrods.com/threads/111967-Bugtracker-users-we-need-your-help-to-collect-all-bugs!?p=1307247&viewfull=1#post1307247

**Edit:** Bug 1013 must be caused by sth. else.


only-a-ptr:
:+1:

Issue #465, opened 2015-12-03
Ties in .4.5 extremely fragile.


DarthCainRor:
The ties used to tie objects down in RoR, appear to be super fragile in .4.5 as compared to .38.

Map: any, but I used simple map for last test.
Vehicles: any trucks, trailers, and loaded vehicles. I tested with this:
     default Dodge Ram http://www.rigsofrods.com/repository/view/1717
     Squishy skidsteer (mine only differs on the wheel mesh and texture) http://www.rigsofrods.com/repository/view/1962
trailer: Interstate 8st  http://www.rigsofrods.com/threads/100457-Interstate-Trailer-Pack 

When loading the bobcat, almost immediately after hitting "O" to tie it down, the rear ties break. Shortly after another breaks, and the last usually breaks after driving around. 

Log:http://pastebin.com/5giiMu3S
Two ties already broke:
![screenshot_227](https://cloud.githubusercontent.com/assets/10333582/11556695/1754d00c-996e-11e5-83ea-810982696dc4.png)
one more broken:
![screenshot_228](https://cloud.githubusercontent.com/assets/10333582/11556698/1c4aec5e-996e-11e5-82fc-9bd9ef1521d3.png)
after pretty normal driving the last one broke:
![screenshot_229](https://cloud.githubusercontent.com/assets/10333582/11556701/230b4d4a-996e-11e5-8b38-fc243ce4f4b3.png)

Same combination works fine in the old stable versions.

only-a-ptr:
Looks like something I'm mis-calculating in the spawner. The spawn process does a *lot* of complicated post-processing and I don't understand half of it :worried:

ulteq:
> Looks like something I'm mis-calculating in the spawner. 

The ```ParseTies``` function is not the issue. I've just verified that it correctly parses the tie parameters.

References:
https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/rig_file_input_output/RigDef_Parser.cpp#L4109-L4160
http://www.rigsofrods.com/wiki/pages/Truck_Description_File/#Ties

ulteq:
> Looks like something I'm mis-calculating in the spawner.

The ```ProcessRopable``` function is not the issue. I've just verified that it parses the ropable parameters correctly.

References:
https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/physics/input_output/RigSpawner.cpp#L2924-L2934
[...]

Issue #189, opened 2015-05-18
Hooks crashing the game


tgp1994:
> The bug occurs in relation to hooks on a vehicle. When multiple objects with hooks are interacting with each other, the game crashes at some point. Please notice the Debug information attached [here](http://www.rigsofrods.com/attachment.php?attachmentid=470370&d=1395352480).
**How to reproduce:** Stack some containers with the Reachstacker. As the bug only occurs occasionally, it might be hard to reproduce. The problem seems to occur more often with a bigger amount of containers stacked.
**Content needed:** http://www.rigsofrods.com/repository/view/5294, http://www.rigsofrods.com/repository/view/4251

*This was moved from the wiki bugtracker. [Original post.](http://www.rigsofrods.com/threads/111967-Bugtracker-users-we-need-your-help-to-collect-all-bugs!?p=1307247&viewfull=1#post1307247)*

PR #681, opened 2016-01-31
[Bugfix] Removes race condition in calcBeams

ulteq:
Prevents unsynchronized read/write access to `p2->Forces`, when `p2truck != 0`.

Should fix: #465
Related with: #189


only-a-ptr:
:+1:

Issue #664, opened 2016-01-30
Game crash when going in setting ingame


WillyB1337:
Hello excellent developers!

I discovered that the famous "UTF-8 continuation byte error" happened to me when I'm using my second screen in 1080p by HDMI and my G27.

[mygui.txt](https://github.com/RigsOfRods/rigs-of-rods/files/110601/mygui.txt)

For me, the thing could come for the fact it cannot support a 1080p thing :|

WillyB1337:
Here's additional files for reference : 
[RoRConfig.txt](https://github.com/RigsOfRods/rigs-of-rods/files/110602/RoRConfig.txt)
[RoR.txt](https://github.com/RigsOfRods/rigs-of-rods/files/110603/RoR.txt)

DirtGamer301:
I can second this bug: https://github.com/RigsOfRods/rigs-of-rods/issues/237
I don't use a 2nd monitor, and not 1080p, too, so I think that shouldn't be a thing.

I just saw I closed my issue ticker, I'll reopen it because it seems like I keep getting "Bad UTF-8 continuation byte error", so it's not actually solved, I can't remember what might have happaned some time ago, and can't remember it being fixed either ??

only-a-ptr:
@WillyB1337 Yes the error is getting famous. The reason is known: https://github.com/RigsOfRods/rigs-of-rods/issues/237#issuecomment-177342715

Issue #455, opened 2015-11-19
Bad UTF-8 Continuation type while regenning CACHE


disloyalpick:
I'm not sure which vehicles or stuff is causing this, but I will delete contents of all the log files, and try to clear/regen CACHE and upload the log files for someone to check out.
![ss 2015-11-19 at 02 43 03](https://cloud.githubusercontent.com/assets/13305403/11282179/01cd793e-8ecc-11e5-84b2-02f3641e4bc6.png)

AngleScript.log: http://pastebin.com/9rb2hNFq

ConfigLog.txt: http://pastebin.com/2WgNXuYn

mygui.log: http://pastebin.com/cHSNU6Pc

RoRConfig.log: http://pastebin.com/xw13p3d4

RoR.log(Exceeded pastebin max size limit)   http://puu.sh/lrqiT/166c8fc19f.log

Thank you for taking time to look this over. Due to this I can't even play RoR. Good luck fixing it if you find any problem!!

Hiradur:
Could you check if caboverpete.truck has a similar problem to what is described here: https://github.com/RigsOfRods/rigs-of-rods/issues/364

disloyalpick:
It does

http://puu.sh/lrrfN/2f9f9ac073.png

DirtGamer301:
You can try removing the mail adress, but it's actually supposed to be there so I personally don't think it causes problems.

only-a-ptr:
"@" is an ASCII character, that's not the issue. The problem is somewhere else.

And yes, this is the same problem as https://github.com/RigsOfRods/rigs-of-rods/issues/364

I'm starting to see the issue: I used MyGUI's UStrings to handle loaded data. UStrings don't only store the data, but try to decode them as well.

only-a-ptr:
@Hiradur I reproduced the other issue.

I suggest incorporating the "utf8cpp" library (tiny, header-only) http://sourceforge.net/p/utfcpp/code/HEAD/tree/v2_0/source/, licensed with "boost license"  (OSI approved) https://tldrlegal.com/license/boost-software-license-1.0-explained. With this library, I can sanitize input from truckfiles, which are (by nature) saved in a variety of ANSI/OEM encodings.

mikadou:
[...]

Issue #385, opened 2015-08-12
Crash when going into ingame settings (Bad UTF-8 continuation Byte)


DirtGamer301:
So, I had it since ingame settings were being added and yet none of the advices helped.

The problem is simple to explain, but you most likely won't be able to reproduce it.
I press the "settings" tab and then the game crashed, giving me an error message: "Bad UTF-8 continuation Byte"
RoR.log: http://www.mediafire.com/download/fua4zd123lh5qvo/RoR.log

Cache clearing, regenerating, deleting whole RoR 0.4 folder and running RoRConfig, deleting the stuff in RoR 0.4 folder and running RoRConfig didn't help.

Hiradur:
Is 'bandit' your real Windows user name? What system locale do you use? Perhaps there is a setting in RoR.cfg using an Umlaut or other non-ASCII characters?

DirtGamer301:
Yes, from "Smokey and the bandit" ^^ Couldn't think of anything better

I installed all RoR versions on C.

I found this in RoR,cfg:
Program Path=c:\progra~2\rigsof~1\
Replay mode=Yes
Resources Path=c:\progra~2\rigsof~1\resources\

And this:
plugins.cfg=C:\PROGRA~2\RIGSOF~1\plugins.cfg

It btw says it's a runtime error so I did what I always did when having a runtime error, too, but it didn't help either.

Hotrod55:
This same error also happens in .truck files: #364

Hiradur:
@DirtGamer301 Is that the correct log? It looks like a normal session to me without a crash.

DirtGamer301:
Oh, are there any other information beside gameplay in RoR.log?
Idk, but maybe this RoR.log was created the session before and as I didn't play to get this log (only pressed on the "Settings" tab) it might not be created.
I'd then have to upload another file, but which one?

Hiradur:
upload the whole logs folder, maybe there is something useful in one of the logs.

DirtGamer301:
I'll do so tomorrow, I sadly have nomore time today.

DirtGamer301:
There we go, I zipped the folder:
http://www.mediafire.com/download/pq344lrap5e5uep/logs.zip

Hiradur:
hmm, I couldn't see anything suspicious. Can you upload your config folder, please?

DirtGamer301:
If a video helps I can record one.
[...]

Issue #364, opened 2015-08-07
Can not load vehicles with special characters in authors section 


Hotrod55:
NothingToDo posted this in the RoR Skype group, I thought I'd post a bug report on it since I found the cause of it.
 
Using this vehicle: http://www.rigsofrods.com/repository/view/5548
The authors section has a special character in it, causing a runtime error to occur.
author Lampe fl**è**ches -1 dolmen62
![2015-08-07_08-50-42](https://cloud.githubusercontent.com/assets/11002490/9136154/53683db0-3ce2-11e5-9234-ae57c3effd76.png)
RoR.log says the parser ignores the line:
```
08:50:22:  ===== LOADING VEHICLE: afc4UID-a4e3UIDsprinterasf.truck
08:50:22:  == Parsing vehicle file: afc4UID-a4e3UIDsprinterasf.truck
08:50:22:  == Parsing done, report:

#FF3300 ERROR #FFFFFF (Section globals)
	Line (11): author Lampe flèches -1 dolmen62
	Message: Invalid line, ignoring...
#FFFF00 WARNING #FFFFFF (Section meshwheels)
	Line (673): -36
	Message: Encountered node with illegal negative number: '-36', parsing as positive '36' for backwards compatibility. Please fix as soon as possible.



08:50:22:  == Validating vehicle: Sprinter ASF
08:50:22:  == Validating done OK
```
Does not happen in 0.4.0.7.

only-a-ptr:
@Hotrod55 Thanks for reporting.

only-a-ptr:
@Hotrod55 The problem is RoR assumes truckfiles saved in UTF-8 encoding. This truck is in https://en.wikipedia.org/wiki/Windows-1252. I discovered it by saving file to disk, opening it in firefox and then switching [Display->character set->*] options until it matched. There is no way to auto-detect it.

I'll figure something out tho.

only-a-ptr:
Reproduced. Fixing in progress....

only-a-ptr:
Fixed in upstream.

PR #683, opened 2016-01-31
[fix] Added utf-8 sanitization; resolves the inflamous "Bad UTF-8 continuation byte" error.

### LINKED ISSUE #237: Game crashes when opening the ingame settings
When opening the ingame settings, the game will crash. It'll just "stay" in the menu and give me this error:

http://i.imgur.com/iPj9M6H.png

Wich log should I add or just all?

only-a-ptr:
Added full-scale UTF-8 sanitization; no more crashes.
This resolves the inflamous "Bad UTF-8 continuation byte" error dialog. Fixes #237 #385 #664 #455 #364
Incorporated UTFCPP library: (http://utfcpp.sourceforge.net/; License: boost) and added UTF-8 sanitization to all text-file processing.

@InvalidUTF8character: YOU SHALL NOT PASS! :smile:


ulteq:
Compiles fine, runs fine, looks fine. :+1: 

But I can't test if it resolves those crashes, because I've never had them.

only-a-ptr:
@ulteq At least https://github.com/RigsOfRods/rigs-of-rods/issues/364 has to apply to you too. The truckfile contains non-utf8 characters in 'authors'.

Hiradur:
:+1: 
:name_badge: (boring work hero badge)

ulteq:
I'll test with #364

only-a-ptr:
@ulteq Interesting ... check if the following error appears in your spawner log window:
```
#FF3300 ERROR #FFFFFF (Section globals)
    Line (11): author Lampe flèches -1 dolmen62
    Message: Invalid line, ignoring...
```
The exception comes from MyGUI::UString which is used internally for GUI display. If the regexes don't identify the globals-line as mismatch, the error doesn't display and hence UString never meets the invalid chars.

BTW 4K resolution is sick :smile:

ulteq:
> check if the following error appears in your spawner log

Yes, it does.

only-a-ptr:
> Yes, it does.
Curious. Something's different in UString on your system. :penguin:

graysonk95:
Is there a workaround for this?

I'm stuck in an infinite loop of auto-regen>crash, auto-regen>crash, etc. 0.4.6.0-something, 0.4.7.0-RC1, and 0.4.8.0-NB-131-jdmgmuma all crash on regen with this error.
I tried manually regen'ing in 0.4.0.7, but 0.4.6-0.4.8 insist on doing their own regen, so I'm effectively locked out of all versions of 0.4 because they're all crashing for one reason or another.

only-a-ptr:
@graysonk95 Workaround for the fix? No, it's hardwired. Workaround for the UTF-8 encoding problem? Just fix the content which causes the crash.

If 0.4.7 or older version crashes with this error, file a bug including:
 * RoR.log
 * links to test content.
 * Is there a specific terrain/truck causing this for you?
 * What's your OS/system language? Does your userdir contain special characters? Does filenames of any of your mods contain any special characters?

graysonk95:
@only-a-ptr - The older 0.4 crashes in game because my current truck uses new features it does not like. Irrelevant to this issue.

The newer (and current) 0.4 builds crashed on cache regen seemingly for every single vehicle, RoR.log just started spitting our OGRE Exceptions for every vehicle.
I managed to fix it by removing all content in Documents\Rigs of Rods 0.4 and adding back everything in groups of about 100.
Where before 1 failed cache regen would take like 3-4 minutes, 8 smaller successful cache regens took about 15 seconds each.
I have 7GB of content installed (~800 files), so perhaps it just could not handle everything at once.
That was my workaround.

PR #698, opened 2016-02-02
[Codechange] Some changes to the parser ('cab', 'rigidifiers', 'turboprops2')

ulteq:
Truly skip unsupported `rigidifiers` and `turboprops2` sections.
* Less warnings when parsing the [1949 Ford](http://www.rigsofrods.com/repository/view/2579).

Interpret `submesh` sections without subsections as `cab` subsection.
* Fixes `cab` parsing on the [Audio 80 Quattro](http://www.mediafire.com/?tc927day8vat1q7).

### DIFF
--- a/source/rig_file_input_output/RigDef_Parser.cpp
+++ b/source/rig_file_input_output/RigDef_Parser.cpp
@@ -529,6 +529,7 @@ void Parser::ParseLine(Ogre::String const & line)
 
 			case (File::KEYWORD_RIGIDIFIERS):
 				AddMessage(line, Message::TYPE_WARNING, "Rigidifiers are not supported, ignoring...");
+				new_section = File::SECTION_NONE;
 				line_finished = true;
 				break;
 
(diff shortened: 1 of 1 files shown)


only-a-ptr:
rigidifiers+turboprops :+1: 
submesh/cab: I don't understand how this can fix any vehicle. Unless I missed something, cabs without keyword could never have worked. v0.4.0.7's parser would parse cabs [ONLY if keyword cabs was encountered](https://github.com/RigsOfRods/rigs-of-rods/blob/bd680916356ea033716c96582305a37347644b9a/source/main/physics/input_output/SerializedRig.cpp#L2767-L2812) Parser from v0.38.66 [has the same flow. ](https://github.com/only-a-ptr/ror-legacy-svn-trunk/blob/db0b4b922c0eb8de240ac720edbc3560e5cd1399/source/main/physics/input_output/SerializedRig.cpp#L2709-L2754). But if it improves things... :+1:

ulteq:
> I don't understand how this can fix any vehicle.

The Audi 80 has probably always had a few missing cabs.

> But if it improves things...

I don't know if we should allow missing subsections. It's your decision.

only-a-ptr:
> I don't know if we should allow missing subsections. It's your decision.

Generally, we shouldn't parse anything the 0.38/0407 parsers wouldn't. But let's make an exception this time, I think we're safe enough with the regex.

PR #704, opened 2016-02-02
[Codechange] Hotfix for the RoRConfig build

ulteq:


tritonas00:
Works :+1:

only-a-ptr:
:-1: I don't believe there's not a better solution. Let me code for a moment...

ulteq:
-> https://github.com/RigsOfRods/rigs-of-rods/pull/705

PR #710, opened 2016-02-03
[fix] Utf8: do not throw on incomplete last character.

only-a-ptr:
Rare issue, observed with: http://www.rigsofrods.com/threads/77547-Jantor-BigMachines


ulteq:
:+1: It does fix the crash during cache regeneration for me.

Hiradur:
:+1:

PR #714, opened 2016-02-04
[Bugfix] Fixes build errors with older GCC versions

ulteq:
The `std::asctime` output is less pretty, but good enough as fallback.


mikadou:
I agree, it's not a super pretty solution, but fixes the Travis builds.

PR #717, opened 2016-02-04
[Bugfix] Fixes set_beam_defaults parsing

ulteq:
The parser ignored `set_beam_default` lines which ended on a comma.

Example truck: [Burnside Taxi](http://gabefink.com/Burnside.zip)


only-a-ptr:
:+1: Now it'll accept ```set_beam_defaults``` with any invalid trailing string, too.

PR #719, opened 2016-02-05
[Codechange] Let the parser ignore any invalid trailing string

ulteq:
Example truck, where the parser failed: http://www.rigsofrods.com/repository/view/5449


only-a-ptr:
Welcome to my house of pain, Ulteq!

I thought about doing it this way, but there's a catch: If there are optional args in the line (```E_CAPTURE_OPTIONAL()``` is used) ,there's ```E_CAPTURE_OPTIONAL( E_ILLEGAL_TRAILING_STRING )``` and the line contains a syntax flaw in the optional arg, then bang: E_CAPTURE_OPTIONAL(opt-arg) fails but the E_CAPTURE_OPTIONAL( E_ILLEGAL_TRAILING_STRING ) eats it and the regex ends with "success" => and then it either crashes on data parsing (the good case) or it parses bogus data (the expectable case).

ulteq:
So we just revert the `set_beam_defaults` parsing to how it was before we allowed trailing comments.

only-a-ptr:
Let's make a fallback like this: https://github.com/RigsOfRods/rigs-of-rods/blob/10bb9768766a3fc9b2b314ff0b64f17a8526ef31/source/rig_file_input_output/RigDef_Parser.cpp#L51
The regex will still report a warning if the syntax sucks, but it won't reject old truckfiles.

ulteq:
> Let's make a fallback like this: https://github.com/RigsOfRods/rigs-of-rods/blob/10bb9768766a3fc9b2b314ff0b64f17a8526ef31/source/rig_file_input_output/RigDef_Parser.cpp#L51

Agreed.

PR #736, opened 2016-02-06
[Bugfix] Also reset the event timer when aborting the selector

ulteq:


ulteq:
-> https://github.com/RigsOfRods/rigs-of-rods/pull/738

PR #743, opened 2016-02-06
[Cleanup] Moved the skeleton coloring to a central place

ulteq:
This also changes the `stress` <--> `color` relationship.


Hiradur:
:+1:

PR #744, opened 2016-02-06
[Bugfix] Prevent selector window from popping up right after canceling

ulteq:
This completely solves the selector window issue (in combination with #738).

Related with: https://github.com/RigsOfRods/rigs-of-rods/pull/744


Hiradur:
> Related with: #744

I think you mean this #652 

:+1:

ulteq:
From what I can tell now I would assume that:`game.clearEventCache()` is a placebo and does nothing.

PR #753, opened 2016-02-07
[Bugfix] Fixed up the truck spawning code

ulteq:
* Fixes the position of all kinds of pre-spawned trucks

* Fixes the position of trucks spawned in spawnboxes

* Fixes the position of trucks spawned on mesh terrains

Side-effect: Terrains without pre-spawned trucks won't automatically open a vehicle selector.

Plays nicely with: #752

### DIFF
--- a/source/main/gameplay/RoRFrameListener.cpp
+++ b/source/main/gameplay/RoRFrameListener.cpp
@@ -1276,7 +1276,6 @@ void RoRFrameListener::hideMap()
 #endif //USE_MYGUI
 }
 
-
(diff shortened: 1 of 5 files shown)


only-a-ptr:
:+1: 
The ```InitTrucks()``` method is a leftover from old mainmenu-less RoR flow. IIRC the idea is: after map is loaded and loader automatically appears, the game is in special state, where InitTrucks will be called. Loading/reloading any subsequent vehicle won't use it.

PR #757, opened 2016-02-08
[Cleanup] Cleaned up the collcabrate calculation

ulteq:
One vehicle standing, one vehicle driving at 80 km/h. The screenshots show the state after the collision.

**Without this patch**: 

**With this patch**:

### DIFF
--- a/source/main/physics/Beam.cpp
+++ b/source/main/physics/Beam.cpp
@@ -2572,10 +2572,8 @@ void Beam::interTruckCollisions(Real dt)
 	Vector3 plnormal;
 	Vector3 point;
 	Vector3 vecrelVel;
-	float trwidth;
 	int hitnodeid;
 	int hittruckid;
-	int tmpv;
 	node_t* hitnode;
 	node_t* na;
 	node_t* nb;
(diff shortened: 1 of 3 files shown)


mikadou:
What is `collcabrate`? collision-cab-rate? Are cabs triangles/faces of the submesh? How are rate and distance in the `collcabrate_t` structure related?

only-a-ptr:
@mikadou Collcab rate is a wannabe optimization to reduce amount of collision test per triangle. It might have been broken from the start.

I'll gather all info I got from Ulteq until now and start a forum thread. We need to write an overview wikipage/doxypage about this.

ulteq:
> Are cabs triangles/faces of the submesh?

Yes, `cabs` are triangles of the submesh and `contacters` are points.

> What is collcabrate?

The [collcabrate](https://github.com/ulteq/rigs-of-rods/blob/collcabrate-revert/source/main/physics/BeamData.h#L283-L287) specifies how many physics cycles are to be skipped before the next collision check of a given `cab` is performed. So far, the rate just depends on how long ago the last collision happened.

> How are rate and distance in the collcabrate_t structure related?

`rate` counts down, `distance` counts up.

`rate`: remaining amount of physics cycles to be skipped
`distance`: distance (in physics cycles) to the previous collision check

> Why was the if-block above moved into the for-loop?

Inverting the transformation matrix is one of the most expensive calculations that needs to be done for each `cab`.

`intraPointCD->hit_count` is usually really small (< 10). 

`hitnode->iswheel` is true for all `hit_counts` in 2-5% of all `intraPointCD->hit_count > 0` cases.

And doing 2% less matrix inversions saves more than the extra bool comparison in the for loop.

mikadou:
> The collcabrate specifies how many physics cycles are to be skipped before the next collision check of a given cab is performed. So far, the rate just depends on how long ago the last collision happened.

> rate: remaining amount of physics cycles to be skipped
> distance: distance (in physics cycles) to the previous collision check

Thank you, this clarifies the intention :smile_cat: Could you add these explanations as comments in the code?

mikadou:
> Inverting the transformation matrix is one of the most expensive calculations that needs to be done for each cab.
> 
> intraPointCD->hit_count is usually really small (< 10).
> 
> hitnode->iswheel is true for all hit_counts in 2-5% of all intraPointCD->hit_count > 0 cases.
> 
> And doing 2% less matrix inversions saves more than the extra bool comparison in the for loop.

Makes sense, but the code becomes harder to understand. It might be a good idea to use a wrapper class around Matrix3 which performs lazy initialization.
 
I think it is even possible to avoid the matrix inversion completely:

    bx = na->RelPosition - no->RelPosition;
    by = nb->RelPosition - no->RelPosition;
    bz = fast_normalise(bx.crossProduct(by));
    //coordinates change matrix
    Matrix3 forward( bx[0], bx[1], bx[2],
                               by[0], by[1]. by[2],
                               bz[0], bz[1], bz[2]);

ulteq:
> I think it is even possible to avoid the matrix inversion completely

The matrix inversion is where all the magic happens, we can't avoid it.

>
```
Matrix3 forward( bx[0], bx[1], bx[2],
                 by[0], by[1]. by[2],
                 bz[0], bz[1], bz[2]);
```

That would be the transpose of the matrix, but we need the inverse.

ulteq:
But changing:

`forward.FromAxes(bx,by,bz);`

to

`forward = Matrix3(bx[0], by[0], bz[0], bx[1], by[1]. bz[1], bx[2], by[2], bz[2]);`

might indeed speed things up a bit.

mikadou:
> The matrix inversion is where all the magic happens, we can't avoid it.

Hmm, here's my though process:

 * We have a triangle defined by three points *n0, na, nb*
 * We have two vectors *bx, by* spanning the plane defined by the collision triangle
 * *by* is the normal vector of this plane
 * We have the coordinates of point *p* relative to *n0* (in world coordinates)
 * Now we *project* the point *p* onto the three vectors *bx, by, bz*
  - i.e. `p_projected_onto_bx = bx[0] * p[0] + bx[1] * p[1] + bx[2] * p[2]`

What am I missing?

only-a-ptr:
@mikadou It's quite possible there's a solution which doesn't require the Matrix::inverse() call but computes things directly. I dunno about Ulteq, but personally I can't do math. I know how to use matrices for coordinate transforms, but I can only construct them usin OGRE's math API.

Your solution sounds somewhat familiar. Can you explain it more? I assume ```p_projected_onto_bx``` is a scalar, but what is it's meaning? Also, I assume bx[0] is ```bx.x```. Write a clearer example, please.

ulteq:
> What am I missing?

We still need to check if the point is inside of the triangle or not.

ulteq:
Things we could try to speed it up:
* Cache the transformation matrix and reuse it if the cab did not move
* Quit early if the distance between the point and the triangle is greater than the `collrange`

mikadou:
@only-a-ptr apparently I can't do math either. I'm afraid @ulteq is correct that there is no more efficient way.

PR #763, opened 2016-02-09
[Codechange] Allow anything but spaces in the command options

ulteq:
This would fix the Chinook: http://www.rigsofrods.com/repository/download/1426

Alternative would be fallback to ParseUnsafe

### DIFF
--- a/source/rig_file_input_output/RigDef_Regexes.h
+++ b/source/rig_file_input_output/RigDef_Regexes.h
@@ -959,7 +959,7 @@ DEFINE_REGEX( SECTION_COMMANDS,
     E_CAPTURE( E_POSITIVE_DECIMAL_NUMBER )  // #13 Expand key
     E_CAPTURE_OPTIONAL(                     // #14
         E_CAPTURE( E_DELIMITER )
-        E_CAPTURE( E_STRING_ALNUM_HYPHENS_USCORES_ONLY ) // #16 Options
+        E_CAPTURE( E_STRING_NO_SPACES )     // #16 Options
         E_CAPTURE_OPTIONAL(                 // #17
             E_CAPTURE( E_DELIMITER )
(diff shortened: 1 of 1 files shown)


only-a-ptr:
This would fail if there's a delimiter comma without any whitespace between options and description.
A working solution should be using ```E_STRING_ANYTHING_BUT_DELIMITER```

ulteq:
> A working solution should be using E_STRING_ANYTHING_BUT_DELIMITER

That doesn't work, the regex fails because of that trailing comma.

> This would fail if there's a delimiter comma without any whitespace between options and description.

I know. How common is this case?

> The Chinook above contains invalid trailing commas in "commands", so the above commit should report "invalid character <,> "

It reports something like: "ignoring invalid option ','"

ulteq:
So we just add `ParseCommandsUnsafe` and `ParseCommands2Unsafe` and be done with it?

only-a-ptr:
> That doesn't work, the COMMANDS regex fails because of that trailing comma.
Oh, right. It needs:
```
capture(contract)
capture(delim)
capture(expand)
capture(delim)
capture_opt(
  capture(delim)
  capture_opt(
    capture(options!)
    capture_opt(
      capture(delim)
      capture_opt(
        capture(desc)
      )
    )
  )
)
```

ulteq:
Yeah, that's something we should do everywhere. Will you do it? :grin:

only-a-ptr:
> Yeah, that's something we should do everywhere. Will you do it? :grin: 

We can only do that if we add ```ParseAbcUnsafe()``` everywhere. The current regexes are tuned to 
tolerate per-section-specific quirks, correcting all of them would break content. I have less annoying things to do :tired_face: I'll just fix the commands[2]

ulteq:
> I'll just fix the commands[2]

Fair enough. :+1:

PR #767, opened 2016-02-10
[Feature] collcabrate calculation based on collision velocity

ulteq:
What this does:

a) Estimate the maximum relative velocity at which `contacters` and `cabs` collide.

b) Use this estimate to calculate the maximum amount of physics steps (`collcabrate`) that can be skipped without "missing" any collision.

### DIFF
--- a/source/main/datatypes/rig_t.h
+++ b/source/main/datatypes/rig_t.h
@@ -100,7 +100,6 @@ struct rig_t
 	int free_sub;
 
 	int collcabs[MAX_CABS];
-	int collcabstype[MAX_CABS];
 	collcab_rate_t inter_collcabrate[MAX_CABS];
 	collcab_rate_t intra_collcabrate[MAX_CABS];
 	int free_collcab;
--- a/source/main/physics/Beam.cpp
+++ b/source/main/physics/Beam.cpp
@@ -2588,19 +2588,28 @@ void Beam::interTruckCollisions(Real dt)
 
 	for (int i=0; i<free_collcab; i++)
 	{
+		int tmpv = collcabs[i]*3;
+		no = &nodes[cabs[tmpv]];
+		na = &nodes[cabs[tmpv+1]];
+		nb = &nodes[cabs[tmpv+2]];
+
+		// upper bound of the relative collision velocity
+		float vdiff = interPointCD->max_contacter_speed + no->Velocity.length();
+		if (vdiff > 0.0f)
+		{
+			// upper bound of the amount of physics cycles which can be skipped
+			int skip_rate = floor(collrange / (vdiff * PHYSICS_DT));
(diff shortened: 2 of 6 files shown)


only-a-ptr:
Minor: Looks like ```rig_t::collcabstype``` is dead code: https://github.com/RigsOfRods/rigs-of-rods/search?utf8=%E2%9C%93&q=collcabstype Clean it up pls.

ulteq:
> Clean it up pls.

Done: https://github.com/ulteq/rigs-of-rods/commit/f61310126747de4732569d7160a956c6af02fd16

> Umm, magic 13 -> magic 80 is not a big win

Gone: https://github.com/ulteq/rigs-of-rods/commit/b1e465f4f50e8322bfa9e710c9ec40d61234a3bd

mikadou:
Could you explain a bit more on lines
https://github.com/ulteq/rigs-of-rods/commit/dfafe84ffa3b11b8d6d614dfefd2d2c0e127568f#diff-abbec6ed672585ca1ea0964858716deeR2611
and
https://github.com/ulteq/rigs-of-rods/blob/dfafe84ffa3b11b8d6d614dfefd2d2c0e127568f/source/main/physics/Beam.cpp#L2629 ?

I think I understood everything else.

ulteq:
> Could you explain a bit more on lines
ulteq@dfafe84#diff-abbec6ed672585ca1ea0964858716deeR2611

I have already changed:
`inter_collcabrate[i].rate = std::min(inter_collcabrate[i].distance, 80);`
into
`inter_collcabrate[i].rate = std::min(inter_collcabrate[i].distance, max_rate);`

> Can't we just do nothing in case of a collision and if no collision is detected then set the rate to the maximum rate possible for the current relative velocities

But I think you are right. That sounds like a much better idea.

only-a-ptr:
Great work. I only have minor comments to code clarity.

ulteq:
> if no collision is detected then set the rate to the maximum rate possible for the current relative velocities

I just tested it, it won't work. The `max_rate` is only valid for the past step, not for the following step.
I would for example fail to properly detect collisions on stacked containers.

ulteq:
> The parser should be updated to treat options 'p' and 'u' as aliases to 'c'. Documentation should be updated to reflect this. Constants OPTION_p_10xTOUGHER and OPTION_u_INVULNERABLE should be removed.

Yes, but please let's do that in another PR and after 0.4.6 is out.

I don't want to break something due to hasty cleanups.

only-a-ptr:
> Yes, but please let's do that in another PR and after 0.4.6 is out.
> I don't want to break something due to hasty cleanups.
Agreed :+1:

ulteq:
Let's delay this until after 0.4.6.

PR #769, opened 2016-02-10
[Codechange] Replaced the ApproxMath functions in most cases

ulteq:
Removing those makes the code faster on my CPU, and I think it will be faster on almost any CPU.

Thoughts?

Please do not merge, it's just an experiment for now.

### DIFF
--- a/source/main/physics/ApproxMath.h
+++ b/source/main/physics/ApproxMath.h
@@ -67,39 +67,6 @@ inline float frand_11()
     return( *((float*)&a) - 3.0f );
 }
 
-// Calculates approximate e^x.
-// Use it in code not requiring precision
-inline float approx_exp(const float x)
-{
-    if (x < -15)
-        return 0.f ;
-    else if (x > 88)
-        return 1e38f ;
-    else {
-        int i=12102203*x+1064652319;
-        return *(float *)&i;
-    }
-}
-
-// Calculates approximate 2^x
-// Use it in code not requiring precision
-inline float approx_pow2(const float x)
-{
-    int i = 8388608*x+1065353216;
-
-    return *(float *)&i;
-}
-
-// Calculates approximate x^y
-// Use it in code not requiring precision
-inline float approx_pow(const float x, const float y)
-{
-    float v = x;
-    int i = y * ( (*(int *)&v) - 1065353216) + 1065353216;
-
-    return *(float *)&i;
-}
-
 // Calculates approximate square_root(x)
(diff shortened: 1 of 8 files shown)


only-a-ptr:
I can't test anytime soon, but I agree with removing those functions. Their origin is unknown (they're definitely not author's own work) and the comment ```// Use it in code not requiring precision``` is not guiding enough.
:+1:

Hiradur:
:+1:

mikadou:
approx_pow      32.0342ns       (avg 100000 iterations)
    std::pow        53.0254ns       (avg 100000 iterations)
    approx_exp      19.4734ns       (avg 100000 iterations)
    std::exp        34.884ns        (avg 100000 iterations)
    fast_sqrt       41.9527ns       (avg 100000 iterations)
    std::sqrt       23.504ns        (avg 100000 iterations)

I'm not 100% sure if these measurements are correct ...

ulteq:
@mikadou You're right. My benchmark results using [google benchmark](https://github.com/google/benchmark) (on a Haswell CPU) are:

```
approx_pow      35.0782ns       (avg 100000000 iterations)
std::pow        36.8341ns       (avg 100000000 iterations)
approx_exp      25.2154ns       (avg 100000000 iterations)
std::exp        34.0092ns       (avg 100000000 iterations)
fast_sqrt       34.4802ns       (avg 100000000 iterations)
std::sqrt       27.5797ns       (avg 100000000 iterations)
```

Google Benchmark

Hiradur:
How noticeable is the FPS impact? I'd rather sacrifice 1-2 FPS to get rid of the hacky code.

ulteq:
> How noticeable is the FPS impact?

I don't notice any difference on my machine, but we should test this on older CPU models before merging it.

> I'd rather sacrifice 1-2 FPS to get rid of the hacky code.

Removing the `approx_exp` and `approx_pow` functions in the Collisions.cpp is the only thing that hurts the performance on my CPU. And we might not have to remove them. They do their job just fine.

ulteq:
I tried to reproduce my benchmark results today, and failed. I'm getting completely different results.

Hiradur:
The optimization level probably plays a large role in this, did you keep it constant?

ulteq:
> The compiler optimization level probably plays a large role in this, did you keep it constant?

Yes, I did not recompile between testing.

ulteq:
Let's delay this until after 0.4.7.

PR #775, opened 2016-02-11
[Bugfix] Fixes parsing of commands with trailing delimiter

ulteq:
Fixes the Chinook: http://www.rigsofrods.com/repository/download/1426

Closes: #763


only-a-ptr:
:+1:

PR #779, opened 2016-02-14
[Fix] Added CMake option for building with ogre 1.8

AnotherFoxGuy:
It maybe isn't the best solution, but it does the job


tritonas00:
I compile and use master with ogre 1.8 fine.

Whats the issue? :P

mikadou:
This fix is for Windows. I'd say it's good for now :+1:

PR #783, opened 2016-02-19
Add forward-declared library

tuckertwo:


Hiradur:
Is this required? I can compile without it (Linux, gcc 5.2.1).
Your commit message is also misleading, it's a header, not a library.

ulteq:
> Anyway, your commit message is misleading, it's a header, not a library. Could you correct that?

No response for nearly two months -> Closed

PR #789, opened 2016-02-22
mapbin file fixes

AnotherFoxGuy:
Fixed the game trying to place the mapbin file in the install folder and added the ogre version to the mapbin file
Should fix #685

PR #792, opened 2016-02-22
[Codechange] Reactived the terrain caching (mapbin) on linux

ulteq:
Noticeable speedup when loading maps like Auriga

This PR made it work: #789


Hiradur:
tested, works fine and speeds up loading :+1:

only-a-ptr:
:+1: Thanks

PR #794, opened 2016-02-22
[Fix] Fixed compiling with ogre 1.8

AnotherFoxGuy:
Fixed #789 If you want to compile with ogre 1.8


tritonas00:
:+1:

ulteq:
:+1: 

But I think we're using `ROR_USE_OGRE_1_9` to check if Ogre 1.9 is being used.

only-a-ptr:
@ulteq It's OK like this.

only-a-ptr:
Oops, misclicked :blush:

PR #797, opened 2016-02-23
[Codechange] Emulates the character spawn rotation of older RoR versions

ulteq:
Fixes the initial character orientation on maps like N-Labs or F1 Track

Temporary until we allow defining a spawn rotation in the terrn file

### DIFF
--- a/source/main/main_sim/MainThread.cpp
+++ b/source/main/main_sim/MainThread.cpp
@@ -1220,6 +1220,10 @@ void MainThread::LoadTerrain(Ogre::String const & a_terrain_file)
 	{
 		gEnv->player->setVisible(true);
 		gEnv->player->setPosition(gEnv->terrainManager->getSpawnPos());
+		if (!gEnv->terrainManager->hasPreloadedTrucks())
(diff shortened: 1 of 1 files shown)


Hiradur:
Not sure if I like this, is this really necesarry?

ulteq:
This "fix" allows you to use the same workflow as in previous versions to spawn vehicles. Otherwise you would have to carefully rotate the character into the right direction before spawning a vehicle to make sure it is aligned with the road (for example on the F1 track).

> is this really necesarry?

No.

> Not sure if I like this

I can live without it. It's only purpose is to temporarily make spawning vehicles more convenient.

Hiradur:
I think it's ok but there should be a comment what this is for because otherwise it looks like WTF code.

only-a-ptr:
@ulteq :+1: For backwards compatibility, no matter how dirty it is.
However, it's necessary to add a comment above the snippet:
```
// Classic behavior, retained for compatibility.
// Required for maps like N-Labs or F1 Track.
```

PR #758, opened 2016-02-08
Build system cleanup

mikadou:
Some cleanup and improvements related to the build system.
Also addresses #739

PR #803, opened 2016-03-07
changed to `dist: trusty` in .travis.yml

krichter722:
and removed `boost/latest` PPA because it's not available for trusty

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: cpp
+dist: trusty
 sudo: required
 
 compiler:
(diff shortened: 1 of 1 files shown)


Hiradur:
If we do this we should merge #758 before.
However, CI and CMake build system are currently in a freeze because the 0.4.6.0 release is imminent.

Hiradur:
We decided to not upgrade to trusty before it left beta status. Since your PR has a merge conflict and some other changes would also need to be done for trusty I'm closing this.

PR #807, opened 2016-03-17
Added commandline-argument "-joinserver"

AnotherFoxGuy:
Now you can join an MP-server with
```
RoR.exe -joinserver="server:port"
```
for example
```
RoR.exe -joinserver="ror.ezzg.be:14151"
```

### DIFF
--- a/source/main/main_sim/MainThread.cpp
+++ b/source/main/main_sim/MainThread.cpp
@@ -1,28 +1,28 @@
 /*
-	This source file is part of Rigs of Rods
-	Copyright 2005-2012 Pierre-Michel Ricordel
-	Copyright 2007-2012 Thomas Fischer
-	Copyright 2013-2014 Petr Ohlidal
+This source file is part of Rigs of Rods
+Copyright 2005-2012 Pierre-Michel Ricordel
+Copyright 2007-2012 Thomas Fischer
+Copyright 2013-2014 Petr Ohlidal
 
-	For more information, see http://www.rigsofrods.com/
+For more information, see http://www.rigsofrods.com/
 
-	Rigs of Rods is free software: you can redistribute it and/or modify
-	it under the terms of the GNU General Public License version 3, as
-	published by the Free Software Foundation.
+Rigs of Rods is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License version 3, as
+published by the Free Software Foundation.
 
-	Rigs of Rods is distributed in the hope that it will be useful,
-	but WITHOUT ANY WARRANTY; without even the implied warranty of
-	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-	GNU General Public License for more details.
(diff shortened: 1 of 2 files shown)


tritonas00:
:+1:

Hiradur:
:+1:

mikadou:
:+1: 

Because of the upcoming 0.4.6 release there was a somewhat mutual agreement to defer new features until after the release. Then, however, the forum shutdown happened. I'd say let's go back to normal development mode and merge pending changes while forum and website issues still need to be sorted out. I don't see a new release happening before these things have settled.  Your Thoughts?

disloyalpick:
:+1:  @mikadou

Hotrod55:
I feel like we should just mark 0.4.6 as released on GitHub, 0.4.5.1 is already old and buggy compared to the latest master, and the only way to currently download the 0.4.6 setup is from @AnotherFoxGuy's site, which isn't really "official".

Hiradur:
> I feel like we should just mark 0.4.6 as released on GitHub

Then it may go unnoticed however.

AnotherFoxGuy:
> Then it may go unnoticed, however.  

Now that http://rigsofrods.org/ has most of the active members that were on the old forum, can't we just release it and post a thread on there?

Hiradur:
Well, most people on ror.org already know about the release candidate. I think if we release it after ror.com is finished it would attract more visitors and show the project is indeed not dead.

ulteq:
I like those changes, but could you try to fix the whitespace reformatting?

The [actual changes](https://github.com/RigsOfRods/rigs-of-rods/pull/807/files?w=1) are hard to see in this commit.

ulteq:
Maybe just do this (https://github.com/RigsOfRods/rigs-of-rods/pull/807/commits/7d4f96f8561241db708cfb0814bce9f81d14c6c1) in a separate PR?

only-a-ptr:
@AnotherFoxGuy Very good idea, thanks.

However, this PR is a :-1: because of the horrible diffs and also [coding style!](http://ror.avrintech.net/rorwikibackup/index.php/Coding_style)

The only acceptable diff is https://github.com/RigsOfRods/rigs-of-rods/pull/807/commits/55add7cf8702fbaaeec93e255bd9cc657c1a1eec - short, clean, obvious. Although I have a problem with the `sargs` variable name (formerly `serverargs`) ... should be `server_args`. If that's too long for the expressions in there, then perhaps you should do `int colon = server_args.rfind(":")` instead of invoking `rfind()` 3 times in 2 lines of code. :smile:

The other 2 diffs are plain terrible. Apparently you're using some auto-formatter on everything you edit. Don't do that. Our sources do need re-formatting, we have mixed tabs/spaces and inconsistent `{}`s, but if you wan't to fix this, do it in separate "[cleanup] Whitespace/braces/whatever..." commits instead of mixing it with changes. Further, stick to the templates in our [coding style document](http://ror.avrintech.net/rorwikibackup/index.php/Coding_style). For instance, the license block and file-doc comment were formatted incorrectly.

PR #809, opened 2016-03-21
Updated crashrpt to 1.4.3

AnotherFoxGuy:
Updated the windows crash report tool to the latest version


ulteq:
Are we ever going to use that tool again. Opinions @only-a-ptr, @Hiradur?

Hiradur:
Isn't this Windows only? I can't comment on how useful it would be but I definitely won't be using it.

ulteq:
> what exactly is our target VS version now?

VS 2015

PR #810, opened 2016-03-22
Added angelscript function moveObject

AnotherFoxGuy:
Now you can move an object spawned with AngelScript using
```
game.moveObject("truckshop", position);
```

It doesn't update the collision box


Hiradur:
cool idea but not moving the collision box causes confusion, doesn't it?

ulteq:
What is the intended use case for this function? Map making?

You could call the function `moveObjectVisuals` to avoid confusions.

PR #811, opened 2016-03-22
Credits update

AnotherFoxGuy:
Updated the game credits window

### DIFF
--- a/source/main/gui/panels/GUI_GameAbout.cpp
+++ b/source/main/gui/panels/GUI_GameAbout.cpp
@@ -92,23 +92,22 @@ bool CLASS::IsVisible()
 void CLASS::initMisc()
 {
 	Ogre::UTFString AuthorsText = "";
-	Ogre::UTFString orange      = U("#FF7D02"); // colour key shortcut
-	Ogre::UTFString white       = U("#FFFFFF"); // colour key shortcut
-	Ogre::UTFString color1      = U("#66FF33"); // colour key shortcut
-	Ogre::UTFString newline     = U("\n");
+	Ogre::UTFString orange = U("#FF7D02"); // colour key shortcut
+	Ogre::UTFString white = U("#FFFFFF"); // colour key shortcut
+	Ogre::UTFString color1 = U("#66FF33"); // colour key shortcut
+	Ogre::UTFString newline = U("\n");
 
 	//Authors:
 	AuthorsText = orange + "Authors:" + newline;
 	AuthorsText = AuthorsText + color1 + "Pierre-Michel Ricordel (pricorde):" + white + " Physics Genius, Original Author, Core Developer, retired" + newline;
-	AuthorsText = AuthorsText + color1 + "Thomas Fischer (tdev):" + white + " Core Developer, inactive" + newline;	
-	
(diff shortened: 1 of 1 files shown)


source/main/gui/panels/GUI_GameAbout.cpp
```diff
@@ -131,18 +130,20 @@ void CLASS::initMisc()
...
 	AuthorsText = AuthorsText + color1 + "Klink:" + white + " Terrains conversion, few fixes and tweaks, dashboard designer" + newline;
 	AuthorsText = AuthorsText + color1 + "hagdervriese:" + white + " Linux fixes" + newline;
+	AuthorsText = AuthorsText + color1 + "AnotherFoxGuy:" + white + " Various fixes" + newline;
 
 	//Core Content Contributors
 	AuthorsText = AuthorsText + newline;
 	AuthorsText = AuthorsText + orange + "Core Content Contributors:" + newline;
 	AuthorsText = AuthorsText + color1 + "donoteat:" + white + " Improved spawner models, terrain work" + newline;
-	AuthorsText = AuthorsText + color1 + "kevinmce:" + white + " old Character" + newline;
-	AuthorsText = AuthorsText + color1 + "09Challenger:" + white + " new character animations" + newline;
```
Hiradur: What about 09 challenger? Is it safe to remove him entirely? Nothing left that he contributed?

AnotherFoxGuy: I believe that he only made a few vehicles and the animation for the character

ulteq: Both `kevinmce` and `09Challenger` are also mentioned here: https://github.com/RigsOfRods/rigs-of-rods/blob/028518296ae867f0edc3267a153725bd0b76cc67/AUTHORS.md

I think we should leave them in. What does @only-a-ptr say?

disloyalpick:
I agree with @ulteq No matter how little they contribute, they still help the future of RoR, so I say keep them. :+1:

vido89:
Also agree with @ulteq and @disloyalpick they should not be removed, put them at least in new section "early contrib" or "pre 0.4"

ulteq:
> I believe that he only made a few vehicles and the animation for the character

The animation we currently use?

Hiradur:
> The animation that we currently use?

They were made by hagdervriese, however, I'm not sure if he remade everything or just added missing animations.

only-a-ptr:
I don't like descriptions like "_old_ character" and "_new_ character animations" in the About window. What's _old_? Does this refer to 0.36's rorbot? The panel should either:
* reflect only current state of the game (which it already does, see "active/inactive" contributors) -> remove kevinmce (who made the current model, anyway?) and update 09challenger to "character animations"
* give up on time information altogether, i.e. kevinmce="character", though this is inaccurate. 
* add explicit timing, i.e. kevinmce="character, versions <0.4". But do we even have detailed enough info?

Speciesx:
https://github.com/AnotherFoxGuy/rigs-of-rods/blob/CreditsUpdate/source/main/gui/panels/GUI_GameAbout.cpp#L193
Since boost is gone, you can also remove this.

Hiradur:
and from DEPENDENCIES.md and BUILDING.md

PR #812, opened 2016-03-22
Add FreeBSD specific include

monwarez:
I am trying to get rigs of rods working again on FreeBSD.
Since I get some linking error with 270e084 (like multiple definition of function located at paged geometry). That's why I am looking to build the latest commit instead.
I finished the build but I am still getting multiple definition of function.
Example:
```
/usr/local/lib/libMyGUI.OgrePlatform.a(MyGUI_OgreRenderManager.cpp.o): In function `MyGUI::OgreRenderManager::getTexture(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&)':
/usr/ports/x11-toolkits/mygui-ogre/work/mygui-MyGUI3.2.2/Platforms/Ogre/OgrePlatform/src/MyGUI_OgreRenderManager.cpp:(.text+0x3060): multiple definition of `MyGUI::OgreRenderManager::getTexture(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&)'
/usr/local/lib/libMyGUI.OgrePlatform.a(MyGUI_OgreRenderManager.cpp.o):/usr/ports/x11-toolkits/mygui-ogre/work/mygui-MyGUI3.2.2/Platforms/Ogre/OgrePlatform/src/MyGUI_OgreRenderManager.cpp:(.text+0x3060): first defined here
```
I get same kind of error with paged geometry. And finally I get undefined reference for
`Ogre::SharedPtr<Ogre::Texture>::~SharedPtr()`
For the last one, I think it's an issue with mygui ogre plugin (because it is in libMyGUI.OgrePlatform.a).

### DIFF
--- a/source/main/physics/BeamFactory.cpp
+++ b/source/main/physics/BeamFactory.cpp
@@ -48,6 +48,7 @@ along with Rigs of Rods.  If not, see <http://www.gnu.org/licenses/>.
 #ifdef __FreeBSD__
 #include <sys/types.h>
 #include <sys/sysctl.h>
+#endif
 
 #ifdef USE_MYGUI
 #include "GUIMp.h"
--- a/source/main/physics/BeamFactory.cpp
+++ b/source/main/physics/BeamFactory.cpp
@@ -45,6 +45,10 @@ along with Rigs of Rods.  If not, see <http://www.gnu.org/licenses/>.
 #include <sys/sysinfo.h>
 #endif
 
+#ifdef __FreeBSD__
+#include <sys/types.h>
+#include <sys/sysctl.h>
+
 #ifdef USE_MYGUI
 #include "GUIMp.h"
 #include "GUIMenu.h"


Hiradur:
Nice to see someone trying to get it to work on FreeBSD.

> I get same kind of error with paged geometry. And finally I get undefined reference for
> Ogre::SharedPtr<Ogre::Texture>::~SharedPtr()

At which line of code is this? Are you building with C++11?

It's possible that some #ifdef (in the code itself or in the CMake files) don't trigger on FreeBSD, so I'd check that first.

monwarez:
Yes I am building with c++11(maybe I should try without). I have no idea about the line code, the error come at the very end when linking: `bin/RoR`.

Hiradur:
> the error come at the very end when linking: bin/RoR

You may be missing an #include somewhere then. First you need to find out which header from Ogre declares this and then you have to add an include of that header to the code for the object file which is missing the reference to this function.

> another problem is that I am not 100% sure that my port of paged-geometry and caelum are really working, how can I test them ?

There should be some samples included with these libraries. You can compile them and run them to see if it works.

ulteq:
Does this really fix build errors on FreeBSD? If so, can we merge it?

monwarez:
It fix compiling stage, not linking stage. I think you can merge it. (It's to replace _GNU_SOURCE include with FreeBSD specific include : 'sys/sysinfo.h' replaced by 'sys/types.h' and sys/sysctl.h' see man sysctl: http://www.unix.com/man-page/FreeBSD/3/sysctl/)

PR #813, opened 2016-03-25
Added angelscript function getNodePosition

AnotherFoxGuy:
Now you can get a position of a node using:
```
BeamClass @np_truck0 = game.getTruckByNum(0);		
vector3 vec = np_truck0.getNodePosition(121);
```

### DIFF
--- a/source/main/physics/Beam.cpp
+++ b/source/main/physics/Beam.cpp
@@ -6597,3 +6597,15 @@ bool Beam::inRange(float num, float min, float max)
 {
 	return (num <= max && num >= min);
 }
+
+Vector3 Beam::getNodePosition(int nodeNumber)
+{
+	if(nodeNumber >= 0 && nodeNumber < free_node)
+	{
+		return nodes[nodeNumber].AbsPosition;
+	}
+	else
+	{
+		return Ogre::Vector3();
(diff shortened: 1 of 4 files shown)


source/main/physics/Beam.cpp
```diff
@@ -6597,3 +6597,15 @@ bool Beam::inRange(float num, float min, float max)
...
+
+Vector3 Beam::getNodePosition(int nodeNumber)
+{
+	if(nodeNumber >= 0 && nodeNumber < free_node)
+	{
+		return nodes[nodeNumber].AbsPosition;
+	}
+	else
+	{
+		return Ogre::Vector3();
```
Hiradur: I guess this returns (0,0,0) but what if you query for a not existing node and you wrongly believe the node is at (0,0,0) instead of realizing it doesn't exist? Something like (MAX_INT, MAX_INT, MAX_INT) may be better.

ulteq: You could avoid that by calling `getNodeCount` first.

ulteq:
Looks good.

ulteq:
Can we merge this?

Hiradur:
:+1:

PR #830, opened 2016-04-15
Added CMake option for building with ogre 1.8

AnotherFoxGuy:
Added CMake option for building with ogre 1.8

### DIFF
--- a/cmake/DependenciesConfig.cmake
+++ b/cmake/DependenciesConfig.cmake
@@ -73,19 +73,45 @@ endmacro(importLib)
 	endif(NOT ROR_DEPENDENCIES_DIR)
 
 	#### REQUIRED COMPONENTS
-	set(Ogre_INCLUDE_DIRS
-		"${ROR_DEPENDENCIES_DIR}/includes/${ARCH_DIR}/Ogre/;"
-		"${ROR_DEPENDENCIES_DIR}/includes/${ARCH_DIR}/Ogre/Terrain;"
-		"${ROR_DEPENDENCIES_DIR}/includes/${ARCH_DIR}/Ogre/Paging;"
-		"${ROR_DEPENDENCIES_DIR}/includes/${ARCH_DIR}/Ogre/RTShaderSystem;"
-		"${ROR_DEPENDENCIES_DIR}/includes/${ARCH_DIR}/Ogre/Property;"
-		"The ogre include path to use")
-	importLib(OgreMain Ogre)
-	importLib(OgreTerrain Ogre)
-	importLib(OgrePaging Ogre)
-	importLib(OgreRTShaderSystem Ogre)
-	importLib(OgreProperty Ogre)
-	set(Ogre_LIBRARIES "OgreMain;OgreTerrain;OgrePaging;OgreRTShaderSystem;OgreProperty" CACHE STRING "The Ogre libs to link against")
+	
+	set(ROR_OGRE_VERSION_18     "FALSE" CACHE BOOL "Enable this if you are building with ogre 1.8")
+	
+	if(ROR_OGRE_VERSION_18)
+		
+		set(Ogre_INCLUDE_DIRS
+			"${ROR_DEPENDENCIES_DIR}/includes/${ARCH_DIR}/Ogre/;"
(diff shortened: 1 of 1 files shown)


mikadou:
Nice! Does this compile with VS2012? Or are there still some issues because of C++11 features being used in the code?

source/main/datatypes/node_t.h
```diff
@@ -8,7 +8,7 @@
 #pragma once
 
 #include <OgrePrerequisites.h>
-#include <OGRE/OgreVector3.h>
+#include <OgreVector3.h>
```
ulteq: Can't we just remove that include?

source/main/physics/Beam.cpp
```diff
@@ -6122,9 +6122,9 @@ bool Beam::LoadTruck(
 	}
 
 	// Calculate the approximate median
-	std::vector<Real> mx(free_node, 0.0f);
-	std::vector<Real> my(free_node, 0.0f);
-	std::vector<Real> mz(free_node, 0.0f);
+	std::vector<Ogre::Real> mx(free_node, 0.0f);
+	std::vector<Ogre::Real> my(free_node, 0.0f);
+	std::vector<Ogre::Real> mz(free_node, 0.0f);
```
ulteq: Where does the namespace conflict come from?

AnotherFoxGuy: Real can be found here: https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/physics/collision/CartesianToTriangleTransform.h#L32
And here: https://bitbucket.org/sinbad/ogre/src/7343503880d45d110ef2ab98c3a5e82b76b29eb1/OgreMain/include/OgrePrerequisites.h?at=v1-9&fileviewer=file-view-default#OgrePrerequisites.h-70

source/main/datatypes/node_t.h
AnotherFoxGuy: No, because RelPosition, AbsPosition, Velocity, Forces and smoothpos are Ogre::Vector3

source/main/physics/Beam.cpp
mikadou: Looks like I'm responsible for this :wink: I wonder why gcc didn't complain?

source/main/datatypes/node_t.h
ulteq: But it did work without it, 20 days ago, didn't it: https://github.com/RigsOfRods/rigs-of-rods/commit/341f1fac1433c0c99c16816eea4e80a071f12ca9

source/main/physics/Beam.cpp
ulteq: I think we should get rid of these: https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/physics/collision/CartesianToTriangleTransform.h#L30-L32

ulteq:
I would suggest that we get rid of the `using-directives` in the header file:
* https://github.com/RigsOfRods/rigs-of-rods/pull/831/commits/974add055fc2bee419a3fff609be6cb9d4cd6c83

instead of:
* https://github.com/RigsOfRods/rigs-of-rods/pull/830/commits/97f5d330fc5e597c06481c6da8c3feda04e2c5b1

ulteq:
This [commit](https://github.com/RigsOfRods/rigs-of-rods/pull/830/commits/97f5d330fc5e597c06481c6da8c3feda04e2c5b1) is no longer needed, right?

mikadou:
Just to mention it. There is one more using declaration in a header here https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/physics/collision/Triangle.h#L26

ulteq:
I've created a separate PR that focuses on the header cleanup: https://github.com/RigsOfRods/rigs-of-rods/pull/842

@AnotherFoxGuy Can you close this PR and create a new one that only contains this commit: https://github.com/RigsOfRods/rigs-of-rods/pull/830/commits/83ff912304f705f493f10c755643c3e69597162b?

AnotherFoxGuy:
I've deleted the other 2 commits

Issue #826, opened 2016-04-09
Replay bar doesn't hide with GUI


Hotrod55:
Currently the "u" key hides the notice window and speedo, but it doesn't hide the replay bar. 
Example:
![screenshot_2016-04-08_20-03-47_1](https://cloud.githubusercontent.com/assets/11002490/14400361/1b6977e0-fdc5-11e5-9a2f-31b8e3b4f2af.png)

PR #834, opened 2016-04-20
[Bugfix] Replay bar does not hide with with GUI

ulteq:
The replay panel should be managed by the GUIManager

Fixes: #826


Hotrod55:
:+1:

PR #836, opened 2016-04-21
replaced boost::shared_ptr with std::shared_ptr

Speciesx:
Since everyone use a newer compiler, i think we could replace `boost::shared_ptr` with `std::shared_ptr`.

### DIFF
--- a/source/main/gui/panels/GUI_RigEditorLandVehiclePropertiesWindow.cpp
+++ b/source/main/gui/panels/GUI_RigEditorLandVehiclePropertiesWindow.cpp
@@ -133,8 +133,8 @@ void CLASS::ResetEngineEngoptionForm()
 }
 
 void CLASS::Import(
-	boost::shared_ptr<RigDef::Engine> engine,
-	boost::shared_ptr<RigDef::Engoption> engoption
+	std::shared_ptr<RigDef::Engine> engine,
+	std::shared_ptr<RigDef::Engoption> engoption
 	)
 {
     if (!engine || !engoption)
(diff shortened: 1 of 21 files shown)


tritonas00:
This means that we no longer need boost as dependency?

Speciesx:
no, there is still boost::regex, boost::lexical and boost::optional.

tritonas00:
And those can't be replaced?

Speciesx:
I tried to replace boost::regex, boost::smatch and boost::ssub_match, but ror doesn't start after this change.

only-a-ptr:
> there is still boost::regex, boost::lexical and boost::optional. 

Whaaat? :smile: Who put that there? Can we throw it out? Please investigate.

One note: Quite a few edited files belong to RigEditor, which is obsolete. I won't continue it, I want the editor be written in Python (or some other scripting lang) rather than C++. Could anyone please remove all RigEditor code? Things to look into:
* Main menu (GUI button + startup logic)
* GUI panels (source/main/gui/panels/GUI_RigEditor*)
* Logic (source/main/rig_editor)

tritonas00:
I remember i compiled 0.38 without boost and caleum by removing a few lines. RoR worked fine.

But caleum needs boost i think

mikadou:
> Whaaat? :smile: Who put that in there? Can we throw it out? Please investigate.

I have introduced `boost::optional` in the collision code. Replacing `boost::shared_ptr` with `std::shared_ptr` makes sense as it is now available in the official standard. For other stuff not (yet) in the standard using Boost is fine in my opinion.

Speciesx:
> Replacing boost::shared_ptr with std::shared_ptr makes sense as it is now available in the official >standard. For other stuff not (yet) in the standard using Boost is fine in my opinion.

regex is also available as standard.

Hiradur:
@Speciesx Can you upload your diff to the master branch for your changes not in this pull request so we can take a look?

monwarez:
By the way, do c++1y (a.k.a c++14) is too young or not ? So that we can replace for example `auto def = std::shared_ptr<RigDef::File>(new RigDef::File()); ` by `auto def = std::make_shared<RigDef::File>();`

Hiradur:
@monwarez Switching to C++14 is too early IMO. I think a good measure is if the default compiler in Debian stable (gcc 4.9) has *full* support for the standard. That is not the case (gcc 5 has full support for C++14 [1]). 
There are also the older Ubuntu LTS versions 12.04 (gcc 4.6, but this one dropped out already since it doesn't support C++11 fully) and 14.04 (gcc 4.8). At least the LTS release before the current LTS release from Ubuntu should be supported.

[1] https://gcc.gnu.org/projects/cxx-status.html#cxx14

only-a-ptr:
Hmm, this makes me want to write C++ usage guidelines for RoR, such as these: http://llvm.org/docs/CodingStandards.html. I'll create an issue ticket on it.

@mikadou Please don't introduce any more boost.

PR #846, opened 2016-04-25
replaced boost::regex with std::regex

Speciesx:
I've replaced `boost::regex`, `boost::smatch` and `boost::ssub_match` with `std` and yes, ror still works.

### DIFF
--- a/source/rig_file_input_output/RigDef_Parser.cpp
+++ b/source/rig_file_input_output/RigDef_Parser.cpp
@@ -91,10 +91,10 @@ void Parser::ParseLine(Ogre::String const & line_unchecked)
 	bool scan_for_keyword = true;
 
 	/* Check line type */
-	boost::smatch line_type_result;
+	std::smatch line_type_result;
 	if (m_in_block_comment)
 	{
-		if (boost::regex_match(line, Regexes::CHECK_BLOCK_COMMENT_END))
+		if (std::regex_match(line, Regexes::CHECK_BLOCK_COMMENT_END))
 		{
 			m_in_block_comment = false;
 		}
@@ -102,7 +102,7 @@ void Parser::ParseLine(Ogre::String const & line_unchecked)
 	}
 	else if (m_in_description_section)
 	{
-		if (boost::regex_match(line, Regexes::CHECK_END_DESCRIPTION))
+		if (std::regex_match(line, Regexes::CHECK_END_DESCRIPTION))
 		{
 			m_in_description_section = false;
 		}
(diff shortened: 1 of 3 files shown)


tritonas00:
Nice! Anything else left that needs boot?

Speciesx:
> Nice! Anything else left that needs boost?

apart from Ogre, Caelum and PagedGeometry.

```
source\main\physics\collision\CartesianToTriangleTransform.h(137):    mutable boost::optional<Ogre::Matrix3> m_matrix;  ///< Cached transformation matrix.
source\main\physics\collision\Triangle.h(58):  mutable boost::optional<Ogre::Vector3> m_normal;  ///< Cached normal vector
source\main\gui\panels\GUI_GameSettings.cpp(1076):				if (boost::starts_with(Application::GetInputEngine()->eventIDToName(mapIt->first).c_str(), m_keymap_group->getCaption()))
source\main\gui\panels\GUI_GameSettings.cpp(1083):			catch (boost::bad_lexical_cast) {
source\main\gui\panels\GUI_GameSettings.cpp(1125):				if (boost::starts_with(Application::GetInputEngine()->eventIDToName(mapIt->first).c_str(), _sender->getItemNameAt(_index)))
source\main\gui\panels\GUI_GameSettings.cpp(1132):			catch (boost::bad_lexical_cast) {
source\main\gui\Console.cpp(291):					gValue = boost::lexical_cast<float>(args[1].c_str());
source\main\gui\Console.cpp(307):				water->setHeight(boost::lexical_cast<float>(args[1].c_str()));
```

Hiradur:
Boost for Ogre is optional and a GitHub search for 'boost' only turns up the CMakeFiles in Paged Geometry and Caelum but no code so maybe it can just be removed.

tritonas00:
Ogre, Paged Geometry and MyGUI builds fine without boost. Caelum not :P

Hiradur:
> Caelum not :P

Does the compiler or CMake complain?

Speciesx:
> Boost for Ogre is optional and a GitHub search for 'boost' only turns up the CMakeFiles in Paged Geometry and Caelum but no code so maybe it can just be removed.

I've also noticed that boost nowhere used in Caelum and PagedGeometry.

tritonas00:
@Hiradur 

After removing boost from camke, it fails here https://github.com/RigsOfRods/ogre-caelum/blob/master/main/src/InternalUtilities.cpp#L78

only-a-ptr:
@tritonas00 The function that uses it: https://github.com/RigsOfRods/ogre-caelum/blob/master/main/src/InternalUtilities.cpp#L70-L80 is apparently supposed to convert a ```void+``` pointer to string, but does so in a very convoluted manner. It should be done like this: http://stackoverflow.com/a/1255185

Please research where that function is used (if anywhere) and fix it.

tritonas00:
@hiradur already helped me with that with #include <cstddef>

only-a-ptr:
Anyway, :+1:

PR #847, opened 2016-04-26
[Codechange] Removed the use of the boost::optional class template

ulteq:


only-a-ptr:
I just checked boost implementation: http://www.boost.org/doc/libs/1_34_0/boost/optional/optional.hpp

It does exactly the same thing, this code is at bottom of ```optional_base<>```:
```
    bool m_initialized ;
    storage_type m_storage ;
```
Except the header is MEGA convoluted.

PR #853, opened 2016-04-26
removed boost from cmake files

tritonas00:
needs #846 #847 #849 and probably https://github.com/RigsOfRods/ogre-caelum/pull/1 first

PR #857, opened 2016-04-29
[Codechange] Added linker flag '-lboost_system' on unix

ulteq:
Fixes linker error on Linux, when Ogre is build with Boost enabled:
```
/usr/bin/ld: CMakeFiles/RoRConfig.dir/Configurator.cpp.o: undefined reference to symbol '_ZN5boost6system15system_categoryEv'
/usr/lib/libboost_system.so.1.60.0: error adding symbols: DSO missing from command line
```

### DIFF
--- a/source/configurator/CMakeLists.txt
+++ b/source/configurator/CMakeLists.txt
@@ -65,7 +65,7 @@ ELSEIF(UNIX)
    PKG_CHECK_MODULES  (Ogre OGRE REQUIRED)
    PKG_CHECK_MODULES  (Ois OIS REQUIRED)
 
-   set(OS_LIBS "")
+   set(OS_LIBS "-lboost_system")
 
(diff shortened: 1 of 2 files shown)


Hiradur:
https://github.com/RigsOfRods/rigs-of-rods/pull/853/commits/53d1c95aa41728d4ecff3b4b70e9997a1efb8a6c#diff-204a9fa87a163aed55a2912936914d00L194
is a more elegant way. 
I'm not really  happy with this. Maybe it would be possible to query OGRE if it was build with boost and if yes only then link boost_system.

ulteq:
> I'm not really happy with this.

Me neither.

> Maybe it would be possible to query OGRE with CMake if it was build with boost and if yes only then link boost_system.

Sounds good.

PR #858, opened 2016-04-29
[Cleanup] Streamlined the physics simulation update routine

ulteq:
Read this first: https://github.com/RigsOfRods/rigs-of-rods/pull/858#issuecomment-216222555

* Removed the BeamFactory worker thread
* Merged threadentry() with UpdatePhysicsSimulation()
* Moved UpdatePhysicsSimulation() from Beam to BeamFactory
* Fixed crash when `DisableThreadPool=Yes`

**Edit:**
* Moved all physics related updates into `BeamFactory::calcPhysics`
* Chopped the  `Beam::frameStep` into several smaller functions
* Added new `syncWithSimThread` method, which is now called at the beginning of the `RoRFrameListener::frameStarted()` method
* SimThread and MainThread no longer access the truck engine at the same time
* `updateFlexbodiesPrepare()` and `UpdateEvents()` no longer run in parallel with the SimThread

### DIFF
--- a/source/main/gameplay/RoRFrameListener.cpp
+++ b/source/main/gameplay/RoRFrameListener.cpp
@@ -1502,7 +1502,7 @@ bool RoRFrameListener::frameStarted(const FrameEvent& evt)
 		if (!m_is_sim_paused)
 		{
 			BeamFactory::getSingleton().updateFlexbodiesFinal();   // Waits until all flexbody tasks are finished 
-			BeamFactory::getSingleton().calcPhysics(dt);           // we simulate one truck, it will take care of the others (except networked ones)
+			BeamFactory::getSingleton().calcPhysics(dt);
 		}
 	}
 
--- a/source/main/gameplay/RoRFrameListener.cpp
+++ b/source/main/gameplay/RoRFrameListener.cpp
@@ -1501,8 +1501,9 @@ bool RoRFrameListener::frameStarted(const FrameEvent& evt)
 
 		if (!m_is_sim_paused)
 		{
-			BeamFactory::getSingleton().updateFlexbodiesFinal();   // Waits until all flexbody tasks are finished 
+			BeamFactory::getSingleton().joinFlexbodyTasks();       // Waits until all flexbody tasks are finished
 			BeamFactory::getSingleton().calcPhysics(dt);
+			BeamFactory::getSingleton().updateFlexbodiesFinal();   // Updates the harware buffers 
 		}
 	}
 
--- a/source/main/gameplay/RoRFrameListener.cpp
+++ b/source/main/gameplay/RoRFrameListener.cpp
@@ -1293,7 +1293,7 @@ bool RoRFrameListener::frameStarted(const FrameEvent& evt)
 	dt = std::min(dt, 0.05f);
 	m_time += dt;
 
-	BeamFactory::getSingleton().syncWithSimThread();
+	BeamFactory::getSingleton().SyncWithSimThread();
 
 	RoR::Application::GetInputEngine()->Capture();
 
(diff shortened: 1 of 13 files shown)


source/main/physics/BeamFactory.h
```diff
@@ -181,13 +167,16 @@ class BeamFactory : public StreamableFactory < BeamFactory, Beam >, public Zeroe
...
+	std::shared_future<void> thread_future;
+
 	bool thread_mode;
 	int num_cpu_cores;
 
 	Beam *trucks[MAX_TRUCKS];
 	int free_truck;
 	int previous_truck;
 	int current_truck;
+	int simulatedTruck;
```
only-a-ptr: Ahem... What about `m_simulated_truck` instead of `simulatedTruck`?

I would understand omitting the `m_` for consistency, but camelCase? This still applies you know: http://ror.avrintech.net/rorwikibackup/index.php/Coding_style#The_template :smile:

only-a-ptr:
The PR comments sound pretty good.

I can't really evaluate the code, I lost track of the simulation logic. The 1st commit which replaces custom logic with std::*future looks great, other seem to mostly move code around. But as I said, I haven't worked with it in a while.

One thing - after you're done refactoring, please make a formatting sweep of Beam.cpp and BeamFactory.cpp (eventually elsewhere) - Transform all variables to `underscored_lowercase`, members to `m_prefixed_lcase` and functions to `CamelCase()`, add explict `this->` to member calls. I'd like to avoid new/refactored code making formatting mess just because the code was messy before.

ulteq:
> What about m_simulated_truck instead of simulatedTruck

I will do it like that, before we merge it. I've actually already renamed all of them, but it cluttered the diff and I didn't want to make the commits even harder to understand due to code reformatting.

ulteq:
> other seem to mostly move code around.

Two main things changed:

* The [`Beam::frameStep`](https://github.com/RigsOfRods/rigs-of-rods/blob/b529d010815a51590f1a0c3df48ce6bb082d6d29/source/main/physics/Beam.cpp#L1499-L1503) method is gone: 
That means we no longer have to call it on the `ACTIVATED` truck, to start the simulation of all other trucks

* Our main thread code no longer runs in parallel with the sim thread
Up until now [this](https://github.com/RigsOfRods/rigs-of-rods/blob/b529d010815a51590f1a0c3df48ce6bb082d6d29/source/main/physics/Beam.cpp#L1587-L1642) was the only code section which was executed while the sim thread was paused.

The sim thread now only runs while Ogre is doing its thing in the background, but not while we execute code in the "MainThread".

We [wait](https://github.com/ulteq/rigs-of-rods/blob/simthread-simplification/source/main/gameplay/RoRFrameListener.cpp#L1293) for the sim thread to finish before we enter the main thread code section.
And we [launch](https://github.com/ulteq/rigs-of-rods/blob/simthread-simplification/source/main/gameplay/RoRFrameListener.cpp#L1508) the sim thread after the main thread is done.

We should still move most of our code from the [`frameStarted`](http://www.ogre3d.org/docs/api/1.9/class_ogre_1_1_frame_listener.html#a88238b65f9ce32f0de8c5eeeb7b6c6ad) event into the [`frameRenderingQueued`](http://www.ogre3d.org/docs/api/1.9/class_ogre_1_1_frame_listener.html#a2785cb38589ab9df123fa401cbda31c3) event.

source/main/gameplay/RoRFrameListener.cpp
```diff
@@ -1512,6 +1504,12 @@ bool RoRFrameListener::frameStarted(const FrameEvent& evt)
 				}
 			}
 		}
+
+		if (!m_is_sim_paused)
+		{
+			BeamFactory::getSingleton().updateFlexbodiesFinal();   // Waits until all flexbody tasks are finished 
+			BeamFactory::getSingleton().calcPhysics(dt);           // we simulate one truck, it will take care of the others (except networked ones)
```
only-a-ptr: `// we simulate one truck, it will take care of the others (except networked ones)`
AFAIK this is a deprecated comment. It applied in the `Beam::frameStep()` era, but not anymore.

only-a-ptr:
It all makes sense now.
:+1: 
Thanks

Issue #820, opened 2016-04-06
Props Misplaced/Chassis Deformed


graysonk95:
I haven't come across any other vehicles with this same issue in 0.4.6.0, but on my HDX all of the props are misplaced and/or the flexbodies are deformed in a strange way at spawn.

0.4.0.7 had a weird deformation at spawn which was fixed for 0.4.5.1 and up, but for some reason my bus (which was perfectly fine in 0.4.5.1) now is experiencing a similar deformation in 0.4.6.0 (all nightlies and release candidates).

All of the switches and buttons are quite far off from where they should be
![screenshot_2016-04-06_13-10-05_1](https://cloud.githubusercontent.com/assets/10550037/14326030/20a0da04-fbfa-11e5-86ea-5457e4ae46d8.png)

The doors are flexbodies and they are tweaked at spawn
![screenshot_2016-04-06_13-14-38_1](https://cloud.githubusercontent.com/assets/10550037/14326032/22c80df2-fbfa-11e5-97ea-48639644e89c.png)

I haven't had any time to troubleshoot to see if there's anything on my end I can change (in case this is a parsing issue with beam rigidity or something), but regardless, something changed between 0.4.5.1 and 0.4.6.0.

ulteq:
Are you able to reproduce the props misplacement every time you spawn that truck?

> the flexbodies are deformed in a strange way at spawn

Can you make two comparison screenshots (0.4.5.1 vs. 0.4.6.0)?

> all of the props are misplaced

That sounds like a parsing issue

graysonk95:
> Are you able to reproduce the props misplacement every time you spawn that truck?
Yup, it's identical every time on every map.
My best guess is that it's an issue parsing the set_beam_defaults lines. It seems like the chassis itself is deformed at spawn more so than prop coordinates being improperly parsed/rounded.

> Can you make two comparison screenshots (0.4.5.1 vs. 0.4.6.0)?
0.4.5.1
![screenshot_654](https://cloud.githubusercontent.com/assets/10550037/14353606/c28999d4-fca7-11e5-98f8-82d86f0c368d.png)

0.4.6.0 RC3
![screenshot_2016-04-07_10-00-45_1](https://cloud.githubusercontent.com/assets/10550037/14353618/c8eb545c-fca7-11e5-9cac-afe6ad11b530.png)

tritonas00:
I can reproduce it easily if another vehicles is spawned first.

[...]

PR #862, opened 2016-05-01
[Bugfix] Fixes spawning misplaced props/flexbodies

ulteq:
Fixes: #820


ulteq:
-> This one is better: https://github.com/RigsOfRods/rigs-of-rods/pull/866

PR #865, opened 2016-05-01
[Codechange] Revert cmake boost removal

ulteq:
Fixes linker error on Linux, when Ogre is build with Boost enabled:
```
/usr/bin/ld: CMakeFiles/RoRConfig.dir/Configurator.cpp.o: undefined reference to symbol '_ZN5boost6system15system_categoryEv'
/usr/lib/libboost_system.so.1.60.0: error adding symbols: DSO missing from command line
```
Alternative for: https://github.com/RigsOfRods/rigs-of-rods/pull/857


tritonas00:
But this fails with deps without boost

ulteq:
> Maybe this #857 (comment) is the best option, to check ogre first.

-> Ok, let's continue there: https://github.com/RigsOfRods/rigs-of-rods/pull/857

PR #895, opened 2016-05-12
[Codechange] Implemented a thread-safe truck deletion queue

ulteq:
Trucks can no longer be deleted while the sim thread is running.

**Singleplayer:**
* Reloading a truck now properly deletes the old truck

**Multiplayer:**
* Fixes crash (on linux) when a remote player disconnects after he has spawned a truck

### DIFF
--- a/source/main/gameplay/RoRFrameListener.cpp
+++ b/source/main/gameplay/RoRFrameListener.cpp
@@ -1483,6 +1483,8 @@ bool RoRFrameListener::frameStarted(const FrameEvent& evt)
 			RoR::Application::GetGuiManager()->UpdateSimUtils(dt, curr_truck);
 		}
 
+		BeamFactory::getSingleton().handleTruckDeletion();
+
 		if (!m_is_sim_paused)
 		{
 			BeamFactory::getSingleton().joinFlexbodyTasks();       // Waits until all flexbody tasks are finished
@@ -1678,12 +1680,6 @@ void RoRFrameListener::reloadCurrentTruck()
 		return;
 	}
 
-	// exit the old truck
-	BeamFactory::getSingleton().setCurrentTruck(-1);
-
-	// remove the old truck
-	curr_truck->state = RECYCLE;
-
 	// copy over the most basic info
 	if (curr_truck->free_node == newBeam->free_node)
 	{
@@ -1712,10 +1708,7 @@ void RoRFrameListener::reloadCurrentTruck()
 	RoR::Application::GetGuiManager()->PushNotification("Notice:", msg);
 #endif //USE_MYGUI
 
-	// dislocate the old truck, so its out of sight
-	curr_truck->resetPosition(100000, 100000, false, 100000);
-	// note: in some point in the future we would delete the truck here,
-	// but since this function is buggy we don't do it yet.
+	BeamFactory::getSingleton().removeCurrentTruck();
 
 	// reset the new truck (starts engine, resets gui, ...)
 	newBeam->reset();
--- a/source/main/physics/Beam.cpp
+++ b/source/main/physics/Beam.cpp
@@ -119,6 +119,7 @@ Beam::~Beam()
 	{
 		SoundScriptManager::getSingleton().trigStop(this->trucknum, i);
 	}
+	StopAllSounds();
 #endif // USE_OPENAL
 
 	// destruct and remove every tiny bit of stuff we created :-|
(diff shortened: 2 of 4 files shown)


only-a-ptr:
> Trucks can no longer be deleted while the sim thread is running.

You probably wanted to say "Deleting trucks now possible while sim-thread is running". :smile: 

Thought I'm happy this fixes a bug, I don't like the design much. It adds another explicit ["sync me now!" call](https://github.com/RigsOfRods/rigs-of-rods/pull/895/files#diff-e11c82258dd7c1910ddc4dbc51cafa60R807) + new mutex guard instead of using the existing infrastructure, i.e. processing queued deletes as a rudimentary game task in [frameStarted()](https://github.com/RigsOfRods/rigs-of-rods/blob/9e4a1404ef4808fdd3f51f21987ba30d78b7fe16/source/main/gameplay/RoRFrameListener.cpp#L1275)

I'm probably too lost in my 'proper-mt' vision, but here's how I imagine truck delete mechanism:
* main thread queues a delete request.
* when sim-thread is synced, the truck is excluded from all simulation (should be plain removal of a pointer from array). Actual de-allocation can be postponed if necessary.
* rendering thread removes all truck's visuals. IMPORTANT: all material instances must be unique to truck and removed here,, too!
* sound handling: to be resolved

ulteq:
> It adds another explicit "sync me now!" call

You are right, we don't need that call, because `handleTruckDeletion` is called inside `frameStarted()` after calling `SyncWithSimThread`.

> It adds a new mutex guard instead of using the existing infrastructure

The issue is that delete requests can come from both the main thread and the network thread.

only-a-ptr:
> The issue is that delete requests can come from both the main thread and the network thread.

Then make an extra queue for the network. When synced, transfer the requests to main delete queue.

ulteq:
I just realized that the truck deletes are already queued in the network thread.
I think we can safely remove all syncing / locking in the current truck deletion queue implementation.

1. https://github.com/ulteq/rigs-of-rods/blob/3c2a189b8c30209c97a1ce909f5a2699f009159d/source/main/network/StreamableFactory.h#L144

2. https://github.com/RigsOfRods/rigs-of-rods/blob/c8066e218caff159738fe1f39151c34b69cb42dc/source/main/network/NetworkStreamManager.cpp#L226

3. https://github.com/ulteq/rigs-of-rods/blob/3c2a189b8c30209c97a1ce909f5a2699f009159d/source/main/gameplay/RoRFrameListener.cpp#L1281

ulteq:
> Further, how is spawning over network handled this time around?

Inside of `syncRemoteStreams()`: https://github.com/ulteq/rigs-of-rods/blob/3c2a189b8c30209c97a1ce909f5a2699f009159d/source/main/network/StreamableFactory.h#L101-L150

only-a-ptr:
Hmm... the whole `Stream*whatever` subsystem looks like a lot of OOP cruft for a simple network handling to me. It seems to be designed to allow multiple managers (chat, character, beamfactory) to create their own streamfactories and register streams. Everyone calls `lockStreams()` all he wants. That's bullshit design. For network, we need to manage bandwidth - trucks are important, characters are less, chat is lowest priority. For local data, we can queue. Further, it's full of the "loop all streams to find one by StreamID" crap, just like our sound manager, our input manager, and damn, IIRC even BeamFactory.

How about being punks and rewriting the whole StreamMess into one compilation unit (Networking.{h/cpp}), which works like this:
```
namespace Networking
{

struct Stream { ... };

// Stuff from class Network, known as gEnv->network.
// It's pseudo-singleton anyway, so why not global functions?
SendMessageRaw()
SendMessage()
ReceiveMessage()

Connect()
Disconnect()

SenderThreadStart()
ReceiverThreadStart()
// etc... from Network.h

// Stuff from NetworkStreamManager -- was a singleton 
// For the time being, we can preserve the abstract stream principle
AddLocalStream()
AddRemoteStream()
RemoveStream()
SendStreams()
ReceiveStreams()

// StreamableFactory => eliminate! Use `#include "Networking.h"` and the function calls directly, please!
}
```

ulteq:
Split into separate pull requests:
* https://github.com/RigsOfRods/rigs-of-rods/pull/906
* https://github.com/RigsOfRods/rigs-of-rods/pull/908
* https://github.com/RigsOfRods/rigs-of-rods/pull/907
* https://github.com/RigsOfRods/rigs-of-rods/pull/909

PR #904, opened 2016-05-13
[Cleanup] Update prop animation outside of physics core

mikadou:
Implements TODO comment https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/physics/BeamForcesEuler.cpp#L81


source/main/physics/BeamFactory.cpp
```diff
@@ -886,6 +886,7 @@ void BeamFactory::calcPhysics(float dt)
 		m_trucks[t]->handleResetRequests(dt);
 		m_trucks[t]->handleTruckPosition(dt);
 		m_trucks[t]->updateAngelScriptEvents(dt);
+		m_trucks[t]->UpdatePropAnimations(dt);
```
ulteq: This will call `calcAnimators()` regardless of the `state` of the truck. Do we want that?

[So far](https://github.com/mikadou/rigs-of-rods/blob/0f165f036be41b8fbcceb10ee046653ffac5e829/source/main/physics/BeamForcesEuler.cpp#L1160) it was only called for trucks with states below `SLEEPING`.

ulteq:
This somehow changes the rotational speed of props.

Steps to reproduce:

* Spawn a [Thomas Saf-T-Liner HDX](http://rigsofrods.org/thread-28.html)
* Enter first person view (cinecam)
* Press F6

Then watch the ventilator.

source/main/physics/Beam.cpp
```diff
@@ -6225,273 +6225,270 @@ void Beam::UpdatePropAnimations(const float dt)
 
 		while (props[propi].animFlags[animnum])
 		{
-			if (props[propi].animFlags[animnum])
```
mikadou: Only removed this if statement, the other line changes are due to indentation.

tritonas00:
With this PR, the fans speed in hdx varies as the fps change.

At 60 fps is good fast, but at 30 or lower is noticeable slow.

tritonas00:
Ignore the previous message, it happens without this PR also.

PR #906, opened 2016-05-13
[Codechange] Propagate the state of recycled trucks over the network,…

ulteq:
… to prevent creating dummy trucks

### DIFF
--- a/source/main/physics/BeamFactory.cpp
+++ b/source/main/physics/BeamFactory.cpp
@@ -893,16 +893,16 @@ void BeamFactory::calcPhysics(float dt)
 			m_trucks[t]->calcNetwork();
 			break;
 
-		case RECYCLE:
+		case NETWORKED_INVALID:
 			break;
 
-		case NETWORKED_INVALID:
+		case DELETED:
 			break;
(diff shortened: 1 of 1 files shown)


only-a-ptr:
It appears the truck state `RECYCLE` is not used anymore: https://github.com/RigsOfRods/rigs-of-rods/search?utf8=%E2%9C%93&q=RECYCLE. Please verify and remove.

only-a-ptr:
This needs some cleanup. What about the [`RECYCLED`](https://github.com/RigsOfRods/rigs-of-rods/blob/d1d3d0f3c626c80f49b745ef0636202334acdbff/doc/angelscript/global_functions.h#L50) truck state? It's [assigned](https://github.com/RigsOfRods/rigs-of-rods/blob/91e7690fba455df5dc6c5577ba9c96cb74fc4818/source/main/physics/Beam.cpp#L4557), but obviously it doesn't work as documented. Since the `calcPhysics()` function treats it as `default` and would re-attempt to delete it indefinitely, I'd say the value is outright dangerous. Please resolve.

PR #914, opened 2016-05-14
[CMake] Make boost_system optional

Hiradur:
If Ogre was built with Boost RoR needs to be linked to boost_system library. The idea here is that if boost_system is present on the system Ogre was likely compiled with Boost support. If boost_system is not present then Ogre was probably build without Boost support.

Note: this was not test on a system without Boost yet. I'd appreciate it if somebody could do that.

### DIFF
--- a/cmake/DependenciesConfig.cmake
+++ b/cmake/DependenciesConfig.cmake
@@ -179,6 +179,7 @@ endmacro(importLib)
 
 ELSEIF(UNIX)
    find_package(PkgConfig)
+   find_package(Boost COMPONENTS system)
    PKG_CHECK_MODULES  (GTK gtk+-2.0 REQUIRED)
    PKG_CHECK_MODULES  (GTK_PIXBUF gdk-pixbuf-2.0 REQUIRED)
    include_directories(${GTK_INCLUDE_DIRS})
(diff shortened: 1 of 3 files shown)


tritonas00:
Doesn't work in boost free system.

http://pastebin.com/A9s4RGub

Without this PR works, but it will fail at linking i suppose.

Hiradur:
@tritonas00 Is your system entirely boost free? I wonder why CMake claims it found Boost version 0.0.0 (which is suspicious).

tritonas00:
Yes, completely boost free

Hiradur:
I'm going to reinstall my system soon, then I will test with a complete boost-free environment.

susnux:
Maybe this is useful:
https://github.com/MyGUI/mygui/issues/15

So using the "official" FindOGRE will check for boost automatically if needed, just setup cmake to search for the correct FindOGRE file (on my Linux system /usr/lib64/OGRE/cmake/...) and it will include boost if needed.

tritonas00:
Current master builds almost fine in a boost free system

RoR: OK (boost and without boost)
RoRConfig: boost ok, without boost needs to remove `-lboost_system` from https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/configurator/CMakeLists.txt#L23 and it links fine

PR #918, opened 2016-05-14
[Feature] Added RoR.cfg option to skip the main menu

ulteq:
Appending `SkipMainMenu=Yes` to your RoR.cfg file allows you to skip the main menu during startup.


Hiradur:
I also thought of something like this but hotkeys for the main menu may be even better.

PR #924, opened 2016-05-16
Unified search for dependencies in build scripts

mikadou:
This implements a unified approach to search for dependencies on Linux and Windows.

It should work with dependencies that are
- installed in system paths (e.g. by Linux package manager)
- created by @Hiradur s build scripts for Linux
- generated with https://github.com/RigsOfRods/ror-dependencies/pull/1 (Linux & Windows)

For the latter two scenarios the `CMAKE_PREFIX_PATH` needs to be pointed to the location where the dependencies are located on the filesystem.


tools/travis/linux/build.sh
```diff
@@ -4,6 +4,7 @@ mkdir -p ./build
 cd ./build
 
 cmake -DCMAKE_PREFIX_PATH=$DEPS_INSTALL_DIR \
+-DBoost_USE_STATIC_LIBS=FALSE \
```
mikadou: Forgot to remove :blush:

ulteq:
Works fine for me.

only-a-ptr:
Just tested on Win7x64, VS2015update2, cmake-3.5.2-win32-x86
Configuring worked fine, except DirectX9 wasn't found (My DxSDK is under non-standard path "D:\Lib\DirectX_SDK_June_2010" and it isn't even installed) and the path setting wasn't visible in CMake-gui.

Logs and cache files:
https://www.dropbox.com/sh/bn3qm1cvdopy73k/AAD-XO0v8S_zjYnwbm_lY1Sya?dl=0

mikadou:
@only-a-ptr thanks alot for testing!

> I added CMAKE_PREFIX_PATH as described above. Quite un-intuitive; could use a hint directly in CMake-gui output window.

I agree with this step being quite unintuitive. Setting `CMAKE_PREFIX_PATH` is the commonly used mechansim to expand the search paths to non-default installation locations. Specifying this as an option on the commandline is great, but cumbersome in the GUI (because one manually needs to add the entry). Maybe it makes sense to add a `ROR_DEPENDECIES_PATH` cache variable which's content will be appended to the `CMAKE_PREFIX_PATH`? Suggestions welcome.

> Configuring worked fine, except DirectX9 wasn't found (My DxSDK is under non-standard path "D:\Lib\DirectX_SDK_June_2010" and it isn't even installed) and the path setting wasn't visible in CMake-gui.

I assume you stumbled accross this error:

    CMake Warning at source/main/CMakeLists.txt:543 (message):
      Could not copy RenderSystem_Direct3D9.dll to output directory!
    Call Stack (most recent call first):
      source/main/CMakeLists.txt:553 (post_build_copy_commands)

I assume this problem is not inside the build system of RoR itself, but already caused during building of the ror-dependencies project. It seems Ogre was built without Direct3D9 support (probably because it wasn't found in the non-standard path). Could you provide a listing of your `D:/Lib/ror-dependencies-git-build/Dependencies_Windows_Visual-Studio-14-2015/lib/OGRE` directory?

Here is how Ogre searches for DirectX: https://github.com/OGRECave/ogre/blob/master/CMake/Packages/FindDirectX.cmake#L29-L39
You could try to build ror-dependencies again. This time run `cmake-gui` from an environment where you set the environment variable `ENV_DXSDK_DIR` to `D:\Lib\DirectX_SDK_June_2010`.

Anyway it seems that configuration otherwise finished succesfully. Did you try building it?

only-a-ptr:
> Anyway it seems that configuration otherwise finished succesfully. Did you try building it?

No, I went to bed instead :smile: I'll try later today.

> I assume this problem is ... caused during building of the ror-dependencies project

Makes sense. I'll look into it when I get home.

Hiradur:
Although this is great I dislike that we'd have to maintain our own CMake modules for the dependencies.

mikadou:
@Hiradur I understand your point, but I don't believe this should be considered as an argument against this PR.

The recommended way to do dependency searching in CMake is the `find_package` command (probably the best source of information is https://cmake.org/cmake/help/v3.0/manual/cmake-packages.7.html). The `find_package` command either makes use of a so called **Config-file** provided by the corresponding upstream dependency (I don't think any of our current deps does this) or relies on a so called **Find-module** (which is a script that searches for the dependency). CMake ships readily with Find-modules for many common dependencies (such as OpenAL).

The current dependency search implemented in RoR's build system doesn't consistently use the `find_package` command and differentiates between Windows and Linux. For Linux it uses the `find_package` command to find packages for which CMake itself provides a Find-module, other dependencies are found using the `pkgconfig` wrapper script. Worse on Windows, paths are hard-coded.

As a bottomline, this PR consequently uses the `find_package` command for dependency search. Granted, in order to make things work, I had to implement some Find-modules myself. However these are quite trivial. On the other hand, I copied the available Find-module from upstream if one was available (e.g. FindOgre.cmake). This still requires to manually update the Find-module in our repo if it changes upstream, so really the ideal solution would be that every dependency provides a Config-file on its own. And I agree with you, that we might want to push this. Until this vision becomes true, however, providing our own Find-modules is the best we can do, and shouldn't be too hard to maintain (at least not worse than a mix of pkgconfig and hard-coded path searching).

To sum up, I'm convinced this PR puts us in a very good position for the future. Once upstream dependencies provide a Config-file we simply can remove the corresponding Find-module from our repository and everything will continue to work just fine.

Hiradur:
:+1:
I'm fine with merging it. For the long-term we should convince the dependency developers to ship their own Find-modules. This would not only benefit us but also other projects.

Hiradur:
When do we want to merge this? I think this could greatly simplify the build process on OS X where proper Find-modules are missing.

only-a-ptr:
@AnotherFoxGuy We should merge it as soon we're ready to go along with it. This means the continuous integration infrastructure is prepared and all developers are able to use the new deps.

I'd like to merge as soon as possible, but last time I tried, I had issues with it on windows - some dependencies were never detected, and IIRC even hardcoding paths in CMake scripts didn't help. But it's been a while. I'll try again in the coming days.

Regarding the conflicts, I'm going to rebase the branch and create new PR when we're ready.

only-a-ptr:
Closing this in favor of updated version: https://github.com/RigsOfRods/rigs-of-rods/pull/1328

PR #926, opened 2016-05-16
[Feature] Overworked the truck spawning logic

ulteq:
* Trucks now spawn centered with their rotational center

* Trucks now auto avoid other trucks when spawned


only-a-ptr:
:+1:

Issue #905, opened 2016-05-13
Segmentation fault in DashBoardManager::windowResized


ulteq:
Steps to reproduce:
* Connect two clients (A and B) to a server
* Spawn a truck on client A
* Spawn a truck on client B
* Delete the truck on client A
* Move the mouse back into client B

```
Thread 1 "RoR" received signal SIGSEGV, Segmentation fault.
DashBoardManager::windowResized (this=0x0) at /home/archie/rigs-of-rods/source/main/gui/DashBoardManager.cpp:212
212		for (int i=0; i < free_dashboard; i++)
(gdb) bt
#0  DashBoardManager::windowResized (this=0x0) at /home/archie/rigs-of-rods/source/main/gui/DashBoardManager.cpp:212
#1  0x00000000007786e9 in BeamFactory::windowResized (this=0x2ec6370) at /home/archie/rigs-of-rods/source/main/physics/BeamFactory.cpp:998
#2  0x0000000000687986 in RoR::GUIManager::windowResized (this=0x23ff1c0, rw=0x7ffff7e731d8) at /home/archie/rigs-of-rods/source/main/gui/GUIManager.cpp:209
#3  0x00007ffff7a7ab1d in ?? () from /usr/lib/libOgreMain.so.1.9.0
#4  0x00007ffff7a7af6c in Ogre::WindowEventUtilities::messagePump() () from /usr/lib/libOgreMain.so.1.9.0
#5  0x00000000005607e5 in RoR::MainThread::EnterGameplayLoop (this=this@entry=0x7fffffffe420) at /home/archie/rigs-of-rods/source/main/MainThread.cpp:850
#6  0x00000000005668b8 in RoR::MainThread::Go (this=this@entry=0x7fffffffe420) at /home/archie/rigs-of-rods/source/main/MainThread.cpp:458
#7  0x000000000053d91a in main (argc=<optimized out>, argv=<optimized out>) at /home/archie/rigs-of-rods/source/main/main.cpp:239
```

only-a-ptr:
If I understand it correctly, the problem is client B trying to update dashboard on networked truck A which was deleted in the meantime. This feels like a bug in truck-removal code: the dashboard should have been removed along with client-A-truck.

PR #931, opened 2016-05-17
[Bugfix] Fixes 'nullptr' dereferencing

ulteq:
Fixes: https://github.com/RigsOfRods/rigs-of-rods/issues/905


only-a-ptr:
This is an [uglyhack]. I hate this "blind nullptr checking" - it says "Devs don't know what's going on in their code, so they step carefully".

Like I commented in the previous PR, the problem is one part of the code deletes the dashboard while other part tries to update it. The only right solution is to add a mechanism to queue the dashboard destruction and make the _responsible_ code delete it.

PR #934, opened 2016-05-17
[Cleanup] Removed unused class 'LobbyGUI'

ulteq:
This also removes the unused class 'IRCWrapper', including the IRC library.


only-a-ptr:
:+1: :100:

PR #935, opened 2016-05-17
[Cleanup] Untangled the 'Streamable' packet queue manipulation

ulteq:


only-a-ptr:
Kudos for fightting the `Stream*` monster. Cleanups are good. But I have some notes.

source/main/network/Streamable.cpp
```diff
@@ -39,6 +39,8 @@ Streamable::~Streamable()
 void Streamable::addPacket(int type, unsigned int len, char* content)
 {
 #ifdef USE_SOCKETW
+	std::lock_guard<std::mutex> lock(m_send_work_mutex);
```
only-a-ptr: Ahem.... commit says **[Cleanup]**
Adding a new mutex is totally not a cleanup

source/main/physics/Beam.cpp
```diff
@@ -4498,13 +4496,14 @@ void Beam::updateDebugOverlay()
 
 void Beam::updateNetworkInfo()
 {
+	if (!gEnv->network) return;
```
only-a-ptr: No no _no_ NO! :angry: 
This is just so damn wrong. It's our codebase, our function, yet we're unable to call it right, so we check with a global state? Sorry but this is one of the worst kinds of bullshit RoR is plagued with. Worse yet, it comes in a **[cleanup]** commit ... seriously?

ulteq: I did not add that line, I only moved it to the top of the function.

source/main/network/Streamable.cpp
ulteq: Better than having a race on the packet queue?

source/main/physics/Beam.cpp
only-a-ptr: Oh, sorry... I really should stop skimming thru the code from work :blush:

source/main/network/Streamable.cpp
only-a-ptr: > Better than having a race on the packet queue?

Absolutely, but that's a honorable [bugfix], not a cleanup.

source/main/physics/Beam.cpp
ulteq: No problem.

only-a-ptr:
Now it's perfect.
Keep up the good work.

Issue #27, opened 2015-01-15
Multiplayer: Unused Vehicles Teleported to 0,0,0


graysonk95:
In multiplayer, as users leave the server, their vehicles are teleported to 0,0,0 and deactivated, rather than being removed.
After a while on one server, this pile of vehicles can become massive and really hurt performance.

![glitch2](https://cloud.githubusercontent.com/assets/10550037/5762155/35480a10-9cae-11e4-9003-67e059bfa55c.JPG)

Forum thread: http://www.rigsofrods.com/threads/118133-Vehicles-Teleported-to-0-0-0?p=1368918&viewfull=1#post1368918

Max98:
I will take a look as soon as only_a_ptr will merge his multiplayer fixes into the main repo.

Aperion:
Could this be a server issue ?

Max98:
I don't think so, i still didn't take a look, will do so soon.

only-a-ptr:
I'm not sure whether this is a bug or feature :( 

@graysonk95 When did you first observe this? Hasn't it always worked this way?

Anyway, sorry I've been overlooking this for so long

Hiradur:
Isn't this client-side only? As in if you restart and log in again the trucks aren't there anymore?

ulteq:
> Isn't this client-side only?

Yes it is.

PR #936, opened 2016-05-17
[Codechange] Implements proper truck deletion in multiplayer

ulteq:
Fixes: #27

### DIFF
--- a/source/main/physics/BeamData.h
+++ b/source/main/physics/BeamData.h
@@ -143,9 +143,7 @@ enum {
 	GOSLEEP,        //!< active but ordered to sleep ASAP (synchronously)
 	SLEEPING,       //!< not active, sleeping
 	NETWORKED,      //!< not calculated, gets remote data
-	NETWORKED_INVALID,	//!< not calculated, size differs from expected
-	RECYCLE,        //!< waiting for reusage
-	DELETED,        //!< special used when truck pointer is 0
+	NETWORKED_INVALID //!< not calculated, size differs from expected
 };
 
 enum {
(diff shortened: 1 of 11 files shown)


only-a-ptr:
You're calling `NetworkStreamManager::getSingleton().removeStream()` from 2 places. It seems the call inside destructor is a safety net in case `calcPhysics()` fails to do it's job. That's not the way to go. Can't we come up with a reliable mechanism for deleting networked trucks?

Further, `NetworkStreamManager::getSingleton().removeStream()` looks like it might throw an exception. Beware: exception thrown in destructor = program terminated! http://stackoverflow.com/a/130123

ulteq:
> You're calling `NetworkStreamManager::getSingleton().removeStream()` from 2 places.

Yes, for two different cases.

a) `state != NETWORKED` -> truck is local
* Truck is deleted here: https://github.com/RigsOfRods/rigs-of-rods/blob/688814c016b4f0abd85c398601aa6f147609709b/source/main/physics/BeamFactory.cpp#L701

b) `state == NETWORKED` -> truck is remote
* Truck is deleted here: https://github.com/RigsOfRods/rigs-of-rods/blob/688814c016b4f0abd85c398601aa6f147609709b/source/main/network/StreamableFactory.h#L169

> Beware: exception thrown in destructor = program terminated!

I've put it into the destructor, because the stream itself is added in the constructor:
https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/physics/Beam.cpp#L1568

only-a-ptr:
> Yes, for two different cases.

Oh, right, `(gEnv->networking != nullptr)` means "we're in multiplayer, each local truck has streams attached".

> I've put it into the destructor, because the stream itself is added in the constructor:

Hmm, as I can see, the whole network setup is plagued with the `if (!gEnv->network) return;` garbage. Anyway, bulky constructors with side effects, like `Beam()`, are wrong. The more we remove from there, the better. Factory methods (create/destroy) are the way to go.

source/main/physics/BeamData.h
```diff
@@ -144,7 +144,6 @@ enum {
 	SLEEPING,       //!< not active, sleeping
 	NETWORKED,      //!< not calculated, gets remote data
 	NETWORKED_INVALID,	//!< not calculated, size differs from expected
-	RECYCLE,        //!< waiting for reusage
```
only-a-ptr: :100:

only-a-ptr:
:+1:

PR #937, opened 2016-05-18
[Cleanuped] Removed unused TwoDReplay feature

ulteq:


only-a-ptr:
How do we know it's unused? It sure appears unused, I never heard about anyone using it, but I'd still prefer asking some old member (Aivar, maybe) about it's purpose.

ulteq:
> I'd still prefer asking some old member (Aivar, maybe) about it's purpose.

Agreed.

AnotherFoxGuy:
I think it was meant to export replays to an external program as a proof of concept, but I have never found the program it was meant to export to.

ulteq:
The visualization was done with javascript in the web browser.

only-a-ptr:
Question asked in the open: http://rigsofrods.org/thread-182.html

PR #945, opened 2016-05-19
[Cleanup] Minor network cleanup

ulteq:
* Simply run `GUI_Multiplayer::update` every two seconds
* Replaced mutex construct with std::atomic
* Removed duplicate `getUID()` method
* Only hold `m_send_work_mutex` when needed

### DIFF
--- a/source/main/MainThread.cpp
+++ b/source/main/MainThread.cpp
@@ -918,39 +918,23 @@ void MainThread::MainMenuLoopUpdate(float seconds_since_last_frame)
 		return;
 	}
 
-/*	if (RoR::Application::GetGuiManager()->getMainSelector()->IsFinishedSelecting())
-	{
-		RequestExitCurrentLoop();
-	}*/
-
 	// update GUI
 	RoR::Application::GetInputEngine()->Capture();
 
-	// update network gui if required, at most every 2 seconds
+#ifdef USE_SOCKETW
+#ifdef USE_MYGUI
+	// Update network gui every two seconds
 	if (gEnv->network)
 	{
 		netcheck_gui_timer += seconds_since_last_frame;
 		if (netcheck_gui_timer > 2.0f)
 		{
-#ifdef USE_MYGUI
-#ifdef USE_SOCKETW
-			if (BeamFactory::getSingleton().checkStreamsResultsChanged() || gEnv->network->getNetQualityChanged())
-			{
-				GUI_Multiplayer::getSingleton().update();
-			}
-#endif // USE_SOCKETW
-#endif // USE_MYGUI
+			GUI_Multiplayer::getSingleton().update();
 			netcheck_gui_timer = 0.0f;
 		}
-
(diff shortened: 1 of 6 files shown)


only-a-ptr:
:-1: All the individual items in the commit's description should have been separate commits. For example `Net quality: replaced explicit mutex with std::atomic`

ulteq:
> All the individual items in the commit's description should have been separate commits.

Agreed.

> Further, as I already said - right now we don't need minor cleanups, but a plan.

First throwing away as much junk as possible feel right for me.

only-a-ptr:
> First throwing away as much junk as possible feel right for me.

I see, but there's a risk of wasting time throwing junk around instead of out.

ulteq:
> but there's a risk of wasting time

Yes, but I can at least control the amount of time I spend.

only-a-ptr:
> Yes, but I can at least control the amount of time I spend.

True, but the measure of time spent shouldn't be 'amount', but 'quality'. If you don't have a plan, then you can't know whether your time spent has a purpose. Specifically with the networking, you can try cleaning it up the way it is, but since the design itself is a mess, you're bound to just keep refactoring back and forth until it settles, somehow. That's a plausible strategy, sure, but we can do better.

ulteq:
My plan is to finish this: https://github.com/RigsOfRods/rigs-of-rods/pull/936, which means adding a `MSG2_STREAM_UNREGISTER` message. That's the only reason I'm messing with the current code.

only-a-ptr:
> My plan is

In that case you're fragmenting your effort into 2-3 PRs for no good reason. As I said before, I don't see much sense in commiting "test code with 5sec placeholder" first rather than experimenting with the message directly.

ulteq:
> In that case you're fragmenting your effort into 2-3 PR

Yes, because I don't want to have both cleanup and feature commits in a single PR.

> I don't see much sense in commiting "test code with 5sec placeholder" first rather than experimenting with the message directly.

I won't merge the 5 second placeholder.

It's only there to test if `NetworkStreamManager::getSingleton().removeStream()` works as intended.

only-a-ptr:
> I don't want to have both cleanup and feature commits in a single PR

That's good, but either way, bugfixing and refactoring at the same time doesn't really work. Not because of code organization, but because of mind organization. I'd say you should focus on adding the `MSG2_STREAM_UNREGISTER` now and then do a full-scale overhaul of the networking. You can do it.

> I won't merge the... It's only there to test...

OK, I see your work pattern now. I'll just let you code and stop drawing premature conclusions. However, I suggest you mention "this is WIP" in the PRs so this is clear.

ulteq:
https://github.com/RigsOfRods/rigs-of-rods/pull/936#issue-155382932

```
Work in progress:

TODO: Implement MSG2_STREAM_UNREGISTER messages
```

> I'd say you should focus on adding the MSG2_STREAM_UNREGISTER now and then do a full-scale overhaul of the networking.

Agreed.

PR #971, opened 2016-05-24
[Cleanup] Replaced hand written 'sleepMilliseconds' with C++11 'sleep_for'

ulteq:


only-a-ptr:
:+1: At first I was like :anguished:, but then I :relaxed:

PR #1005, opened 2016-05-31
[Tools] Travis: Use gcc 6

Hiradur:


tritonas00:
Nice. gcc6 is lot faster

Hiradur:
... and it has colored output which is great.

Hiradur:
Note to self: Travis needs to pull a newer version of Ogre 1.9.

ulteq:
:+1:

PR #1009, opened 2016-06-01
[Feature] 'LALT' modifies the character rotation speed

ulteq:


only-a-ptr:
When did you get this idea?

Personally I always thought the accelerated character movement in multiplayer is on purpose - otherwise it moves reallistically, but takes ages to get anywhere.

And ~20 minutes is not really enough time to discuss a PR :smile:

PR #1030, opened 2016-06-04
New loading pictures

Hotrod55:
This replaces and removes some loading pictures, some of the old ones had issues such as mods without original author's permission, fps counter in the corner, wrong resolution, etc. 
I have also added a README which contains links to most of the mods used in the pictures, not complete though.
New screenshots:
loading_02
loading_03
loading_13
loading_16
loading_20


ulteq:
> I have also added a README which contains links to most of the mods used in the pictures

:+1:

mikadou:
The 4th picture looks plain awesome! The first one is too dark on my screen, can hardly see anything.

This PR reminded me of Blender which uses an individual 'splash screen' for each release. Wouldn't this be feasible for RoR, too? It would visually distinguish a newer release from previous versions on startup.

We could either use a single image per release, or a set of images (as it is done now).

ulteq:
Would it be possible to also bring back this one, but without the fps meter?:

Hotrod55:
Thanks for your feedback!
@mikadou I've replaced loading_02.jpg with something a bit more 'green' :smile: 

@ulteq 
> What does that mean?

In loading_10.jpg, there is a modified Chevy S10. which `gabester` doesn't usually give permission for mods of his vehicles. plus I've seen that truck shown around the old Skype groups. It's best to just have unmodified/released vehicles in the loading pictures. 
> Can't we cut them to fit the resolution and get rid of the fps counter at the same time?

I've edited out the fps meter for loading_22.jpg, I'll remove the fps counter on loading_18.jpg soon. :smile:

Speciesx:
loading_18

Hiradur:
:+1: for the thorough documentation.
I love wallpaper three.

only-a-ptr:
:+1: Very nice initiative. I also like the attention to credits and used content.

Mikadou's "version specific background" idea sounds good, but doesn't go together with our "random image" mechanism. It would make sense if it demonstrated new content/features and we had a slideshow-style background.

For now, let's just keep partially the backgrounds for every release.

DirtGamer301:
While we're at loading pictures in general, I'll just drop my idea here to (if wanted) add custom loading pictures for maps, maybe uploaded somewhere on the internet not to increase the filesize so much and link it to the mod in the terrn2 file for example.

DarthCainRor:
The one with the Kenworth T800 semi should also probably be replaced. It was a project someone on the old forums was doing as a mod, with my permission, using the T800SH I released. But I doubt it'll ever be released.

PR #1039, opened 2016-06-05
[Feature] Implemented a simple terrain editor

ulteq:
Controls:

* Toggle Editor `CTRL+Y` (`editor_out.cfg` is overwritten when you leave the terrain  #editor)
* Toggle object auto tracking (aligns the character with the currently selected object) `SPACE`
* Enter nearest object (in relation to character / free camera position) `Enter`
* Translate / Rotate object (similar to advanced truck reset)
* Cycle through all terrain objects `CTRL+[` and `CTRL+]`
* Cycle through all rotation axes `r`
* Reset the current object `i`

Structurally complete, but far from perfect.

### DIFF
--- a/bin/resources/skeleton/config/input.map
+++ b/bin/resources/skeleton/config/input.map
@@ -185,7 +185,7 @@ COMMON_RESET_TRUCK             Keyboard             EXPL+I
 COMMON_SCREENSHOT              Keyboard             SYSRQ 
 COMMON_SECURE_LOAD             Keyboard             O 
 COMMON_SHOW_SKELETON           Keyboard             K 
-COMMON_START_TRUCK_EDITOR      Keyboard             SHIFT+Y 
+COMMON_TOGGLE_TERRAIN_EDITOR   Keyboard             EXPL+CTRL+Y 
 COMMON_TOGGLE_CUSTOM_PARTICLES Keyboard             G 
 COMMON_TOGGLE_MAT_DEBUG        Keyboard             EXPL+CTRL+SHIFT+F 
 COMMON_TOGGLE_RENDER_MODE      Keyboard             E 
--- a/source/main/gameplay/RoRFrameListener.cpp
+++ b/source/main/gameplay/RoRFrameListener.cpp
@@ -506,7 +506,7 @@ bool RoRFrameListener::updateEvents(float dt)
 	static std::vector<TerrainObjectManager::object_t> object_list;
 	static bool terrain_editing_track_object = true;
 	static bool terrain_editing_mode = false;
-	static int terrain_editing_rotation_axis = 2;
+	static int terrain_editing_rotation_axis = 1;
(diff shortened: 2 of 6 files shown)


tritonas00:
Nice! Toggle Editor is SHIFT+Y

corvette81:
This is great, hope to see this feature and more evolve and help make the process of content creation easier. Keep it up guys.

ulteq:
> Toggle Editor is SHIFT+Y

I fixed it. It now is `EXPL+CTRL+Y`.

only-a-ptr:
:-1: 

Once again there's the conflict between wonderful and thrilling but hacky new features vs. organized strategic development.

In-game editor is a good idea. It's always at-hand, it shares 3D rendering/resources infrastructure with the game (less coding for us, less config for the user). Disadvantages: it only works for heightmapped terrains which may or may not be a future. A tool for making meshed terrains would best be made as Blender plugin.

However, the code you presented is *absolutely* unacceptable. You're mixing editor logic into our main simulation loop, which is a giant step back - we're trying to clean up and organize the main loop better, not to clutter it further. The idea of "hey, let's run one loop and then do wonky branching to determine what it actually does" is a plague of RoR which we're trying to eliminate, not to expand further. Further, static vars in function? Eww! Yeah, the frame listener is de-facto singleton, but come on! This isn't what I had in mind with the `RoR::Network*`-style design. Finally, our map system suffers of `***Manager` infestation, which is a subject for a cleanup, not extension.

We agreed to split our logic to visuals-only and sim-only, so the future code should look something like this:

```
class VisualScene
{
    struct SceneObject { Ogre::Mesh* mesh, Ogre:;Entity* entity };

    Ogre::Terrain* m_ogre_terrain;
    std::vector<MapObject> m_objects;
};

class PhysicsWorld
{
    StaticObject* m_ground;
    std::vector<StaticObject> m_objects;
};
```

VERDICT: It would be nice to have an in-game map editor, but not this way. If you really want to introduce and support one, I suggest you refactor it like this:
1. Introduce a new [application state](https://github.com/ulteq/rigs-of-rods/blob/master/source/main/Application.h#L45-L52) and create a corresponding class, `MapEditor` with it's own main loop.
2. Add it to our main menu loop, similarily how RigEditor used to be: https://github.com/ulteq/rigs-of-rods/blob/master/source/main/MainThread.cpp#L340-L445
3. Duplicate all map-loading code you need - you may use existing functions, but only as long as they don't require any modification
4. Move your map-editing code into `class MapEditor`

tritonas00:
> :-1: 

Party pooper :P

only-a-ptr:
OK, you guys convinced me to approve this PR. 

I talked to Ulteq in the dev chat and he agreed to refactor it when he has time. But he's busy now, so let's just merge it as-is and refactor it later. I might do it myself.

:+1:

PR #1062, opened 2016-06-25
[Fix] Special prop dashboard backward compatibility

mikadou:
Adresses the bug reported here: http://forum.rigsofrods.org/thread-263-post-2420.html#pid2420


only-a-ptr:
:+1:

PR #1064, opened 2016-07-01
[Fix] Build on OS X

Hiradur:
Thanks to @chilledfrogs for testing and helping.

- RoRConfig doesn't build yet
- only tested without the optional dependencies so far


chilledfrogs:
Still working on build instructions, which are for the moment partially updated...

PR #1074, opened 2016-07-13
Main character updated animations

vido89:
I could not update jump animation, hope u will like it


Hiradur:
Could you add a detailed list of what you changed?

vido89:
- added sidestep animation and adjusted speed
- updated Spot_swim animation and adjusted speed
- updated Idle_sway animation and adjusted speed
- updated Driving animation

bin/resources/meshes/Character.cpp
```diff
@@ -1,704 +0,0 @@
-/*/
```
Hiradur: This commit is ugly, can you remove it and clean up the other one so that `Character.cpp` doesn't get added to `bin/resources/meshes/Character.cpp`?

vido89: Im not sure how to delete last commit

Issue #450, opened 2015-11-15
Some fonts don't comply with the GPL


Hiradur:
Some fonts are not entirely free and thus don't comply with the GPL.
Namely:
```
bluebold.ttf
bluecond.ttf
bluehigh.ttf
Cyberbit.tff
digital-7.mono.ttf
micross.ttf
```

Some of these fonts may not be used at all, this has to be investigated. However, the Cyberbit font seems to play a major role for localization.

Among licensing issues this also means that RoR can't be packaged for the Debian "main" software repository right now since these fonts break with the Debian Free Software Guidelines ( http://www.debian.org/social_contract ).

Hiradur:
From a quick fulltext search I think it'd be safe to remove these fonts as I don't think they are used anywhere:
```
bluebold.ttf
bluecond.ttf
bluehigh.ttf
micross.ttf
```

``` digital-7.mono.ttf ``` appears to be used by dashboards and ```Cyberbit.tff``` is used in many places so we need a replacement font with a free/libre license for these.
We already have VeraMono so that may be a candidate. Needs investigation. Replacing Cyberbit.ttf won't be so easy since it's used for localization and as such we need a font with a large character set.

Hiradur:
After replacing ```Cyberbit.tff``` it's important to check if and how code ranges* for localization are affected.

* more about code ranges here: http://www.rigsofrods.com/wiki/pages/Language_Translation#How_it_works

only-a-ptr:
@Hiradur Thanks for looking into this.

"BlueHighway" sounds familiar. Either it used to be part of Ogre sample media, or it used to be part of CEGUI. I remember seeing the name in Ogre logs when doing tutorials.
Anyway, we're not the only ones having this concern: http://forum.freegamedev.net/viewtopic.php?p=59035&sid=caf6519b31a7ae55ebad8e7245dc90aa#p59035

Hiradur:
@only-a-ptr Yes, they were used for OGRE samples and have since been removed from OGRE due to licensing issues: https://ogre3d.atlassian.net/browse/OGRE-76

Hiradur:
The Vera fonts can't replace Cyberbit because they contain only latin characters. 
The DeJaVu fonts are often proposed but don't contain the chinese characters for example (http://dejavu.sourceforge.net/samples/DejaVuSans.pdf )

The biggest free font is GNU Unifont:
http://unifoundry.com/unifont.html

[...]

PR #1075, opened 2016-07-14
[Fix] Fixed vehicle info not fitting in the info box

### LINKED ISSUE #1068: Vehicle Info not showing completly
Hello again...

This issue is actually there since a few Night Build.

I want you developers to make this issue a priority.
image

There's a few information such as damage %.

AnotherFoxGuy:
Fixes #1068

screenshot_2016-07-14_12-47-06_2


tools/linux/l10n/from_transifex.sh
```diff
@@ -13,4 +13,6 @@ do
 	echo "Now copying $file"
 	mkdir -p ../../../languages/$file/LC_MESSAGES/
 	cp $file.po ../../../languages/$file/LC_MESSAGES/ror.po
+	cp ../../../SourceSansPro-Regular.ttf ../../../languages/$file/LC_MESSAGES/SourceSansPro-Regular.ttf
```
Hiradur: This seems inefficient to me, can't you just put it in the `resources` folder?

tools/linux/l10n/SourceSansPro-License.txt
```diff
@@ -0,0 +1,91 @@
+Copyright (c) 2010, Adobe Systems, Inc. (http://www.adobe.com/)
```
Hiradur: This text should be appended to DEPENDENCIES.md

Hiradur:
It's good but  having to copy the MyGUI_FontConfig.xml file to every language folder still seems to be inefficient. Why aren't we using SourceSansPro-Regular.ttf fo English as well so one MyGUI_FontConfig.xml file in the bin/resources/mygui folder should be enough?
One issue with SourceSansPro is that it's rather limited. As far as I can tell it doesn't have e.g. kyrillic, arabic or asian characters. IMO GNU Unifont would be a good choice as a "one for all" font. Also see #450

ulteq:
@AnotherFoxGuy, can you comment on what @Hiradur said?

AnotherFoxGuy:
>Why aren't we using SourceSansPro-Regular.ttf fo English as well so one MyGUI_FontConfig.xml file in the bin/resources/mygui folder should be enough?

I've moved SourceSansPro-Regular.ttf and MyGUI_FontConfig.xml to the resources folder.

ulteq:
> Edit: it seems that Travis is broken?

Nah, it works now.

Issue #1094, opened 2016-08-07
"make: *** No rule to make target 'install'.  Stop."


krichter722:
`cmake . && make && sudo make install` fails due to `make: *** No rule to make target 'install'.  Stop.`.

experienced with 0.4.5.1-1063-g64ad6f6

only-a-ptr:
Is this still relevant? Update, please.

PR #1095, opened 2016-08-07
added 'sudo make install' to .travis.yml

krichter722:
#1094 shows that such a check might be useful.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: cpp
-sudo: false
+sudo: required
 
 cache:
   directories:
--- a/tools/travis/linux/build.sh
+++ b/tools/travis/linux/build.sh
@@ -24,4 +24,6 @@ then
   make doc-doxygen > /dev/null 2>&1
 fi
 
+sudo make install
+
 cd ..


Hiradur:
How is #1094 related to this PR at all? The Travis CI builds work. Adding `sudo: required` would mean that they would fall back to the slow legacy infrastructure which is an absolute no go. The Travis CI build scripts are not meant for local building of RoR, you should use https://github.com/RigsOfRods/ror-linux-buildscripts instead.

krichter722:
>  The Travis CI builds work.

But only because it doesn't test the `install` target which fails/doesn't exists (see #1094). Apparently you don't have that, i.e. RoR can't be installed/has to be used from the build directory (without the mentioned scripts). That's very odd, but fine.

> Adding sudo: required would mean that they would fall back to the slow legacy infrastructure which is an absolute no go.

It should be possible to install in a location which doesn't require `sudo` and skip the change to legacy infrastructure, but that only makes sense if there's an `install` target.

Hiradur:
> But only because it doesn't test the install target which fails/doesn't exists (see #1094). Apparently you don't have that, i.e. RoR can't be installed/has to be used from the build directory (without the mentioned scripts). That's very odd, but fine.

I can't look into this but there's always been an `install` target and it always worked. It may be an issue on your end.

You can install to a custom location by setting `CMAKE_PREFIX_PATH` prior to the cmake run. The build scripts linked above install RoR this way.

PR #1111, opened 2016-09-02
Updated main character animations: Walk,Run,Idele,Swim,Turn,Sidestep

vido89:
-Updated Walk,Run,Swim_loop,Spot_swim,Idle_sway,Driving animations
-Add Sidestep animation
-Updated speed of animations


vido89:
Great Im looking forward to see this

only-a-ptr:
Everything's fine except the "standby" animation. RoRBot's neutral position here is "leaning strongly on left" instead of standing straight. It's very unpleasant when walking and stopping - the straight walk posse changes in the left-leaning standby pose.

I'm using OgreMeshy 1.6 (http://www.yosoygames.com.ar/wp/2015/08/ogremeshy-1-6-released/) to check what's wrong.

vido89:
My intention was to make it rest on left foot and rest on right foot like when you stand and change rest position from one position to another, I dont know, main trouble is the animation not playing from first "frame" every time rather that playing (continue) from last frame

only-a-ptr:
I see, but since we switch directly from "walk/run" to "idle", using a non-straight default pose looks like a glitch.

Making any sophisticated idle animation would probably require tuning the code. Playing from last frame is really not optimal.

vido89:
So should I update Idle animation and remove current movment ?

vido89:
Also I forget to mention any animation is played like Idle but that is less noticeable and instead playing from they stopped they should be played from beginning each time

only-a-ptr:
I appreciate your effort to make the idle animation reallistic. Real people really sway like that when standing still. But under current conditions, your idle animation doesn't look right.

You can try to make all animations play from "0" each time by adding
```
as->setTimePosition(0.f);
```
here: https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/gameplay/Character.cpp#L249-L250 (Docs: http://www.ogre3d.org/docs/api/1.9/class_ogre_1_1_animation_state.html#ae69ff5a068d086be7ecc048a71ed8efc)

Or, simply remove the sway movement from the idle anim.

vido89:
I'll just make it more subtle, less movement will be better

tritonas00:
Not very good with some vehicles (try agora or citaro bus)
2016-09-07-200414_1920x1080_scrot

vido89:
Should be fixed now in agora bus, I dont have citaro bus but think it should be fine

tritonas00:
Great, agora now is fine. Citaro almost :p

2016-09-08-070918_1920x1080_scrot

https://www.dropbox.com/s/poef0st939mty5o/mercedes_citaro_K.zip?dl=0


Also it would be nice keeping legs/hands moving while walk/run+ rotate the character

vido89:
Fixed Citaro issue but Im not sure what you think when you say "keep legs/hands moving"

tritonas00:
turn the bot right/left while moving

Do hands and legs still animate?

vido89:
You mean turn left/right while walking, that would be W+A or W+D, Im not sure whats going on there as I can recall they worked earlier, I mean animation has played Im not sure why not playing now

vido89:
If I comment Character.cpp:251
				as->setTimePosition(0.f);
animation W+D all well played

ulteq:
Pretty nice! The swimming animations are superb.

The only real dealbreaker is what happens when you press W+A or W+D.

The character simply slides diagonally without moving the arms / legs at all.

Michael10055:
The idle animation just seems unnatural:

only-a-ptr:
@Michael10055 What's wrong with it?

Michael10055:
Have you compared the old ones?
Old animation: 

New animation:

vido89:
ror
Is this better

Michael10055:
Looks good now. 😄

Issue #1109, opened 2016-08-30
Videocamera and Camera "jitter" is back!


graysonk95:
Version 0.4.6.0-dev-5857a08
(Reincarnation of https://github.com/RigsOfRods/rigs-of-rods/issues/734 ?)
This issues has occurred before in these dev builds, and I made an issue of it and it was corrected. Now it seems to be back, but with a new twist. Props (not flexbodies) now seem to be updating very slowly in videocameras and it produces a jitter effect, best demonstrated by a video:
https://youtu.be/7UjXTiLF1Vo
Also, the interior camera seems jittery (or the vehicle itself is jittery, I can't tell yet).

Related or unrelated, it seems like the view distance of the videocamera mirrors is much shorter in this build. The sky reflects black, rather than blue.

Also, I'm glad to see legacy mirror support is here! I've been beating that drum since the day they were replaced by videocameras.

ulteq:
I can reproduce video camera stuttering. Changing the point of time when the video camera is updated might be enough to fix this.

> Camera "jitter" is back

I don't get any camera "jitter" in current upstream.

graysonk95:
@ulteq - Here's a video of the camera jitter, pay attention to the rear of the truck around 0:10, it is easily noticeable then.
https://youtu.be/WqqN85V5DpE
(This clip was recorded at 30fps, but the jitter is still present at 60fps with vsync enabled).

ulteq:
I can reproduce it. But I think it's the truck that jitters, not the camera.

ulteq:
I think the jitter is caused by the way we calculate the camera look at position:

```
https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/physics/Beam.cpp#L1220-L1226
```

graysonk95:
As an update, I was messing around in the nightly build from a few days ago and the shuddering as depicted in previous posts seems to have been minimized/resolved.
I did notice, however, that the Free Cam now moves in a very jumpy manner, not smooth like it used to be.

ulteq:
> I did notice, however, that the Free Cam now moves in a very jumpy manner, not smooth like it used to be.

The prop stuttering in the video camera screen is completely gone though?

-> I think we should close this and create a new issues for the camera bugs.

a) The (new) free camera issue
b) The orbit camera stuttering when you drive a circle

ulteq:
It could related to volatile frametimes.

Sudden FPS changes could cause the camera stuttering.

PR #1116, opened 2016-09-10
[Bugfix] Fixes videocamera stuttering

ulteq:
Related with: #1109


only-a-ptr:
Definitely a better way to do it, but this will make the reflected image 1 frame out of date, won't it? Not a big deal, but may be noticed under lower FPS.

ulteq:
I just reverted it to how it was: https://github.com/only-a-ptr/ror-legacy-svn-trunk/blob/0a55a993899d6a06cdc58fd98b5b675b371eb89a/source/main/gameplay/RoRFrameListener.cpp#L3135

> this will make the reflected image 1 frame out of date

I don't think so. But I'm not sure about it.

Issue #1105, opened 2016-08-21
Camera keys are inverted in the lastest build


WillyB1337:
Hello there.
I found out really clearly that the keys for the Camera on the Numpad are Inverted
**EX : Pressing NUMPAD 4 (Turn camera left) turn right instead**

This issue also happens with the input file for my XBOX 360 controller.
I've tried to copy the default input.map from the Programs Files Directory.
I also tried a backup input.map I always keep, it still does the same thing.

Here's what in the **original input.map**

> CAMERA_ROTATE_DOWN             Keyboard             NUMPAD2 
> CAMERA_ROTATE_LEFT             Keyboard             NUMPAD4 
> CAMERA_ROTATE_RIGHT            Keyboard             NUMPAD6
> CAMERA_ROTATE_UP               Keyboard             NUMPAD8

FuturzMinion:
I don't like the way it works now, if I'm in the outside view and I press 4 on the numberpad, it moves right now. If I'm inside and I press 4, it moves right.

ulteq:
And how should it be like?

FuturzMinion:
It should be like: 4 is look left, 8 is look up, 2 is look down, and 6 is look right. That is not how it is in the game right now. It doesn't feel right being inverted.

vido89:
Its been fixed in latest build

ulteq:
I'm confused now.

ulteq:
@WillyB1337, what is wrong with the way it currently is?

If the camera is on the rotation center, then right means turn camera right (e.g. look to the right).

If the camera is off the rotation center, then right means rotate the camera to the right (e.g. look to the left).

I will leave it as is, unless most people vote against it.

Michael10055:
The camera rotation works exactly the same as all previous versions of RoR. I don't understand why it needs to be changed.

Michael10055:
Old issue, there's nothing wrong with the current camera keys.

PR #1117, opened 2016-09-10
[Bugfix] Inverted numpad camera rotation keys

ulteq:
Fixes: #1105


only-a-ptr:
:+1:

ulteq:
It looks like we need to handle this differently when you are inside of a car.

disloyalpick:
maybe write some small code to invert the camera controls when cockpit view is detected?

Issue #1047, opened 2016-06-09
Containers in Multiplayer spawn at 0, 0, 0 for other players


tritonas00:
Spawn a container in MP, you see it there fine. But the other players see it at 0, 0, 0 and the rorbot also stays there in crucifixion :P

http://i.imgur.com/OnMInXr.png

Also this sometimes happens with normal vehicles too

https://files.gitter.im/tritonas00/RtT1/2016-06-06-002722_1366x768_scrot.png

ulteq:
This could be solved by including the initial truck position and rotation into the stream setup message:
https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/network/protocol/rornet.h#L163-L185

only-a-ptr:
working on it: http://forum.rigsofrods.org/thread-458-post-4466.html#pid4466

graysonk95:
The issue with weird things happening at 0,0,0 has been around for as long as I can remember (and I've been using multiplayer since it was "fixed" for v37). This exact issue was especially bad about two or 3 years ago. Whenever I was playing with my private MP group vehicles were spawning at 0,0,0 and we couldn't see each other and all had to keep reloading RoR until we all could see each other and we had to deal with the giant crucified RoRbot.
I'm glad to see this is being worked on.

I made this 0,0,0 issue not too long ago and it has since been closed: https://github.com/RigsOfRods/rigs-of-rods/issues/27

ulteq:
#27 is unrelated to this new issue.

tritonas00:
Still valid

tritonas00:
the clients had somewhat different *.zip.

i downloaded the same as the other player, and the issue resolved.

PR #1118, opened 2016-09-13
[Bugfix] Fixes 'SLEEPING' vehicles spawn at (0,0,0)

ulteq:
Fixes: #1047

### DIFF
--- a/source/main/physics/BeamFactory.cpp
+++ b/source/main/physics/BeamFactory.cpp
@@ -974,11 +974,15 @@ void BeamFactory::update(float dt)
 		default:
 			if (m_trucks[t]->state != SIMULATED && m_trucks[t]->engine)
 				m_trucks[t]->engine->update(dt, 1);
-			if (m_trucks[t]->state == SIMULATED && m_trucks[t]->networking)
-				m_trucks[t]->sendStreamData();
-			if (m_trucks[t]->state < SLEEPING) {
+			if (m_trucks[t]->state < SLEEPING)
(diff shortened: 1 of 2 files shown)


source/main/physics/BeamFactory.cpp
```diff
@@ -974,11 +974,15 @@ void BeamFactory::update(float dt)
...
-				m_trucks[t]->sendStreamData();
-			if (m_trucks[t]->state < SLEEPING) {
+			if (m_trucks[t]->state < SLEEPING)
 			    m_trucks[t]->UpdatePropAnimations(dt);
-                        }
+			if (m_trucks[t]->networking)
+			{
+				if (m_trucks[t]->state == SIMULATED)
+					m_trucks[t]->sendStreamData();
+				if (m_trucks[t]->state == SLEEPING && m_trucks[t]->netTimer.getMilliseconds() < 10000)
```
only-a-ptr: Where did the 10s timeout come from? :grimacing:

ulteq:
> This doesn't change anything which affects vehicle position - what's going on here?

This only affects `NETWORKED` trucks in multiplayer.

> Where did the 10s timeout come from?

Gut instincts

only-a-ptr:
I don't understand this fix. It appears to stop position updates for 10sec each time the vehicle goes to sleep. But how does it help with spawn position?

ulteq:
This code only runs on the client that owns the truck.

And it causes the client to send stream data for `SLEEPING` trucks during the first 10 seconds of their lifetime.

only-a-ptr:
:+1:

PR #1120, opened 2016-09-13
[Cleanup] Simplifies creation of visual beams

ulteq:
Fixes visual glitch when ties become visible


only-a-ptr:
Hmmm... I like this change a lot, but I'm worried it will introduce some weird visibility glitch elsewhere. Or crash RoR at some unexpected point.

I've put the logic in there because 0.38x did it that way. IDK why, I've been putting `//wtf` comments around it since I came.

This needs extra testing with variety of vehicles and scenarios (attaching/detaching loads etc...)

ulteq:
> This needs extra testing with variety of vehicles and scenarios

Agreed.

only-a-ptr:
:+1:

Issue #1053, opened 2016-06-13
Collisions with multiple .otc files not working


AnotherFoxGuy:
_See http://forum.rigsofrods.org/thread-240.html_
The 0.4 terrain2 system has been hard coded to only load the collisions of the first otc

https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/terrain/TerrainGeometryManager.cpp#L301

AnotherFoxGuy:
@ulteq https://github.com/RigsOfRods/rigs-of-rods/commit/a8b8267eab7bfa235d54c14fd0e171e4939ce33b

PR #1140, opened 2016-10-07
[Bugfix] Brings back old heightfinder

ulteq:
Fixes: #1053

Temporary solution until our custom heightfinder supports multiple terrain pages


tritonas00:
This fixes the nhelens problem also.

Issue #498, opened 2015-12-22
Trailer total mass doesn't include load mass


Speciesx:
HD11
![ror_2015_12_22_14_05_12_111](https://cloud.githubusercontent.com/assets/1196375/11956024/3f3c9d18-a8b7-11e5-85b3-ed47830223c0.jpg)

Trailer and secured HD11
![ror_2015_12_22_14_05_22_507](https://cloud.githubusercontent.com/assets/1196375/11956027/3ff43c02-a8b7-11e5-965c-e51a0e37ed27.jpg)

PR #1142, opened 2016-10-08
[Bugfix] Proper destruction of inter-truck-beams

ulteq:
* Fixes the handling of inter-truck-beams (hooks and ties) on truck deletion.
* Fixes segfaults and state inconsistencies when trucks are removed or reset while they are hooked or tied to other trucks
* Fixes: #498

Side-effect: Skeleton view changes are now properly propagated through all interconnected trucks.


only-a-ptr:
:100: for dealing with the overcomplicated system of hooks/ties/ropes/interbeams ...

TBH I don't know exactly how it works (both before and after your fix) and I'm not going to study it now.

Issue #381, opened 2015-08-12
Tracked vehicles not working correctly.


DarthCainRor:
It appears as if tracked vehicles that use rotators to move slidenodes around the rails, are slightly broken. From what I can see it looks like the submeshes attached to the rotator part are no longer smoothly moving the slidenodes along, causing the tracks to jerk in unrealistic manners.

Vehicles tested with: 
 http://www.rigsofrods.com/threads/115690-Hitachi-EX1200-5?p=1347026&viewfull=1#post1347026

http://www.rigsofrods.com/threads/112719-volvo-EC140C

And a private tracked vehicle I'm working on.

Max98:
Can you try disabling Intra-Collision from RoRConfig->Advanced.

DarthCainRor:
That causes them to not work at all, as it needs to collide to move.

Max98:
Are you sure that you disabled intra and not inter?

DarthCainRor:
![](https://i.gyazo.com/b1931f1835a56ba1b26d0c1fd939cb0a.png)

Quite certain. The way tracks are designed they need intra vehicle collisions. It's how they work. Submeshes are attached to the rotator, with sit between slidenodes and push the nodes along the tracks.

DarthCainRor:
No further comments? :-/ I mean they're definitely not working correctly, and there is no error related to them in the log files. >.<

fidoj:
the error seems to be at the looping part of the slide rails, some of the nodes jump at the intersection causing shaking

DarthCainRor:
any word on this? been working on a few things with tracks, and it'd be nice if they actually worked :-/

DarthCainRor:
I'm still not certain tracked vehicles are operating 100% the same as in .38. I would definitely say they are definitely better now than they were in .4.5, and are semi usuable, but they're still no where near as smooth moving as they used to be.

only-a-ptr:
@DarthCainRor I completely understand your point of view, but fixing this isn't a matter of reverting some particular changes or looking into some specific area of simulation. From a developer's standpoint, the only way to restore 0.38's behavior is to roll back all changes since that version and never touch the code again. Or, we may tune the simulation to get closer to what it was before, but to do that, we need a lot more detailed description of the problem. Please do some more testing and describe what exactly is disturbing you.

[...]

PR #1146, opened 2016-10-11
[Bugfix] Fixes tracked vehicles

ulteq:
Fixes: #381

### DIFF
--- a/source/main/physics/collision/DynamicCollisions.cpp
+++ b/source/main/physics/collision/DynamicCollisions.cpp
@@ -228,8 +228,11 @@ void intraTruckCollisions(const float dt, PointColDetector &intraPointCD,
             intra_collcabrate[i].rate--;
             continue;
         }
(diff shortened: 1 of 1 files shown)


ulteq:
This would be a better solution than this: https://github.com/RigsOfRods/rigs-of-rods/pull/1145, if it works as good.

only-a-ptr:
I can't figure anything from the code. If testing shows an improvement, let's take it in.

AnotherFoxGuy:
This also fixes the Mercedes-Benz Atego w/ Comax rear flaps from spazzing out.

ulteq:
> I can't figure anything from the code.

This decreases the chance of overlooking a `contacter` <--> `cab` collision by introducing a 10 second (20000 physics steps) cool-down time for frequently hit `cabs`.

This effectively disables the `collcabrate` optimization for those cabs.

PR #1158, opened 2016-10-15
[Cleanup] Minor cleanup in various places

ulteq:


only-a-ptr:
For some reason GitHub won't let me comment on commits anymore...

https://github.com/RigsOfRods/rigs-of-rods/pull/1158/commits/c787b9dffdc9246569412856e23bedfd9baf32a3
`DofManager` should be moved under `CameraManager`. It's the only class which ever accesses it. There's absolutely no reason why to have it in gEnv.
Besides, I'd like to remove gEnv completely. we have `App` namespace now. However, I'd like to minimize the amount of global objects there, too. Most of the time, they only need to be global because the code is badly designed and accesses them from places it shouldn't

source/main/physics/Beam.cpp
```diff
@@ -5932,32 +5923,28 @@ bool Beam::LoadTruck(
...
 					{
 						if (engine->getMaxRPM() > 3500)
 						{
 							//7000 rpm tachometer thanks to klink
 							dash->loadDashBoard("default_dashboard7000.layout", false);
 							// TODO: load texture dashboard by default as well
 							dash->loadDashBoard("default_dashboard7000.layout", true);
-						}
-						else
+						} else
```
only-a-ptr: This style of writing `else` is undesirable. I know it's everywhere and that I introduce inconsistencies, but it's just pain to look at for me. I'd like to eliminate it completely.

ulteq: The old style or the new style?

ulteq:
> DofManager should be moved under CameraManager.

Agreed.

source/main/physics/Beam.cpp
only-a-ptr: The old one, because it's mixed:
```
// okay
if (a)
{
    foo();
}
else
{
    bar();
}
// also okay (Java)
if (a) {
    foo();
} else {
    bar();
}
// the awful mix
if ()
{
    foo();
} else
{
    bar();
}
```

PR #1166, opened 2016-10-16
[Cleanup] Fixes warning 'C4715'

ulteq:
Also fixes some code formatting inconsistencies

### DIFF
--- a/source/main/utils/Settings.cpp
+++ b/source/main/utils/Settings.cpp
@@ -745,83 +745,110 @@ void Settings::LoadSettings(std::string filepath)
 	setSetting("User Token Hash", String(usertokensha1result));
 }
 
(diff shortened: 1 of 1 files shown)


only-a-ptr:
You left one inconsistency: missing newline between `int func()` and `{`
Anyway, was this really an inconsistency? All enum->str functions were written that way. I thought it was a clever domain-specific style.

I'm for using specific formatting where it's handy.

ulteq:
> Anyway, was this really an inconsistency?

The only reason I touched the code were the 'C4715 warnings.

> I thought it was a clever domain-specific style.

I don't think we should start introducing any more styles.

> You left one inconsistency: missing newline between int func() and {

Agreed, I will change it.

PR #1182, opened 2016-10-17
[Codechange] Reworked the sleeping mechanism

ulteq:
Trucks now stay active for at least 10 seconds before they go sleeping again.

Let sets of `SIMULATED` trucks sleep if none of them is "active" (player in them).


ulteq:
It can't be done like this.

PR #1187, opened 2016-10-18
[Codechange] Reworked the sleeping mechanism

ulteq:
* Trucks now stay active for at least 10 seconds before they go sleeping again.
* Trucks won't go sleeping if another `SIMULATED` truck is nearby.

But it also brings back parts of the 0.38.67 sleeping behavior:
* Sets of `SIMULATED` trucks can go sleeping if neither of them is active (player in them)

### DIFF
--- a/source/main/physics/BeamFactory.cpp
+++ b/source/main/physics/BeamFactory.cpp
@@ -481,83 +481,105 @@ Beam *BeamFactory::getBeam(int source_id, int stream_id)
 	return nullptr;
 }
 
-bool BeamFactory::truckIntersectionAABB(int a, int b)
+bool BeamFactory::intersectionAABB(Ogre::AxisAlignedBox a, Ogre::AxisAlignedBox b, float scale)
 {
-	return m_trucks[a]->boundingBox.intersects(m_trucks[b]->boundingBox);
+	if (scale != 1.0f)
+	{
+		Vector3 a_center = a.getCenter();
+		Vector3 a_half_size = a.getHalfSize();
+		a.setMaximum(a_center + a_half_size * scale);
+		a.setMinimum(a_center - a_half_size * scale);
+
+		Vector3 b_center = b.getCenter();
+		Vector3 b_half_size = b.getHalfSize();
+		b.setMaximum(b_center + b_half_size * scale);
+		b.setMinimum(b_center - b_half_size * scale);
+	}
+
+	return a.intersects(b);
 }
 
-bool BeamFactory::predictTruckIntersectionAABB(int a, int b)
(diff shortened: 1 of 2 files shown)


source/main/physics/BeamFactory.cpp
```diff
@@ -481,83 +481,105 @@ Beam *BeamFactory::getBeam(int source_id, int stream_id)
...
 {
 	if (!m_trucks[j] || m_trucks[j]->state != SIMULATED) return;
 
+	visited.set(j, true);
+
 	for (int t=0; t < m_free_truck; t++)
 	{
-		if (t == j || !m_trucks[t]) continue;
-		if (predictTruckIntersectionCollAABB(t, j))
+		if (t == j || !m_trucks[t] || visited[t]) continue;
```
only-a-ptr: Doesn't this completely defeat the point of the recursion? Unless I'm missing something, this means each truck will be evaluated exactly 1x no matter how many trucks may actually intersect with it, and that's a broken design.

ulteq:
`BeamFactory::UpdateSleepingState` runs once per frame. It creates a directed acyclic graph. The vertices equate to trucks and the edges to bounding box intersections.

The current truck (the one the player sits in) is the only root of the graph.

> Now that you touched this area...

It's not the first time I touched it. I've already simplified it greatly in this PR: https://github.com/RigsOfRods/rigs-of-rods/commit/a0a85c8d860dc675cbd1e91a0844325a77515400

> Our sleep mechanism doesn't cache AABB intersections.

It doesn't need to cache them, because we never check the same edge twice.

> Also, performance; Ogre's AABBs are slow

It's not worth to mess with them before we upgrade Ogre in my opinion.

source/main/physics/BeamFactory.cpp
ulteq: > Doesn't this completely defeat the point of the recursion?

It prevents creating a cyclic graph.

only-a-ptr:
> It's not the first time I touched it. I've already simplified it greatly in this PR: a0a85c8

That was a big win indeed. 

> a directed acyclic graph. ...The current truck is the only root of the graph.

I'm afraid that's not an universal solution. It works for the most stereotypical RoR scenario. But what if you build a tower from paper boxes and then shoot it with a tennis-ball cannon?

The graph is still fine, but you'll need one per each moving element. Falling boxes may wake up other boxes. Under these circumstances, both caching and AABB efficiency come back to play.

> It's not worth to mess with them before we upgrade Ogre in my opinion.

Agreed. And nobody implemented anything like the tennis-ball scenario so far. But it'll come (I hope!). It's something to keep in mind.

PR #1190, opened 2016-10-19
[Codechange] Allow snowball effect truck activation

ulteq:
This automatically activates all trucks which might soon get hit by any moving truck in the scene.

### DIFF
--- a/source/main/physics/BeamFactory.cpp
+++ b/source/main/physics/BeamFactory.cpp
@@ -608,11 +608,19 @@ void BeamFactory::UpdateSleepingState(float dt)
 	{
 		current_truck->state = SIMULATED;
 	}
+
+	std::bitset<MAX_TRUCKS> visited;
+	// Recursivly activate all trucks which can be reached from current_truck
 	if (current_truck && current_truck->state == SIMULATED)
 	{
 		current_truck->sleeptime = 0.0f;
-		std::bitset<MAX_TRUCKS> visited;
 		this->RecursiveActivation(m_current_truck, visited);
+	}
+	// Snowball effect (activate all trucks which might soon get hit by a moving truck)
+	for (int t=0; t<m_free_truck; t++)
+	{
+		if (m_trucks[t] && m_trucks[t]->state == SIMULATED && m_trucks[t]->sleeptime == 0.0f)
+			this->RecursiveActivation(t, visited);
 	}
 }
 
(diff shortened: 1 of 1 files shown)


ulteq:
> I'm afraid that's not an universal solution. It works for the most stereotypical RoR scenario. But what if you build a tower of paper boxes and then shoot it with a tennis-ball cannon?

@only-a-ptr, this should work now.

source/main/physics/BeamFactory.cpp
```diff
@@ -584,7 +584,7 @@ void BeamFactory::RecursiveActivation(int j, std::bitset<MAX_TRUCKS> &visited)
 	}
 }
 
-void BeamFactory::UpdateSleepingState(float dt)
+void BeamFactory::UpdateTruckStates(float dt)
```
only-a-ptr: I liked the old name. "Truck state" is vague.

source/main/physics/BeamFactory.cpp
```diff
@@ -608,12 +608,20 @@ void BeamFactory::UpdateSleepingState(float dt)
...
 	{
 		current_truck->sleeptime = 0.0f;
-		std::bitset<MAX_TRUCKS> visited;
 		this->RecursiveActivation(m_current_truck, visited);
 	}
+	// Snowball effect (activate all trucks which might soon get hit by a moving truck)
+	for (int t=0; t<m_free_truck; t++)
+	{
+		if (m_trucks[t] && m_trucks[t]->state == SIMULATED && m_trucks[t]->sleeptime == 0.0f)
+			this->RecursiveActivation(t, visited);
```
only-a-ptr: Unless I'm missing something, this call won't have any effect because all trucks were already visited.

ulteq: Only the trucks reachable from `current_truck` were already visited.

This new for-loop adds root nodes to the graph.

only-a-ptr: I've re-read the code and this is what I got:
- the `RecursiveActivation(truck)` marks the `truck` immediately as visited.

Hence:
 1. `RecursiveActivation(curr_truck)` marks curr_truck visited.
 2. The loop inside will traverse ALL other trucks: https://github.com/ulteq/rigs-of-rods/blob/f45980ccabb3a921be1f283024a7f52dfd72babc/source/main/physics/BeamFactory.cpp#L570 - hence marking them visited

ulteq: The `this->RecursiveActivation(t, visited);` calls inside of the `RecursiveActivation` function will never be called if `t` was visited before.

> From where, then?

From the snowball effect for-loop.

ulteq: > The snowball pass will do nothing.

I just noticed that the snowball pass does more than needed.

`this->RecursiveActivation(t, visited);` does not need to be called if `visited[t] == true`.

Issue #1163, opened 2016-10-15
RoR.cfg saving issues


Michael10055:
Some options do not get saved properly on RoR exit:

- ~~Replay Mode~~ (#1160)
- ~~Disable Overview Map~~ (#1161)
- ~~"Preselected Map"~~ (1191)
- ~~"FOV External" value is overwritten by the current ingame value.~~ (#1192)
- ~~Vegetation resets to "None" after loading a terrain with vegetation.~~ (#1164)

only-a-ptr:
@Michael10055 Please look at this table: https://github.com/RigsOfRods/rigs-of-rods/wiki/Console-variables and tell me which GVars should be saved on exit and which not.

Michael10055:
Everything looks fine to me besides the preselected map and the FOV issue.

disloyalpick:
The FOV issue is also present on Linux. But can you tell me how you would reproduce the preselected map issue, please?

Michael10055:
In RoRConfig, go to the `Debug` menu, and type a terrain name in the "Preselected Map:" textbox (ex. simple2) and start RoR. It should automatically load the preselected map. Now close RoR and check the textbox in RoRConfig again, the value will be empty.

only-a-ptr:
I assume the "Preselected Truck" settings should also persist between sessions, like "Preselected Map"

Currently, all these settings would get reset:
 * "Preselected Truck";
 * "Preselected TruckConfig";
 * "Enter PreselectedTruck";

Michael10055:
Yes, both the Preselected Truck/Map options should be persistent.

Michael10055:
👍

PR #1191, opened 2016-10-19
[Bugfix] Fixes 'Preselected map' setting

ulteq:
Tested with `SkipMainMenu=Yes/No` in both singleplayer and multiplayer.

Related with: #1163


only-a-ptr:
I wouldn't think to put it in "diag" myself, but I suppose it's mostly useful for modding, so it pretty much belongs there

:shipit:

PR #1192, opened 2016-10-19
[Bugfix] Fixes 'FOV' setting

ulteq:
Makes 'FOV External' and 'FOV Internal' plain cvars.

Related with: #1163

### DIFF
--- a/source/main/utils/Settings.cpp
+++ b/source/main/utils/Settings.cpp
@@ -540,13 +540,25 @@ void App__SetGfxEnvmapRate(std::string const & s)
     App::SetGfxEnvmapRate(rate);
 }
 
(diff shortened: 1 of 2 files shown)


only-a-ptr:
I'm already working on a generic solution for this: https://github.com/only-a-ptr/rigs-of-rods/commits/gvars-3s

ulteq:
I like your generic solution, but i still think this should be merged first.

> I'm already working on a generic solution for this

Will it be ready for 0.4.7.0?

> It turned out there are more gvars which need to retain config setting

The `preselected` settings are supposed to be read-only.

only-a-ptr:
> I like your generic solution, but i still think this should be merged first.

Agreed :shipit:

Issue #1100, opened 2016-08-14
Steering wheel not rotating when animators are used


Aivar-RoR:
When a truck has the default dashboard (dashboard.mesh or dashboard-small.mesh), the steering wheel does not rotate.
Using latest Nightbuild (120).

Tried with the Agora bus from this thread:
http://forum.rigsofrods.org/thread-424.html
Also attached to the next post.

Aivar-RoR:
This issue seems to be caused by animators. The 0.4 version of the bus causes the issue.
As soon as the animators line gets commented out, the steering wheel rotates correctly.

animators
;extends with actual vehicle speed
31, 265, 0.8, airspeed | inv

I'm attaching the buses .zip from the aforementioned thread over here, because I'm scrapping the version with animators until this issue gets fixed.
[Aivar-agora-d3-ver2.zip](https://github.com/RigsOfRods/rigs-of-rods/files/428428/Aivar-agora-d3-ver2.zip)

ulteq:
Most likely caused by this: https://github.com/RigsOfRods/rigs-of-rods/pull/904

ulteq:
What is the animator line supposed to do?
```
31, 265, 0.8, airspeed | inv
```

Aivar-RoR:
Once the vehicle is in motion, the animator extends with the vehicle's actual speed (i.e. airspeed). The animator then extends a trigger (defined between the same nodes as the animator) that blocks the bus doors from being opened. 
There were no problems with the steering wheel in 0.39.5 and 0.4.0.7 (the versions I have installed atm) with this animator line.

PR #1199, opened 2016-10-20
[Bugfix] Fixes conflict between animators and hydros

ulteq:
The steering wheel rotation angle (`hydrodirwheeldisplay`) must not be affected by animators.

Fixes: #1100

### DIFF
--- a/source/main/physics/BeamForcesEuler.cpp
+++ b/source/main/physics/BeamForcesEuler.cpp
@@ -814,7 +814,7 @@ void Beam::calcForcesEulerCompute(int doUpdate, Real dt, int step, int maxsteps)
 			if (hydroInertia)
 				cstate=hydroInertia->calcCmdKeyDelay(cstate,i,dt);
 
-			if (!(beams[hydro[i]].hydroFlags & HYDRO_FLAG_SPEED))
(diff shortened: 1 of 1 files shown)


only-a-ptr:
:+1: You did it!

ulteq:
Yeah, but what I don't understand:

> [There were no problems with the steering wheel in 0.39.5 and 0.4.0.7 (the versions I have installed atm) with this animator line.](https://github.com/RigsOfRods/rigs-of-rods/issues/1100#issuecomment-247409159)

This line of code is very old: https://github.com/only-a-ptr/ror-legacy-svn-trunk/blob/master/source/main/physics/BeamForcesEuler.cpp#L1708-L1709

The issue must have been hidden by a different control flow in previous versions.

I suspected this commit: https://github.com/RigsOfRods/rigs-of-rods/commit/ef3b5293e8390715c13d707411101b4df0222584, but reverting it did not fix the bug.

only-a-ptr:
> Yeah, but what I don't understand

Strange, but RoR is still mysterious. Might as well be multithreading - traction/handling stuff is done half mainthread half simthreads. I don't recall exactly where are the collision surfaces, but I do remember how my scripted engine sim blew everything up.

PR #1204, opened 2016-10-21
MP serverlist GUI now fully functional

### LINKED ISSUE #1193: In-game serverlist not loading
The in-game server list isn't woking yet, I would like to see it finished before releasing 0.4.7.
The server list is already able to return the list in JSON: http://multiplayer.rigsofrods.org/server-list?json=true

only-a-ptr:
ror-serverlist-yay

It all works!

Fixes #1193


ulteq:
Do we really need to add almost 10.000 lines for this?

> Only manual refreshing works.

I think that's just fine. I don't see the purpose of auto refreshing.

only-a-ptr:
> Do we really need to add almost 10.000 lines for this?

Most of that is JsonCpp library (to be moved out to deps) and the cloned/patched MyGUI::MultiListBox

Initially I thought I'll handle HTTP with rorserver code+SocketW, so I incorporated JsonCpp for simplicity. But now that I had to use C++Requests library externally, update of our deps is needed anyway, so we can move JsonCpp out right away.

> I think that's just fine. I don't see the purpose of auto refreshing

We don't need periodic auto-refresh, but IMO it should update once when opened.

ulteq:
> I meant it doesn't auto-update when opened.

Ok, that should be changed.

tritonas00:
Great, but this PR is extremely linux unfriendly to compile

https://paste.ee/p/yEQ7s

only-a-ptr:
> Great, but this PR is extremely linux unfriendly to compile

It appears to be some hassle between Linux and libcurl. They somehow fight about OS headers.

source/main/utils/Utils.cpp
```diff
@@ -28,7 +28,9 @@
 #include <Ogre.h>
 
 #ifndef _WIN32
-#include <iconv.h>
+#   include <iconv.h>
+#else
+#   include <windows.h> // Sleep()
```
ulteq: Typo?

source/main/gui/panels/GUI_MultiplayerSelector.cpp
```diff
@@ -35,94 +35,246 @@
 #include "GUIManager.h"
 #include "Application.h"
 #include "MainThread.h"
+#include "Http.h"
```
ulteq: Do we still need this?

source/main/main.cpp
```diff
@@ -45,6 +45,10 @@ using namespace Ogre;
 #include "ShellAPI.h"
 #endif //OGRE_PLATFORM_WIN32
 
+#ifdef USE_CURL
+#   include <curl/curl.h>
+#endif //USE_CURL
```
ulteq: Do we need `include <stdio.h>` here?

source/main/CMakeLists.txt
```diff
@@ -410,6 +425,7 @@ target_include_directories( ${BINNAME} PRIVATE
  threadpool
  utils
  utils/profiler
+ utils/json
```
ulteq: This line seems to break the build process on linux.

source/main/gui/panels/GUI_MultiplayerSelector.cpp
```diff
@@ -35,94 +35,246 @@
...
 
+#include "json.h"
```
ulteq: Only `#include "json/json.h"` seems to work on linux.

source/main/main.cpp
only-a-ptr: I don't know. But older includes of curl.h were always coupled with stdio.h

only-a-ptr:
@ulteq I squashed your build fix and CMake alphabetic cleanup

tritonas00:
Now compiles and works fine  :)

Issue #1058, opened 2016-06-20
VY/VX node not found flexbody error -> crash


Michael10055:
If you have a flexbody line with invalid forsets, the game will just crash with this line in the log:
```
12:29:55: FLEXBODY ERROR on mesh BK_73Suburban_RBump.mesh: VX node not found
```
In a older build I have (68f5b34), the error is ignored, Unfortunately I don't know of any released mods that have this error.

ulteq:
Needs to be retested one this one is merged: https://github.com/RigsOfRods/rigs-of-rods/pull/1137

Michael10055:
Still crashes on the latest upstream.

AnotherFoxGuy:
Id: 90    | Current: Yes    | StackSize: 20  
---|---|---  
Function | File(Line)  
**OgreMain** | ** \- **  
RoR!FlexBody::FlexBody | f:/rigs-of-rods/source/main/physics/flex/flexbody.cpp(560)  
RoR!RoR::FlexFactory::CreateFlexBody | f:/rigs-of-rods/source/main/physics/flex/flexfactory.cpp(88)  
RoR!RigSpawner::ProcessFlexbody | f:/rigs-of-rods/source/main/physics/rigspawner.cpp(1868)  
RoR!RigSpawner::SpawnRig| f:/rigs-of-rods/source/main/physics/rigspawner_processcontrol.cpp(16707566)  
RoR!Beam::LoadTruck | f:/rigs-of-rods/source/main/physics/beam.cpp(5756)  
RoR!Beam::Beam | f:/rigs-of-rods/source/main/physics/beam.cpp(5468)  
RoR!BeamFactory::CreateLocalRigInstance | f:/rigs-of-rods/source/main/physics/beamfactory.cpp(215)  
RoR!RoRFrameListener::updateEvents | f:/rigs-of-rods/source/main/gameplay/rorframelistener.cpp(1308)  
RoR!RoRFrameListener::frameStarted | f:/rigs-of-rods/source/main/gameplay/rorframelistener.cpp(1571)  
OgreMain |  \-   
OgreMain |  \-   
RoR!RoR::MainThread::EnterGameplayLoop | f:/rigs-of-rods/source/main/mainthread.cpp(862)  
RoR!RoR::MainThread::Go | f:/rigs-of-rods/source/main/mainthread.cpp(439)  
RoR!main | f:/rigs-of-rods/source/main/main.cpp(250)  
[...]

PR #1205, opened 2016-10-22
[Bugfix] Fixes game crash on invalid forset line

ulteq:
This only hides the underlying issue.

Fixes: #1058


AnotherFoxGuy:
👍 This fixes the crash on windows

only-a-ptr:
> This only hides the underlying issue.

If that's so, pls comment it in the code. I guess this avoids the "Flexbody: node ABCD whatever not found" issue, but I don't know exactly

ulteq:
> I guess this avoids the "Flexbody: node ABCD whatever not found"

It only avoids a crash to desktop

PR #1211, opened 2016-10-22
[Bugfix] Fixed a typo

ulteq:
Typo was introduced by this commit: fa85f6a412c6c8999b59c0a6c2a15cb7c252b0f8


disloyalpick:
awsome. rollbacks can finally be properly used again. :+1: 

Now how about figuring out why the submeshes are so slippery and implementing some sort of a fix to make loading vehicles onto trucks and/or trailers

ulteq:
I have already tried to figure it out, but without success so far.

PR #1218, opened 2016-10-23
[Bugfix] Dashboard frozen after map switch

### LINKED ISSUE #1148: Back to menu makes counters not working
Load any map and a car with counters in dashboard (gavril omega).
Press Back to menu, reload the map and car. [Counters](https://cloud.githubusercontent.com/assets/2660424/19351116/1e593ab2-9163-11e6-8c6e-11176a0d5257.png) no longer work.

Same happens if you dont load a car, press Back to menu and then load it.

only-a-ptr:
Problem: duplicate RTT texture "dashtexture". It's strange that OGRE didn't complain when re-creating texture with the same name.

Fixes #1148


tritonas00:
Works!

PR #1221, opened 2016-10-24
Moved JSON CPP to dependencies (Windows only)

AnotherFoxGuy:


only-a-ptr:
Thanks :heart: 

Where can I get updated build script for the deps now? I used SpeciesX's builds before.

Issue #1220, opened 2016-10-24
Back to menu stretches airplanes HUD


tritonas00:
As the title says: https://cloud.githubusercontent.com/assets/2660424/19629154/ef456c3e-9976-11e6-92bb-ff9c30850694.png

PR #1223, opened 2016-10-24
[Bugfix] Fixes streched airplane HUD

ulteq:
Fixes: #1220

It's ugly, but it works.


source/main/gui/OverlayWrapper.cpp
```diff
@@ -62,6 +62,8 @@
 
 using namespace Ogre;
 
+static bool m_is_scaled = false; // ugly
```
only-a-ptr: `m_` is reserved for class member variables. I use `g_` for module-scope variables.

However, why a static var when you're in a class? And you even treat the variable like member. And it's prefixed `m_`. That's uber-confusing, which isn't a problem, but there's no reason for it, and that's a problem.

Otherwise, the solution is completely fine.

PR #1224, opened 2016-10-24
[Bugfix] Crash when dragging map selector GUI

### LINKED ISSUE #917: Crash when dragging the terrain selector menu
Steps to reproduce:
- Go to single player
- Try to move the terrain loader window

The game crashes with an Error 6: FileSystemArchive::open

only-a-ptr:
Fixes #917

### DIFF
--- a/source/main/gui/panels/GUI_MainSelector.cpp
+++ b/source/main/gui/panels/GUI_MainSelector.cpp
@@ -124,7 +124,8 @@ void CLASS::BindKeys(bool bind)
 
 void CLASS::NotifyWindowChangeCoord(MyGUI::Window* _sender)
 {
-	ResizePreviewImage();
+	if (m_Preview->isVisible())
+		this->ResizePreviewImage();
 }
 
 void CLASS::EventKeyButtonPressed_Main(MyGUI::WidgetPtr _sender, MyGUI::KeyCode _key, MyGUI::Char _char)
@@ -880,6 +881,7 @@ void CLASS::SetPreviewImage(Ogre::String texture)
 	}
 	catch (...)
 	{
+		Ogre::LogManager::getSingleton().stream() << "[RoR|SelectorGUI] Failed to load preview image: " << texture;
 		m_Preview->setVisible(false);
 	}
 }
@@ -887,7 +889,17 @@ void CLASS::SetPreviewImage(Ogre::String texture)
 void CLASS::ResizePreviewImage()
 {
 	MyGUI::IntSize imgSize(0, 0);
-	Ogre::TexturePtr t = Ogre::TextureManager::getSingleton().load(m_preview_image_texture, Ogre::ResourceGroupManager::AUTODETECT_RESOURCE_GROUP_NAME);
+	Ogre::TexturePtr t;
+	try
+	{
+		t = Ogre::TextureManager::getSingleton().load(m_preview_image_texture, Ogre::ResourceGroupManager::AUTODETECT_RESOURCE_GROUP_NAME);
+	}
+	catch (...)
+	{
+		Ogre::LogManager::getSingleton().stream() << "[RoR|SelectorGUI] Failed to load preview image: " << m_preview_image_texture;
+		m_preview_image_texture.clear();
+		m_Preview->setVisible(false);
+	}
 	if (!t.isNull())
 	{
 		imgSize.width = (int)t->getWidth() * 10;


ulteq:
I think we should either add `if (m_preview_image_texture.empty()) return;` inside of `ResizePreviewImage()` or surround the call to `Ogre::TextureManager::getSingleton().load(...)` with a try ... catch block.

only-a-ptr:
Both suggestions are substantially worse IMO.

ad A: When you know you have no money, do you still go to the shop and then back home? This latest-possible-moment-bailout style is one of code plagues of RoR. I'm removing it where I can.

ad B: That's outright wrong, except if you have an OOP fetish. Exceptions are a very slow mechanism designed for handling unexpected occurences. When you already KNOW you're gonna hit the wall, why hit the wall?

ulteq:
> When you already KNOW you're gonna hit the wall, why hit the wall?

`m_preview_image_texture.empty()` won't detect typos in the texture name.

only-a-ptr:
> m_preview_image_texture.empty() won't detect typos in the texture name.

In general, you're right. Loading resources can always yield unexpected results. I'll add a try/catch in there.

However, typo can't happen here because the image name is auto-detected from ZIPfile.

ulteq:
> Both suggestions are substantially worse IMO.

New suggestion:

```
void CLASS::NotifyWindowChangeCoord(MyGUI::Window* _sender)
{
    if (m_Preview->getVisible())
        ResizePreviewImage();
}
```

only-a-ptr:
> New suggestion:

That's the same thing, except using less obvious condition. But why not...

ulteq:
All I want is to disable `SetPreviewImage` when `SetPreviewImage` fails.

source/main/gui/panels/GUI_MainSelector.cpp
```diff
@@ -887,7 +888,17 @@ void CLASS::SetPreviewImage(Ogre::String texture)
...
+	try
+	{
+		t = Ogre::TextureManager::getSingleton().load(m_preview_image_texture, Ogre::ResourceGroupManager::AUTODETECT_RESOURCE_GROUP_NAME);
+	}
+	catch (...)
+	{
+		Ogre::LogManager::getSingleton().stream() << "[RoR|SelectorGUI] Failed to load preview image: " << m_preview_image_texture;
+		m_preview_image_texture.clear();
+		m_Preview->setVisible(false);
+	}
```
ulteq: Do we still need all this? I thought the `if (m_Preview->isVisible())` check would be enough.

only-a-ptr: Exception shouldn't pass un-noticed. The player (mod author?) should know the image is corrupted.

`m_Preview->setVisible(false);` is appropriate - without it, the exception will be thrown on every UpdateCoord event. No point in that.

`m_preview_image_texture.clear();` is not technically necessary, but feels like an appropriate cleanup.

ulteq: If we catch the exception inside of `ResizePreviewImage()`, won't we then run into issues here:
https://github.com/RigsOfRods/rigs-of-rods/blob/a484facaddbcafe0fa71427c33dd08cb5a731c10/source/main/gui/panels/GUI_MainSelector.cpp#L877-L879

ulteq: I would throw this `Ogre::LogManager::getSingleton().stream() << "[RoR|SelectorGUI] Failed to load preview image: " << m_preview_image_texture;` into the catch block inside of `SetPreviewImage(Ogre::String texture)` and call it a day.

only-a-ptr: yeah.

source/main/gui/panels/GUI_MainSelector.cpp
```diff
@@ -880,14 +881,25 @@ void CLASS::SetPreviewImage(Ogre::String texture)
...
 void CLASS::ResizePreviewImage()
 {
 	MyGUI::IntSize imgSize(0, 0);
-	Ogre::TexturePtr t = Ogre::TextureManager::getSingleton().load(m_preview_image_texture, Ogre::ResourceGroupManager::AUTODETECT_RESOURCE_GROUP_NAME);
+	Ogre::TexturePtr t;
+	try
+	{
+		t = Ogre::TextureManager::getSingleton().load(m_preview_image_texture, Ogre::ResourceGroupManager::AUTODETECT_RESOURCE_GROUP_NAME);
+	}
+	catch (...)
```
ulteq: This won't work. This code: https://github.com/only-a-ptr/rigs-of-rods/blob/99d734803b5448f492b164ce0ca6ad0308a076b8/source/main/gui/panels/GUI_MainSelector.cpp#L879-L880
would be executed after catching everything here.

ulteq: Proposal: https://github.com/ulteq/rigs-of-rods/commit/08f1662586b12c6f6399c31c36aef7ab0daa196a

only-a-ptr:
@ulteq I've reset my `master` branch to upstream, cherrypicked your fix (:+1:) and then committed directly to upstream by accident (who set my default remote to upstream :open_mouth: ?)

PR #1227, opened 2016-10-25
[Codechange] Added kickdown delay

ulteq:
Forbid kickdowns right after forced upshifts


only-a-ptr:
Someone's suggestion or your own driving experience?

ulteq:
My own driving experience.

PR #1233, opened 2016-10-31
[Fix] 45000% faster truckfile parsing

only-a-ptr:
Regexes aren't evil or dangerous, except if the programmer is an idiot :blush: 

Attached is a benchmark with a README introducing a new perf. guideline.


ulteq:
`Cache regen` with ~370 mods (both terrains and trucks):
Before: 47 seconds
After: 19 seconds

Speciesx:
72 terrains + 447 trucks.

Before: 130 seconds
After: 28 seconds

tritonas00:
322 mods

Before: 27 seconds
After: 15 seconds

Yes, everything spawns a lot faster, almost immediately :) :+1:

disloyalpick:
Very much faster on my linux install. Going back to Windows to test. Haven't used linux lately due to RoR not using my Nvidia GPU. And having no idea how to make it do that. Lol.

PR #1243, opened 2016-11-15
Disabled "Allow DirectX9Ex" setting

### LINKED ISSUE #1239: "Allow DirectX9Ex" option causing terrain loading crashes
See this post for reference: https://forum.rigsofrods.org/thread-697-post-6546.html#pid6546

What is the "Allow DirectX9Ex" option supposed to do? I always have it set to `No` - I figured it was a 'added leftover' from the OGRE 1.9 port.

AnotherFoxGuy:
Fixes #1239

### DIFF
--- a/source/configurator/Configurator.cpp
+++ b/source/configurator/Configurator.cpp
@@ -1773,6 +1773,7 @@ void MyDialog::updateRendersystems(Ogre::RenderSystem *rs)
 	filterOptions["VSync Interval"]=true;
 	filterOptions["sRGB Gamma Conversion"]=true;
 	filterOptions["Colour Depth"]=true;
+	filterOptions["Allow DirectX9Ex"]=false;
 
 	if(renderer->GetCount() == 0)
 	{


only-a-ptr:
Good. Could you also remove the checkbox from RoRConfig altogether? Since we'll port to OGRE 2.1 in foreseeable future, we won't ever need it again. Removing it is just a matter of finding the widget setup code (the GUI is fully created in code) and removing it.

AnotherFoxGuy:
Which checkbox do you mean?

only-a-ptr:
@AnotherFoxGuy Ignore that, I was confused. I thought there was a checkbox for this, but it's actually one of the optional droplists under "Render System" tab: https://forum.rigsofrods.org/thread-697-post-6546.html#pid6546

Does the "Allow Dx9Ex" option appear for you? If yes, can you block it from appearing in the options? The loop which generates the option droplists is here: https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/configurator/Configurator.cpp#L1814-L1980 You can tell the option by it's name: https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/configurator/Configurator.cpp#L1869

PR #1265, opened 2017-01-18
New beacon meshes

Michael10055:
Thanks to @negativeice.
Old -> New:
ogremeshtool_2017-01-18_16-28-17
ogremeshtool_2017-01-18_16-28-47


only-a-ptr:
What tool did you use to capture those screenshots?
I've been struggling to find a working editor for viewing OGRE meshes and didn't find much. There's a quite recent build of OgreMeshy by dark_sylinc somewhere, but most places have outdated downloads.

I like the new meshes, but they're very detailed, they need LODs. The detail on the screenshots is only meaningful for close view. IIRC the OgreMeshConverter doesn't generate LODs automatically but there's an option for it. There are also separate tools for it AFAIK. Finally, it can be done manually by hand-creating separate low-poly meshes and supplying them to OgreMeshConverter.

Also, it looks like there's an split edge (Blender's edge-split modifier - "mark sharp" function) between the cylindric part of the beacon and the cap. 
ror-beacon-edgesplit
it shouldn't be the case. The right way to deal with the 
auto-smoothing in this case is to make a ring of very narrow quads.

Michael10055:
I use the OgreMeshTool (found here: https://code.google.com/archive/p/ogre-meshtool/downloads) Last updated in 2011. It supports up to OGRE 1.8.x meshes AFAIK. 

I'll contact @negativeice about the meshes themselves. Thanks!

PR #1266, opened 2017-01-20
MyGUI layout cleanup

Michael10055:
Removed all useless `IncludeDirectory` and `PanelNamespace` lines from the .layout files.

### DIFF
--- a/bin/resources/mygui/DebugOptions.layout
+++ b/bin/resources/mygui/DebugOptions.layout
@@ -56,9 +56,6 @@
         </Widget>
     </Widget>
     <CodeGeneratorSettings>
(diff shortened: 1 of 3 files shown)


only-a-ptr:
Thanks for the effort, but it's not worth doing, really.

The whole `<CodeGeneratorSettings>` is inserted by MyGUI's LayoutEditor and has no meaning otherwise. To keep it clean as you made it, one would have to manually clean it every time.

Also, I intend to drop MyGUI in favor of [Dear IMGUI](https://github.com/ocornut/imgui) because MyGUI gets clumsy very fast if you want some more complex UI (RigEditor for example) - you spend hours in LayoutEditor hand-aligning and click-configuring everything.

PR #1289, opened 2017-03-02
Fixed compiler warnings and undefined operation

susnux:
When building for openSUSE I discovered some warnings (one of them is considered an error by RPMLint).

**First kind of warning:**
> I: Program causes undefined operation
>    (likely same variable used twice and post/pre incremented in the same expression).
>    e.g. x = x++; Split it in two operations.
> W: rigsofrods sequence-point

Triggered by:

1. `TruckHUD.cpp`: It is not defined which value `overlayElement` should have after the `=`.
(Because it is not defined which `=` should be done first).
2. `BeamForcesEuler.cpp`, same kind of issue: It is not defined which should be executed first so `contacted` might be set by `(contacted = ... )` before the  code at the right side of the `|` operator is executed, or it might be executed after that part so the right side can result in two different solutions.

**Second kind of warning (considered an error by RPMLint):**
> I: Expression compares a char* pointer with a string literal.
>    Usually a strcmp() was intended by the programmer
> E: rigsofrods stringcompare

Triggered by `GUI_DebugOptions.cpp`: I think this is clear, here are two pointers be compared which will in this case be always false, I think it was meant to compare the content of the strings so simply remove the c_str() call.

**Third kind of warning:**
> I: Program is likely to break with new gcc. Try -fno-strict-aliasing.
> W: rigsofrods strict-aliasing-punning

Triggered by `BeamFactory.cpp`: [If strict aliasing punning is not clear ](https://blog.qt.io/blog/2011/06/10/type-punning-and-strict-aliasing/). Not sure if this is the best fix, but I think the code readability should not be affected by this fix.

### DIFF
--- a/source/main/gui/TruckHUD.cpp
+++ b/source/main/gui/TruckHUD.cpp
@@ -408,7 +408,7 @@ bool TruckHUD::update(float dt, Beam* truck, bool visible)
         }
 
         // hide command section title if no commands
-        overlayElement = overlayElement = OverlayManager::getSingleton().getOverlayElement("tracks/TruckInfoBox/CommandsTitleLabel");
+        overlayElement = OverlayManager::getSingleton().getOverlayElement("tracks/TruckInfoBox/CommandsTitleLabel");
         overlayElement->setCaption("");
         if (filledCommands > 0)
         {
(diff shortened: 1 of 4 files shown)


only-a-ptr:
:thumbs_up: Many thanks for the contribution an detailed explanation. I'll merge later today.

PR #1292, opened 2017-03-04
Buildsystem allow reproducible builds

susnux:
Use [SOURCE_DATE_EPOCH](https://reproducible-builds.org/specs/source-date-epoch/) environment variable if available to force build date and time.

This allows packagers to override `__DATE__` and `__TIME__` to make the build deterministic (build of same source will result in same binary), for more information please have look at:
https://reproducible-builds.org/


only-a-ptr:
Thanks for the contribution and interesting links.

PR #1298, opened 2017-03-14
Fixed crash on cache-regen if cache directory deleted

### LINKED ISSUE #1285: Crash on cache-regen if cache directory deleted.
If the "MyDocuments\Rigs of Rods 0.4\cache" directory is deleted, RoR attempts to regen cache and crashes with message "cannot write to {MyDocuments\Rigs of Rods 0.4\cache\mods.cache}"

ror-cachedel-crash

This shouldn't happen. The directory should be auto-created instead.

AnotherFoxGuy:
Fixes #1285


only-a-ptr:
Good work, thanks 👍

PR #1302, opened 2017-03-21
Partial workaround for the "no sound" issues.

only-a-ptr:
When the audio device specified in RoR.cfg fails to open for any reason, RoR tries to open system's default audio device. Additionally, the faulty device name is erased from RoR.cfg.

This partially works around the issue with special characters in sound device name.


only-a-ptr:
@AnotherFoxGuy `LOG()` takes single `std::string/Ogre::String`, `LOGSTTEAM` lets me use the `<<` operator and auto-prefixes the log line with `[RoR|Whatever] `, which is handy to recognize OGRE/RoR messages in log.

The way the macro works is not elegant, but super convenient.

Issue #267, opened 2015-06-20
Speed up Travis build


Hiradur:
- [x] Eliminate use of sudo to get access to container based instead of VM-based infrastructure (-> faster boot up)
- [x]  `make install` missing dependencies to a location in home folder 
- [x]   Check if all packages are white-listed: https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-precise 
- [x]  Create white-list request for missing packages: https://github.com/travis-ci/travis-ci/issues/new?title=APT+whitelist+request+for+___PACKAGE___ 
- [x]  Replace `sudo apt-get install` with this: http://docs.travis-ci.com/user/apt/
- [x]  waiting for white-list approval (still without approval: https://github.com/travis-ci/apt-package-whitelist/issues/378)
- [x] Precompile dependencies not available as packages and make Travis download them from a server, should cut build time in half. Use own server or Travis' cache macro: http://docs.travis-ci.com/user/caching/
Preview here: https://github.com/Hiradur/rigs-of-rods/tree/Travisexp
- [x] Look into upgrading to Trusty

PR #1303, opened 2017-03-24
Updated travis to trusty

AnotherFoxGuy:
See #267


Meheh:
has that crashfix problem been fixed yet?

PR #1310, opened 2017-04-05
(WIP) Terrn2 deployment

only-a-ptr:
Terrn2 terrain consists of many specific fileformats: TERRN2, OTC, TOBJ, ODEF, Landuse, Groundmodel and more... At the moment, they're all loaded and processed every time a terrain spawns.

Deployment means the loading and parsing is only done once, on first terrain load. When it's done, a single JSON file (JTERRN) with all the data (pre-transformed for straigtforward spawn) is saved into cache. All subsequent terrain loads are done from this cached JTERRN.

Benefit for player: Fast terrain load. Excepting the "Initializing overview map subsytem" lag, the loads will be practically immediate.

Benefit for developer: a lot cleaner codebase. Instead of the loading/parsing/processing/spawning all happening at once (or in mixed order), everything is parsed first, then processed, then cached, then spawned.

At the moment, the code works in test mode - it generates the JTERRN and spawns from it, but it doesn't cache it.


only-a-ptr:
I'm pulling this PR back because the "deployment" functionality turned out to be more trouble than it's worth and the parsers alone are the signifficant improvement.

I'll create a revised PR soon.

PR #1323, opened 2017-04-23
Fixed water height not loading

### LINKED ISSUE #1318: Water stuck at 0 height with Hydrax
On the latest master, all water is stuck at 0 height instead of using the value specified in the `.terrn2` file.
screenshot_2017-04-16_19-16-33_1
To reproduce, just load a map that has water, such as North St Helens.
Changing the water type doesn't change anything. The `/setwaterlevel` console command still works though.

AnotherFoxGuy:
Fixes #1318


tritonas00:
works!

only-a-ptr:
👍

PR #1324, opened 2017-04-23
Disabled "Fixed Pipeline Enabled" setting

### LINKED ISSUE #1297: Fixed Pipeline=Enabled Causes RoR to crash after launching RoR
Title says it all

Steps to reproduce...

1. Go to Render Subsystem settings under Direct3D9 and Switch Fixed Render Pipeline to "NO"

2. Launch RoR.

3. Crash RoR. 

Not sure what its purpose is, but I just thought I would report.

m6pv0u8

AnotherFoxGuy:
Fixes #1297


AnotherFoxGuy:
@only-a-ptr but if you want to do that I think we also should change it for the [other settings](https://github.com/AnotherFoxGuy/rigs-of-rods/blob/2efc751eec0a68d0e6eded65769a97740cd3984a/source/configurator/Configurator.cpp#L1379-L1386)

only-a-ptr:
@AnotherFoxGuy Good point, but let's only hardcode the options which are known to cause problems:
* "Allow DirectX9Ex" - IIRC somebody tried setting this "on" and got crashes. I wasn't able to google much info on DirectX9Ex, but it seems like some kind of desktop acceleration extension: https://en.wikipedia.org/wiki/Direct3D#Direct3D_9 - seems safe to disable.
* "Fixed Pipeline Enabled"

PR #1325, opened 2017-04-23
Improved RoRConfig log messages

AnotherFoxGuy:
Removes this dialogue box when starting RoRConfig 

rorconfig_2017-04-01_10-18-30


only-a-ptr:
I had no idea `wxLogInfo()` pops up such a log box. I don't think I ever clicked the "save and execute" button.

Anyway, please stick to the coding style in place.
```
if (foo)
{
    bar();
}
```
All the code except yours uses this style.

AnotherFoxGuy:
> @only-a-ptr  However, please stick to the coding style in place.
All the code except yours uses this style.

Written to much javascript this week :smile:
Should be fixed now

PR #1327, opened 2017-05-11
Project CamCon: Camera controller refactor

only-a-ptr:
I couldn't resist :blush: Read individual commit messages for justification of what I'm doing.

Please test camera switching and functioning - nothing should change, but bugs happen.


tritonas00:
I can't notice any difference

only-a-ptr:
More changes to test.

tritonas00:
will this also resolve this https://github.com/RigsOfRods/rigs-of-rods/issues/1109 as well or it's irrelevant?

only-a-ptr:
@tritonas00 Not directly, but it may help me fixing it. From what I observed, the stuttering is actually not coming from the camera, but from the vehicle/character because the scenery around is moving smoothly.

tritonas00:
> More changes to test.

Cam switching broken and causes segfaults

only-a-ptr:
@tritonas00 Which is the last working  commit you tested?

tritonas00:
6ee5831 was ok

PR #1339, opened 2017-06-17
Renamed setDirectionArrow => UpdateDirectionArrow

### LINKED ISSUE #1334: Angelscript races broken
Load up any terrain that has races such as [North St Helens](https://archives.rigsofrods.org/repo/0.4-Terrains-klink/nhelens.zip). You'll notice that no checkpoints are to be seen on the map. This is the error in the Angelscript.log:
```
18:23:20: races.as (1016, 2): Info = Compiling void racesManager::setupArrow(int)
18:23:20: races.as (1046, 9): Error = No matching signatures to 'GameScriptClass::setDirectionArrow(string, vector3)'
18:23:20: races.as (1051, 9): Error = No matching signatures to 'GameScriptClass::setDirectionArrow(string, vector3)'
18:23:20: Failed to build the module
```
In older RoR versions, a yellow arrow would show up giving you a hint on where the next checkpoint is located. This arrow isn't present in 0.4.7+ and would be nice if it returned.
0.4.0.7:
screenshot_7
0.4.7.0:
screenshot_2017-06-09_18-50-11_1

AnotherFoxGuy:
Fixes #1334

### DIFF
--- a/bin/resources/scripts/races.as
+++ b/bin/resources/scripts/races.as
@@ -1043,12 +1043,12 @@ class racesManager {
 		}
 		
 		if( this.raceList[this.currentRace].laps == this.LAPS_NoLaps )
-			game.setDirectionArrow(this.raceList[this.currentRace].raceName+"\ncheckpoint "+position+" / "+(this.raceList[this.currentRace].checkPointsCount-1), vector3(v[0], v[1], v[2]));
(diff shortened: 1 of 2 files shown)


only-a-ptr:
Oops! My bad, I broke it. Thanks for fixing. 👍

PR #1345, opened 2017-06-25
Added FindJsonCpp.cmake

AnotherFoxGuy:


tritonas00:
Doesn't work for me: https://pastebin.com/8e20dm3n

tritonas00:
Ok i found the problem, json features.h conflicted with system features.h, so i renamed json features.h and the problem gone

PR #1346, opened 2017-06-28
pssm material allowance

fidoj:
this allows pssm to work on terrains with 6 materials. for backwards compatibility.
screenshot_2017-06-29_05-21-39_1
the problems stem with limiting the blend maps, this can cause some uneven material edges.
this needs to be discussed and balanced.

### DIFF
--- a/source/main/terrain/OgreTerrainPSSMMaterialGenerator.cpp
+++ b/source/main/terrain/OgreTerrainPSSMMaterialGenerator.cpp
@@ -229,7 +229,7 @@ uint8 TerrainPSSMMaterialGenerator::SM2Profile::getMaxLayers(const Terrain* terr
     }
 
     // each layer needs 2.25 units (1xdiffusespec, 1xnormalheight, 0.25xblend)
-    return static_cast<uint8>(freeTextureUnits / 2.25f);
+    return static_cast<uint8>(freeTextureUnits / 2.16f);
 }
 
 //---------------------------------------------------------------------


only-a-ptr:
Heh, to be honest I lack knowledge in this area. My thoughts/questions:
* where does the initial `freeTextureUnits = 16` budget come from? Doesn't this differ based on HW?
* I didn't even know you can share/divide/whatever texture units like this...
* `// each layer needs 2.25 units (1xdiffusespec, 1xnormalheight, 0.25xblend)` < --The comment suggests a "texture unit" is a 4-channel "RGBA" storage. "diffusespec" means RGB=diffuse, A=specular; "normalheight" means RGB=normalmap, A=heightmap. Using 0.25 of a texture unit makes sense to me - we simply use 1 channel out of 4. But 0.16 of a texture unit? To use 0.64 out of 4 channels? I'm baffled.
* Anyway, we could drop the "height" portion of "normalheight" texture - the heightmap is only useful for [parallax mapping](http://wiki.polycount.com/wiki/Parallax_Map) and may be used for [improved texture blending](http://www.gamasutra.com/blogs/AndreyMishkinis/20130716/196339/Advanced_Terrain_Texture_Splatting.php) - but we do none of that.
* AFAIK the usual solution for limited number of terrain material layers is to partition the terrain into 'pages' - each page is a separate mesh and can use different materials. OgreTerrain supports it but it's currently broken in RoR - when me&Ulteq optimized terrain collisions, we dropped support of it, we always read from page(0,0)

fidoj:
i read it as a shader limit, it allows 16 layers at 2.25 layers each 7.5 layers, the blend maps when generated are half maps that are stored, they become a physical media and use layers the 1.6 is 6*0.16 0.96 (as close to 1.0 without 0.166666666666666667, remember when its a limit rounding up will result in a failure 0.17 0.102

fidoj:
normal height is bump mapping. im using that on my rdesv3 map
screenshot_2017-06-25_03-55-35_1
i dont think dropping that is wise. another solution since shadows are a blend hows about making it so shadows uses a space of 0.5 textures. inverse to what ive done with blend.

only-a-ptr:
@fidoj Nope, "normalheight" carries 2 separate image maps, just like "diffusespecular":
* RGB = normal map, does all of normal mapping. 
* A = heightmap. Usable for "bump mapping", but we don't do that. Also usable for parallax or blending, but AFAIK we don't do that either.

Also, bump mapping and normal mapping are 2 different techniques to archieve the same result, although most internet resources confuse it:
* Normal mapping is a technique where surface normals are encoded into colors: a color value of 0.0 - 1.0 maps to angle -90 - +90 degrees; RGB components read as XYZ (order may be different).
* Bump mapping is a technique where normals are deduced from a grayscale heightmap by comparing values of neighbour pixels. Older technique - less resource demanding, but less accurate.

I absolutely wasn't suggesting to drop normal mapping, it's possibly the most important basic shader effect to date - cheap and extremely visually appealing. :smile:

Michael10055:
Should this be merged? It's kind of a "hacky" way to fix the issue, but it works.

only-a-ptr:
@Michael10055 If it works, then it should be merged. I'm unable to test enough to detect possible issues. Also, I don't really understand how it's working, I need to ask on OGRE forums to be sure. However, if you say it works, then I guess it works.

I'm merging now. If it causes issues, we can always revisit it and make it configurable in terrn2 or revert it.

Issue #1352, opened 2017-07-12
Ability to change cinecam node weight


Michael10055:
This was brought up since we're doing a lightweight vehicle challenge. 
The cinecam node weight seems to be hardcoded at 20kg (as found [here](https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/physics/Beam.cpp#L921) by @disloyalpick)  and cannot be changed using `set_node_defaults`. It should be possible to change the weight of the cinecam node.

disloyalpick:
The ability to change the weight of it would be great. If I knew how to edit the code to make a customizable, I would. Lol

Michael10055:
I changed the weight to 1kg as a test, using Charger's n/b:
![ror_2017-07-12_12-43-10](https://user-images.githubusercontent.com/11002490/28129233-97dc12a6-6700-11e7-9651-c7ccd48c2c64.png)
Of course, this change breaks all other mods since they're tuned to have the weight set to 20kg.

disloyalpick:
As long as it is tuneable via set_node_defaults it should not break any other mods. We could leave it as 20KG default weight.

Without using a set_node_defaults line the cinecam node will default to the 20KG and thus not break a single mod.

Michael10055:
We could also add another option to the cinecam section. Defaults to 20kg if no option is present.

only-a-ptr:
[ 🎶 Schools out for summer 🎵 ](https://www.youtube.com/watch?v=JpM6_EBQoLs&feature=youtu.be&t=44s) 😆 

I don't recall ever seeing so much activity around here 👍 

Anyway, good spot, it's really hardcoded for reasons unknown - it pre-dates our development history. Anyway, @disloyalpick 's suggestion would definitely break mods because they already use `set_beam_defaults` all over the place and they don't expect it to affect the cinecam. However, @Michael10055 's suggestion sounds just fine.

PR #1354, opened 2017-07-13
Added argument 'node_mass' to truckfile/cinecam

only-a-ptr:
This adds a new, optional argument to `cinecam` - mass of the generated node.
```
cinecam
;  x,   y,   z, <---------8 bindings--------->, Optional: spring, damping, node-mass
0.66, 2.0, 1.8, 75, 76, 77, 78, 73, 74, 53, 54,           8000.0,   800.0,        20
```

#closes https://github.com/RigsOfRods/rigs-of-rods/issues/1352

As usual, there are some extra refactors to test 😎


Michael10055:
Hm. I'm kinda confused here.

Charger's lightweight n/b just explodes now, even without the cinecam.
Gavril MV4S explodes as well. However, the Gavril Omega doesn't. There's no differences between the two vehicle's cinecam sections.

only-a-ptr:
@Michael10055 My fault, sorry. My change was short-sighted. I ignored all the node mass adjustments happening after cinecam processing: https://github.com/RigsOfRods/rigs-of-rods/blob/e053e4a504026e84cada331e34f0294ea1dac5a7/source/main/physics/Beam.cpp#L867-L918. 

Fixing in progress....

Michael10055:
Works!

```
cinecam
0.0, 1.218, 0.0, 0, 0, 1, 40, 2, 2, 3, 3, 4000, 150, 1
```

ror_2017-07-14_11-20-19

All other mods seem to work exactly as before. 👍

tritonas00:
I can't notice any difference either, no crashes so far

only-a-ptr:
Please try changing gravity, too. That's a single moment when node masses get re-calculated. It's also the only possible source of problems I can think of.

Michael10055:
screenshot_2017-07-17_14-36-00_1

No issues with changing the gravity.

PR #1371, opened 2017-08-29
Improved Envmap (realtime reflections)

only-a-ptr:
Fixes the old bug where reflections of first terrain used would remain in all other terrains until game restart.

Additionally, removes weird old code I found - when player doesn't drive a vehicle, RoR would render the envmap at middle of the map in 50m height.  After this change, the rendering will simply stop, so when you exit a vehicle, it will retain it's reflections.


tritonas00:
Looks good

Issue #1372, opened 2017-08-29
Angelscript addons broken when building with VS2017


AnotherFoxGuy:
The recent MSVC compiler update broke our angelscript_addons
Apparently [ ``` asMETHODPR(string,resize,(size_t),void)```](https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/angelscript_addons/scriptstring/scriptstring.cpp#L818) isn't valid anymore

See: https://ci.appveyor.com/project/AnotherFoxGuy/rigs-of-rods/build/0.4.8.0-CIBuild-3
```
"C:\projects\rigs-of-rods\RoR.sln" (default target) (1) ->
       "C:\projects\rigs-of-rods\source\angelscript_addons\angelscript_addons.vcxproj.metaproj" (default target) (4) ->
       "C:\projects\rigs-of-rods\source\angelscript_addons\angelscript_addons.vcxproj" (default target) (9) -> (ClCompile target) -> 
         C:\projects\rigs-of-rods\source\angelscript_addons\scriptstring\scriptstring.cpp(744): error C2440: 'static_cast': cannot convert from 'void (__thiscall std::basic_string<char,std::char_traits<char>,std::allocator<char>>::* )(const unsigned int,const _Elem)' to 'void (__thiscall std::basic_string<char,std::char_traits<char>,std::allocator<char>>::* )(::size_t)' [C:\projects\rigs-of-rods\source\angelscript_addons\angelscript_addons.vcxproj]
         C:\projects\rigs-of-rods\source\angelscript_addons\scriptstring\scriptstring.cpp(744): error C2672: 'AngelScript::asSMethodPtr<4>::Convert': no matching overloaded function found [C:\projects\rigs-of-rods\source\angelscript_addons\angelscript_addons.vcxproj]
         C:\projects\rigs-of-rods\source\angelscript_addons\scriptstring\scriptstring.cpp(744): error C2660: 'AngelScript::asIScriptEngine::RegisterObjectMethod': function does not take 3 arguments [C:\projects\rigs-of-rods\source\angelscript_addons\angelscript_addons.vcxproj]
[...]

PR #999, opened 2016-05-30
[Experimental] Updated AngelScript to version 2.31.0

ulteq:
```
/home/travis/build/RigsOfRods/rigs-of-rods/source/angelscript_addons/datetime/datetime.h:12:60: warning: missing terminating ' character
 #error Sorry, this requires C++11 which your compiler doesn't appear to support
```

Issue #1364, opened 2017-08-07
Prebuilt packages for Linux distributions


AnotherFoxGuy:
### TODO:
- [x] Fix the RoRConfig crash (#1400)
~~~Fix the Caelum segmentation fault (#550)~~~
- [ ] Remove the nvidia-cg-toolkit (#1412)

AnotherFoxGuy:
### Dependencies status on package repositories
https://repology.org/metapackage/ogre
https://repology.org/metapackage/mygui
https://repology.org/metapackage/angelscript
https://repology.org/metapackage/openal-soft
https://repology.org/metapackage/ois

Issue #441, opened 2015-09-27
Does not compile with latest angelscript


anordal:
RoR doesn't compile with angelscript 2.29. I got «No conversion from int\* to char**» or something. Installing angelscript 2.22 instead fixed it.

On OpenSuse Tumbleweed, that means installing angelscript_2221-devel instead of angelscript-devel.

Maybe this could be put into the cmake detection script, or just the compilation instructions, as I nearly gave up before finding this out.

Hiradur:
Sorry for the inconvenience. We are aware of this and the required AS version is mentioned here: http://www.rigsofrods.com/wiki/pages/Compiling_3rd_party_libraries#Versions

AS needs to be updated ASAP, also due to #174 

It's on top of my todo list but the PSU of my desktop computer died so at the moment I can't work on it.

Hiradur:
I'm unassigning myself because I don't think I'll have the time and patience to do this anytime soon.

All I can say is that this more complex than the usual dependency upgrade. Fixing compiler errors doesn't seem to be enough. Runtime errors seem to be the real problem here. Some functions may need to be rewritten.
neorej16's attempt to update to AngelScript 2.26.3 may be of help but I don't know what its status is.
https://sourceforge.net/u/neorej16/rigsofrods/commit_browser

Hubbitus:
Compile fixes I have had also made before... It is not so hard and is not also helpful either Unfortunately.

Is there PR of neorej16's attempt of porting?

Hiradur:
The latest attempt to upgrade AS is in this PR: https://github.com/RigsOfRods/rigs-of-rods/pull/999
But fixing compiler errors is not the problem here, the tricky part are the runtime errors.

AnotherFoxGuy:
Fixed in #1373

PR #1373, opened 2017-08-30
Modernizing AngelScript

only-a-ptr:
AngelScript in RoR needs update: http://www.angelcode.com/angelscript/downloads.html

* We're using version 2.22.1 (dated 2011)
* The current stable is 2.31.2 (dated december 2016)

Before the update itself, I want to audit the existing scripting subsystem and remove what we don't really need. I already found some garbage, see commits below.

See also:
* https://github.com/RigsOfRods/rigs-of-rods/issues/1372 - Current addons break with VS2017
* https://github.com/RigsOfRods/rigs-of-rods/pull/999 - Old PR which updates angelscript, useful as reference.
* https://sourceforge.net/u/neorej16/rigsofrods/commit_browser - Old AngelScript update by Neorej16


AnotherFoxGuy:
Also related: #441

only-a-ptr:
Status: build and starts up for me. However, when I try loading Auriga map -> :boom:

only-a-ptr:
Status: Auriga loads fine for me, but the time-attack race in there doesn't work.

only-a-ptr:
Status: Auriga loads fine for me, but the time-attack race in there doesn't work.

only-a-ptr:
Status: I've just raced in Auriga with latest AngelScript! Give me some more maps/scenarios to test!

tritonas00:
> I've just raced in Auriga with latest AngelScript!

https://www.youtube.com/watch?v=4QPdn84DuGc

:smile:

AnotherFoxGuy:
> @only-a-ptr Give me some more maps/scenarios to test!

http://anotherfoxguy.com/Files/Benchmark.zip

tritonas00:
I'm getting: https://pastebin.com/ktshPxm2

tools/travis/linux/dependencies.sh
```diff
@@ -82,15 +82,15 @@ make install
...
+wget http://www.angelcode.com/angelscript/sdk/files/angelscript_2.31.2.zip
 unzip -qq angelscript_*.zip
 cd sdk/angelscript/projects/cmake
 cmake -DCMAKE_INSTALL_PREFIX=$DEPS_INSTALL_DIR \
 -DCMAKE_BUILD_TYPE:STRING=Release \
 -DCMAKE_CXX_FLAGS="-w -O0 -pipe" \
 .
 make -s -j2 
-cp -r ../../lib/libAngelscript.a $DEPS_INSTALL_DIR/lib/libangelscript.a
+cp -r ../../lib/libangelscript.a $DEPS_INSTALL_DIR/lib/libangelscript.a
```
AnotherFoxGuy: Try ```cp -r libangelscript.a $DEPS_INSTALL_DIR/lib/libangelscript.a```

only-a-ptr:
Eh, nothing's working.
I updated Travis dependencies, but it looks like angelscript's build output layout changed or something. It's failing with "no such file or directory". @AnotherFoxGuy I'll try it right away, thanks.
@tritonas00

tritonas00:
yep, now travis proceeded, same AddWaypoint error

tritonas00:
Compiled fine! Now, tried with benchmark map: https://pastebin.com/A0g8LTeE

only-a-ptr:
@tritonas00 Apparently the benchmark map is out of date. The missing function `navigateTo()` was removed by @AnotherFoxGuy himself in a26bd842777c48acacdf624abecf6ea9ed435ab6

tritonas00:
About the Futurotrack crash: https://pastebin.com/wDT1zdBb

only-a-ptr:
@tritonas00 Oops, my bad, that line is bound to crash whenever the searched function doesn't exist. Fix pending...

tritonas00:
compiles and crashes gone :thumbsup:

AnotherFoxGuy:
@only-a-ptr @tritonas00 The link for the benchmark terrain was old, here is a new one: https://www.dropbox.com/s/lgkjj2zurpjhab5/Benchmark.zip?dl=0

tritonas00:
Looks good!
2017-09-03-200517_1366x768_scrot

Why clang keeps failing?

tritonas00:
GPU Benchmark from Benchmark.zip segfaults, gdb output coming.

CPU Benchmark works fine.

only-a-ptr:
@tritonas00 

> Why clang keeps failing?

It's weird. Looks like Clang has somehow (is it even possible?)  disabled support for C++11, at least when building "angelscript_addons". It complains about lack of http://www.cplusplus.com/reference/type_traits/is_trivially_default_constructible/, although the respective header is definitely included, otherwise the offending code would get skipped.

only-a-ptr:
@tritonas00 

> GameScript.cpp line 372: https://github.com/RigsOfRods/rigs-of-rods/pull/1373/files#diff-6f778f681dbfde564fd26c6d79e9ddb7R372

Your excellent testing report pointed me straight to the bug - another unchecked handler-function lookup. I didn't have time to test the fix, hopefully it works.

tritonas00:
Crash fixed! But i get some errors: https://pastebin.com/UhzKMRSP

only-a-ptr:
@tritonas00 🤣 That's an extra error report I added - the error passed silently before. I had no idea it's going to be so noisy.

@AnotherFoxGuy What is the 'e' handler supposed to be?

AnotherFoxGuy:
> @only-a-ptr  What is the 'e' handler supposed to be?

Nothing, the handler isn't used since it doesn't have a collision box

only-a-ptr:
Eh, on second check, I actually introduced a bug in the last commit... the object should be spawned regardless of the handler being found 😊 

That, however, sucks because this way I can't report an error where the user actually wants to use a handler and it's not found by mistake. A solution would be constitute that passing empty string as param `eventhandler` to `spawnObject()`  means "don't search for handler".

tritonas00:
Now works fine, trees loaded and my gpu really sucks :laughing:

only-a-ptr:
It looks like this PR is ready to merge.

The only remaining issue is the Travis/clang build failing due to https://github.com/RigsOfRods/rigs-of-rods/pull/1373#issuecomment-326827928 @AnotherFoxGuy Can you update the build machine once this PR is merged?

AnotherFoxGuy:
> @only-a-ptr Can you update the build machine once this PR is merged?

Yes, I was planning on updating the Travis build anyway because of #1364

only-a-ptr:
Yay, this PR is what I call a success. 🎉 
Thanks everybody.

PR #1387, opened 2017-09-22
✨ Added CMake option for specifying a custom version

AnotherFoxGuy:
This can be used to create a custom build without modifying the cmake file

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,11 +1,8 @@
 cmake_minimum_required( VERSION 3.0.2 )
-
-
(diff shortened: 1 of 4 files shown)


only-a-ptr:
The code looks good, but what is the use-case of this feature?

AnotherFoxGuy:
> @only-a-ptr The code looks good, but what is the use-case of this feature?

At the moment the builds produced by appveyor don't have any build information
rorconfig_2017-09-22_13-24-27

 With this, it is very simple to change the version to the one specified by appveyor

rorconfig_2017-09-22_13-25-25

only-a-ptr:
👍

PR #1392, opened 2017-09-29
Fix memory corruption in VehicleAI.cpp

only-a-ptr:
One fix for potential segfault, and 2 extra cleanups.


only-a-ptr:
Closing this because after double-checking, I discovered I got confused. There wasn't a memory corruption, only VERY UGLY code in AngelScript's 'scriptdictionary.cpp'. I'll research further.

PR #1388, opened 2017-09-22
Replaced JsonCPP with RapidJSON

only-a-ptr:
I want to use RapidJSON only in the remade multiplayer lobby: https://github.com/RigsOfRods/rigs-of-rods/pull/1333#issuecomment-331213832 To make the transition smoother, I want to make the switch in upstream first.

Fixes #1331

PR #1393, opened 2017-09-30
✨ Replaced JsonCPP with RapidJSON && Updated travis

AnotherFoxGuy:
Same as #1388 with updated Travis build system


only-a-ptr:
👍

PR #1397, opened 2017-10-04
Removed collision object count limits on maps

### LINKED ISSUE #1395: Collision limit for maps.
So when .39+ was released, it was said that the collision limit for maps, had basically been removed, and you no longer had to worry about them. As I've been working on the Community map, I've noticed that not to be the case, as there seems to be a hard coded limit, that causes the log to repeatedly say `COLL: The hashtable is full.` once the limit is reached, thus freezing the game upon attempting to load said map.

only-a-ptr:
I think I nailed it. The messy code in 'Collisions.h/cpp' made it hard to see what's going on (I'm looking at you, estama!), but I figured it out. I quickly tested by driving Desperado buggy on Auriga - everything seemed to work just fine.

## Changes
* The collision box limit is now 1 million
* The collision triangle limit is now 4 billion
* The memory is dynamically sized - you can forget the 'collision-tris' entry from .terrn2
* The `COLL: The hashtable is full` error is history.

Fixes #1395

### DIFF
--- a/source/main/physics/collision/Collisions.cpp
+++ b/source/main/physics/collision/Collisions.cpp
@@ -113,7 +113,6 @@ using namespace Ogre;
 
 Collisions::Collisions(RoRFrameListener* sim_controller)
     : m_sim_controller(sim_controller)
-    , collision_count(0)
     , debugMode(false)
     , forcecam(false)
     , free_eventsource(0)
@@ -775,13 +774,6 @@ int Collisions::addCollisionTri(Vector3 p1, Vector3 p2, Vector3 p3, ground_model
 
     m_collision_tris.push_back(new_tri);
     return new_tri_index;
-}
-
-void Collisions::printStats()
-{
-    LOG("COLL: Collision system statistics:");
-    LOG("COLL: Cell size: "+TOSTRING((float)CELL_SIZE)+" m");
-    LOG("COLL: Hashtable collisions: "+TOSTRING(collision_count));
 }
 
 bool Collisions::envokeScriptCallback(collision_box_t *cbox, node_t *node)
--- a/source/main/physics/collision/Collisions.cpp
+++ b/source/main/physics/collision/Collisions.cpp
@@ -2,7 +2,7 @@
     This source file is part of Rigs of Rods
     Copyright 2005-2012 Pierre-Michel Ricordel
     Copyright 2007-2012 Thomas Fischer
-    Copyright 2013+     Petr Ohlidal & contributors
+    Copyright 2013-2017 Petr Ohlidal & contributors
 
     For more information, see http://www.rigsofrods.org/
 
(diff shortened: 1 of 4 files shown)


tritonas00:
Compiles fine, no crashes so far, :+1:  from me

Michael10055:
Hm, doesn't seem to work from my tests. A while back I tried merging both the original NeoQ and NeoQ2.0 into one map just to see if it was possible, it goes over the collision limit. This change doesn't really seem to make a difference - spams the log with  "hashtable full" errors after a certain point.
````
15:59:34: Mesh: Loading NeoQ2-0highway-section-AColisionante.mesh.
15:59:34: WARNING: NeoQ2-0highway-section-AColisionante.mesh is an older format ([MeshSerializer_v1.40]); you should upgrade it as soon as possible using the OgreMeshUpgrade tool.
15:59:34: Can't assign material ModeloColisionante to SubEntity of Ogre/MO1099 because this Material does not exist. Have you forgotten to define it in a .material script?
15:59:37: COLL: The hashtable is full.
```
And both maps definitely don't equal over 4 billion 😆

only-a-ptr:
@Michael10055 I checked and things are worse than I thought. Apparently this message has nothing to do with triangle/box count, it relates only to the internal hashtable. The problem is that it's not a regular hashtable but some pseudo-sparse hybrid, and it's aparently broken.

As I already said in #1395, analyzing what's wrong here isn't worth the work - the whole RoR static collision system would best be thrown away and replaced by collision detection facilities from some proven physics engine - I favor Bullet which explicitly offers this in the docs.

But I'll make one last try - I'll try to reduce the lookup system to an ordinary hashtable which can handle overflows, so it doesn't get "full".

DarthCainRor:
I'm going to post a test branch of the community map later that will cause this issue for testing.

DarthCainRor:
https://github.com/RigsOfRods-Community/community-map/tree/Collision-breaker

only-a-ptr:
I pushed a new update. I can't test with the community map right now (at work), I just quickly drove Desperado on Auriga again - collisions seemed just fine.

only-a-ptr:
I created a PR for the 'pointlight crash' (https://github.com/RigsOfRods/rigs-of-rods/pull/1398) and also included it here to test (will be excluded in the end).

Then I tested the CommunityMap 'collision breaker' edition... didn't break me!
screenshot_2017-10-05_18-13-34_1

Michael10055:
Seems to work fine now 👍 
image
The test map I mentioned in a earlier comment works now as well.

only-a-ptr:
@Michael10055 I'm thrilled that it's working. It may not look like a miracle on the surface, but as programming goes, I really surprised myself here. The Collisions code is dreadful, littered with ambiguous (dangerous!) programming constructs and downright ugly writing (I hate you, Estama! 😆  ). For the longest time, I was simply avoiding it, so it's relieving to know I can actually work with it.

Those `COL:` stats in RoR.log were probably ever meant for a developer, for example the `Cell size:` line just prints a constant value. I already removed couple lines from there. And the `hashtable collisions` is dead now, I removed the (ugly!) counting code.

Issue #1400, opened 2017-10-17
RoRConfig crashes when Documents\Rigs of Rods 0.4 doesn't exist 


AnotherFoxGuy:
**Expected behaviour:** RoRConfig should create Documents\Rigs of Rods 0.4 and extract skeleton.zip 
**Actual behaviour:** RoRConfig creates Documents\Rigs of Rods 0.4 and crashes

only-a-ptr:
I found the issue - something's wrong with the 'recursive_zipper.py' script; the generated zipfile contains only the files (with paths containing a subfolder), but no entries for the subfolders themselves. Hence the subfolders are never created and file unzipping fails.

I debugged with both auto-created 'skeleton.zip' (Windows7 SP1 64bit, Python 3.6.3) and a hand-created one (TotalCommander's internal zipper).

only-a-ptr:
Resolved by #1401

PR #1401, opened 2017-10-18
Hotfix for RoRConfig unpacking 'skeleton.zip'

only-a-ptr:
See #1400

### DIFF
--- a/source/configurator/SystemTools.cpp
+++ b/source/configurator/SystemTools.cpp
@@ -115,55 +115,60 @@ wxString LoadInputDevicesInfo(WXWidget wx_window_handle)
 
 bool ExtractZipFiles(const wxString& aZipFile, const wxString& aTargetDir)
 {
-    // from http://wiki.wxwidgets.org/WxZipInputStream
-    bool ret = true;
-    //wxFileSystem fs;
-    std::unique_ptr<wxZipEntry> entry(new wxZipEntry());
-    do
+    wxFileInputStream in(aZipFile);
+    if (! in.IsOk())
     {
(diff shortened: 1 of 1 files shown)


only-a-ptr:
@AnotherFoxGuy Are you talking about checking the "MyDocuments\Rigs of Rods 0.4" itself, or the individual subdirs? In the latter case I oppose because we couldn't use this method for creating empty directories: cache, terrains, vehicles... We'd have to revert to using the dummy 'empty' files, which would be a huge step back. That's what directory entries in ZIP format are for.

AnotherFoxGuy:
At the moment the code is this:
``` c++
wxFileOutputStream file(name);

if (!file)
{
	// Complain that it can't write the file 
}
```
I think it should have a check if the folder exists before trying to write to it:
``` c++
if(!wxDir::Exists(dir))
{
	wxFileName::Mkdir(dir);
}
 
wxFileOutputStream file(name);

if (!file)
{
	// Complain that it can't write the file 
}
```

> @only-a-ptr revert to using the dummy 'empty' files

We need the empty files anyway since GitHub doesn't save empty directories
vivaldi_2017-10-19_11-52-58

only-a-ptr:
> @AnotherFoxGuy We need the empty files anyway since GitHub doesn't save empty directories

[You're right](https://stackoverflow.com/a/115992). I'm stumped.

Allright then, directory check it is. However, it won't be as elegant as your snippet because we must extract the subdir(s) from [wxZipEntry's path](http://docs.wxwidgets.org/3.0/classwx_archive_entry.html#a7f7c8fc858136626dcdf1865e646cef9) manually.

Issue #539, opened 2016-01-07
Skidmarks are not on ground if you flip vehicle and skidmarks are made


disloyalpick:
So the vehicle is the 1990 Chevrolet kodiak modded by BrKillpack. Cant find link. May have been taken down. But should work with anything else that can flip easily.

![screenshot_29](https://cloud.githubusercontent.com/assets/13305403/12184164/ab6942e6-b561-11e5-9874-4f2b1be33b02.png)
![screenshot_30](https://cloud.githubusercontent.com/assets/13305403/12184163/ab5c2bec-b561-11e5-8a8d-abcff1f4eff0.png)

only-a-ptr:
:smile: 
This is tricky. It means skidmarks were never properly aligned with the ground, but relied on the wheel position.That means skidmarks on sloped/bumpy road would be glitched, too.

First step (bugfix) - align skidmarks properly with ground.

Second step (enhancement) - make skidmarks more physically accurate - detect how wide portion of the wheel actually touches ground and adjust the skidmark width accordingly.

DirtGamer301:
Just going to add, skidmarks aren't stopped by the ground either:
![screenshot_741](https://cloud.githubusercontent.com/assets/12358849/12195497/eb05c028-b5f7-11e5-83bb-fef532d87c10.jpg)

(I was so sure I already issued that but can't find anything...)

only-a-ptr:
I looked into it and found the issue - The function `Skidmark::updatePoint()` adds new skidmark polygons with exactly the same orientation as the wheel's axis. No account for terrain orientation.

The fix should be simple: Just query ground height at the positions of the axis endpoints, and add the new vertices to the resulting positions.

The problem is, skidmarks are broken in current upstream and I don't have time to test where it got broken. I need help testing. See also https://github.com/RigsOfRods/rigs-of-rods/pull/1363#issuecomment-320500411

tritonas00:
> The problem is, skidmarks are broken (not displaying) in current upstream and I don't have time to research where it got broken.

just compiled 84ef4dcd704bef386778e53c5024de7607946010 and skidmarks work fine.

only-a-ptr:
@tritonas00 Good tips, thanks, can you try building them?

tritonas00:
https://www.youtube.com/watch?v=E781QE7ZQK8

tritonas00:
[...]

PR #1314, opened 2017-04-09
Terrn2: 2 new options

only-a-ptr:
New features:

## Support for custom terrain material in "Terrn2" format.

This will enable porting the remaining 0.38 maps (such as ashes48's "around the mountain") that were blocked by lack of custom material support.

This is Auriga-0.4 map hacked to use it's heightmap for ground texture:
```
[General]
CustomMaterial=TEST_MAT_heightmap_for_texture
```

screenshot_2017-04-10_00-14-31_1

## Custom minimap image for Teleport.

Until now, the minimap relied on "Survey map" feature which can be disabled in config, thus making Teleport minimap blank.

Again, Auriga-0.4 with it's heightmap used for Teleport minimap:
```
[Teleport]
NavigationMapImage=31-heightmap.png
```
screenshot_2017-04-09_23-57-56_1

PR #1404, opened 2017-10-25
[WIP] Floating skidmarks fix

only-a-ptr:
I'm on my way to fix the infamous "floating skidmarks" glitch, see https://github.com/RigsOfRods/rigs-of-rods/issues/539

## Status: Still broken, refactored again, needs testing.

To fix the skidmarks, I need some extra info from the collision system - did the wheel touch only ground or also a static geometry? To get the info, I need to slightly modify the collision code, so I cleaned it up for myself first. Please test if collisions are working normally.

### DIFF
--- a/source/main/datatypes/node_t.h
+++ b/source/main/datatypes/node_t.h
@@ -19,6 +19,9 @@ struct node_t
     node_t()               { memset(this, 0, sizeof(node_t)); }
     node_t(size_t _pos)    { memset(this, 0, sizeof(node_t)); pos = static_cast<short>(_pos); }
 
+    inline bool IsWheel() const            { return iswheel != 0; }
+    inline bool IsWheelOuter() const       { return (iswheel != 0) && (iswheel % 2 != 0); }
+
     Ogre::Vector3 RelPosition; //!< relative to the local physics origin (one origin per truck) (shaky)
     Ogre::Vector3 AbsPosition; //!< absolute position in the world (shaky)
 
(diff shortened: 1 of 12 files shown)


tritonas00:
I can't notice any difference

only-a-ptr:
Is it just me or are skidmarks rather cr*ppy in the current upstream? They look decent when you drive straight (altough there are gaps in them), but when you're drifting, the result is a fuzzy mess. The trails roughly follow the trajectory of the wheel, but they have nothing to do with it's direction and angle. I also observed that drifting easily generates skidmarks which protrude up to the air (and possibly also under the ground, didn't check).

Anyway, today's revision is another refactor, aimed to get more precise 'skidding data' from the physics code. I tested and it works, and it feels like more skidmarks are made now.

Michael10055:
Skidmarks have always been glitchy when sliding/drifting, even on 0.36.3:
screenshot_71
screenshot_70
Also since you're working on skidmarks now, would it be possible to bring back the ability for skidmarks to appear on other surfaces (such as dirt/sand)? They do on 0.36.3:
screenshot_68

only-a-ptr:
> @Michael10055 Skidmarks have always been glitchy when sliding/drifting, even on 0.36.3

Whew, thanks! It's always such a relief to know things didn't get broken on your watch.

> @Michael10055 ... skidmarks to appear on other surfaces (such as dirt/sand)? They do on 0.36.3:

This should actually still work, can you link me to the map you tested with? The builtin skidmarks are defined in [skidmarks.cfg](https://github.com/RigsOfRods/rigs-of-rods/blob/master/resources/skeleton/config/skidmarks.cfg) and they map to terrain using groundmodel name - so a map needs to use the predefined names to display skidmarks at all.

tritonas00:
> I'm pretty sure this comes from upstream

I can't reproduce the crash in current upstream. I tried with S10

Michael10055:
Okay, Maybe you were right.. Just crashed when backspacing the 2014 Superduty  (Single cab version) on a build with #1314 merged, which I doubt that PR would cause the same crash.

PR #1407, opened 2017-11-02
 ✨ Auto-configure plugins.cfg on Linux

AnotherFoxGuy:


only-a-ptr:
👍 

I checked where "plugins.cfg.in" comes from and discovered following:
* Official OGRE repo at Bitbucket doesn't contain any branches older than v2.0. The tags are still in place though, so I checked whether "plugins.cfg.in" was already present in v1.8.1 (which @tritonas00 uses). [Yes it's there](https://bitbucket.org/sinbad/ogre/src/525a7f3bcd4e51b1085c42a8b55fe07eb7700458/CMake/Templates/?at=v1-8-1) :)
* Github fork 'OGRECave' contains oldest branch 'v1.9'. The file is also present: https://github.com/OGRECave/ogre/blob/v1-9/CMake/Templates/plugins.cfg.in

PR #1409, opened 2017-11-12
Added console command spawnobject

AnotherFoxGuy:
This command spawns an odef object at the player's position 
For instance ```spawnobject opentruckshop``` spawns a opentruckshop at the player's position

This can be useful for quickly adding new objects and moving them with the [terrain editor](http://docs.rigsofrods.org/terrain-creation/editing-terrain-objects/)

### DIFF
--- a/source/main/gui/panels/GUI_GameConsole.cpp
+++ b/source/main/gui/panels/GUI_GameConsole.cpp
@@ -38,6 +38,7 @@
 #include "Scripting.h"
 #include "Settings.h"
 #include "TerrainManager.h"
+#include "TerrainObjectManager.h"
 #include "Utils.h"
 
 #if MYGUI_PLATFORM == MYGUI_PLATFORM_LINUX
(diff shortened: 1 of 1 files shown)


only-a-ptr:
Not a bad idea but this PR is a no-go 👎 

First, you're obscuring your revisions with a lot of formatting changes - again. Do a `:triangular_ruler: Just formatting` first and then the change later! Second, it's not acceptable to change console command format out of spite - people are going to be confused and report "console not working!" bugs. I admit I  also don't like the '/' format, but the most we can do is to make it optional.

Also, the "terrain editor" can't stay this way. I understand that any sort of editor is better than none, that's why I accepted the original code from Ulteq, but at the moment the "editor mode" works by hijacking code written for simulation - free camera, rorbot, notification box, console (your revision). This would soon become a maintenance nightmare. The editor mode absolutely needs it's own camera mode and GUI.

PR #1410, opened 2017-11-17
✨ Automatically set the path to the dependencies

AnotherFoxGuy:


only-a-ptr:
✔️ 
I didn't even know there were `ROR_DEPENDENCY_DIR` and `RoR_Dependencies_CMAKE_DIR` vars, I never noticed them in CMakeGui - apparently they're leftovers from the old non-cmake dependency pack.

PR #1333, opened 2017-06-05
New GUI library: Dear IMGUI

only-a-ptr:
imgui-preview

I want to drop our current library (MyGUI) and remake everything in [Dear IMGUI](https://github.com/ocornut/imgui) because:
- Much less code! MyGUI requires terrible amount of boilerplate code to set up bindings with XML definitions. There's code generator, but you still have to deal with the code afterwards. And even if you create GUI in code, you need a lot of boilerplate for simple things like [X] close button on windows.
- Automated layout! MyGUI does have a visual editor, but no layout tools - you need to position every single widget by hand. It gets very tiresome. With IMGUI, you just say "show 3 buttons, separator, label and 2 more buttons" and it's nicely laid out for you.
- Professional looks! The GUI looks like Blender or some other professional-level software - by default! MyGUI's skin system is awful and undocumented - for years, RoR had been stuck with it's default skin, re-painted to (awful!) orange. The current skin by Max98 is orders of magnitude better, but it's still not very clean.
- The GUI is actually capable! Just look at it's GitHub page's screenshots. All complex editor/diagnostic UIs! Graphs! Plots! MyGUI isn't really capable GUI system - even a multi-column selectable table (multiplayer lobby) was tricky. With RigEditor (unfinished, circa 2014), I ultimately gave up - MyGUI was slowing me down to a complete halt.
- The GUI is fast and __fun__ to create!

For a demonstration, I remade the top menubar. The old one behaved weird and was bloated. UPDATE: Now it's complete, the vehicles menu was restored.

PR #1414, opened 2017-11-26
New GUI system: DearIMGUI (restart)

only-a-ptr:
This is a lightweight edition of https://github.com/RigsOfRods/rigs-of-rods/pull/1333, only carrying the signifficant enhancements and omitting the "nice to have" stuff. I decided to split the branch because it had become convoluted, containing both important changes that I need in upstream ASAP and various other enhancements that can actually wait.

### Features:
* **GVars - final concept**: GVars are configuration variables which are read from RoR.cfg and in the future will be editable by in-game console. This system is inspired by Quake engine's "cvars". The documentation is in the repo's wiki: https://github.com/RigsOfRods/rigs-of-rods/wiki/GVars:-global-(console)-variables and will be added to our docs when ready.
* **DearIMGUI**: New GUI library, absolute must-have for further development, see #1333 for reasoning.
* **New MP serverlist**: The old one was lacking.
* **New settings window**: The old one was firmly tied to the "pre-GVars" way of handling ror.cfg, so I just deleted it altogether and created a new window. 
* **A lot of internal improvements**: When working on the above items, I ran into a lot of other things that were feasible to update - for easier maintenance and further development.

obrazek

Please test and see if anything got broken. If not, I'll merge this into upstream and I'll add the other changes in a following PRs.


Michael10055:
:+1:
I think RoR should return to the main menu if you press Escape while the settings window is open rather than exiting.

only-a-ptr:
@Michael10055 Very good point, I recall someone already asked this for the MP selector, and I forgot... I'll create a patch.

tritonas00:
next week i will start testing again

@only-a-ptr do you plan to complete the new gui and remove mygui for this release?

only-a-ptr:
> @tritonas00 do you plan to complete the new gui and remove mygui for this release?

Nope, too much work left to do. Aside from the GUI, MyGUI also powers the "*.layout" dashboard HUD system. And it generally works, so there's no rush. I'll save the features for 0.4.9 😄

PR #1415, opened 2017-11-30
[EXPERIMENT] Soundsources3 - without the '3'

only-a-ptr:
Hello everyone.

I recently answered a forum PM saying "OK, if soundsources3 (now removed from docs) are not coming, is there at least some workaround to link sounds to commands"? Just to show good will, I re-visited my old denouncing post on the forum and... I found out I was wrong there. Next, I checked the codebase - the unfinished soundsources3 code from v0.3x was still in place. Althought I hated it, I tried to build on top of it, just to see how it comes out.

Instead of adding 'soundsources3', I added extra named arguments to 'soundsources2' because it's more flexible - if more arguments are needed or some are dropped, it will not cause a mess in the truckfiles.

## Features

Section 'soundsources2' in Truckfile fileformat now allows linking soundscripts to individual command hydros ('commands' in RoR jargon) using 2 new parameters:
* link_type: can be 'command_extend' or 'command_retract'
* link_num: Index of the 'command' as defined in Truckfile; first item has index 0.

The new 2 params have form NAME=VALUE without space around '=' and go after all 'soundsources2' parameters. The ordering doesn't matter. Example:
```
soundsources2
; nodeNum, type, soundscriptName, optional: link_type, link_num
10, -2, mySound1, link_type=command_extend, link_num=0
11, -2, mySound2, link_num=0, link_type=command_retract
```

In Soundscript file, the value of required 'trigger_source' must be "user_defined".

## Status

Builds and runs, but I didn't test the new functionality - no time.


only-a-ptr:
@ulteq Can you take a look at this for me please? It's almost a year old 😊

only-a-ptr:
Closing as outdated, will be revisited in the future.

PR #1422, opened 2017-12-12
🐛 Fixed configure_file Problem

AnotherFoxGuy:


tritonas00:
works, thank you!

only-a-ptr:
@AnotherFoxGuy Glad it's working, but how? This revision just adds a "tmp" subdirectory to the CMakeFiles directory and puts the file there, otherwise it works the same.

PR #1427, opened 2017-12-15
Fixed `EnvmapUpdateRate` default value

Michael10055:
If `RoR.cfg` does not contain the `EnvmapUpdateRate` line (the Configurator does not write this value, so RoR writes it once you exit) the default value would be set to 0, thus causing vehicles to not have any reflections:
screenshot_2017-12-15_16-53-53_1
This sets the default value to `2` (from here: https://github.com/RigsOfRods/rigs-of-rods/wiki/GVars:-global-(console)-variables)


only-a-ptr:
👍 Nice spot.

Since the envmap settings (and most others) are now configurable in the new in-game settings window, we may just as well remove them from RoRConfig altogether.
obrazek

Issue #1256, opened 2016-12-04
Crash on loading miniredmars map


Bob-Z:
I have a crash when loading miniredmars map.

GDB back trace, Valgrind trace and dumb attempt follow:

**This is gdb back trace:**

Rig loading profiler report:
	Beam::Beam()                 | prepare loading: 0.000156 sec
	Beam::LoadTruck()            | open file:       0.000030 sec
	Beam::LoadTruck()            | create parser:   0.000011 sec
	Beam::LoadTruck()            | prepare parser:  0.000015 sec
	Beam::LoadTruck()            | run parser:      0.000211 sec
	Beam::LoadTruck()            | finalize parser: 0.000187 sec
	Beam::LoadTruck()            | post-parse:      0.000013 sec
	Beam::LoadTruck()            | setup validator: 0.000007 sec
	Beam::LoadTruck()            | run validator:   0.000002 sec
	Beam::LoadTruck()            | post-validation: 0.000007 sec
	Beam::LoadTruck()            | setup spawner:   0.000009 sec
	Beam::LoadTruck()            | add modules:     0.000000 sec
	Beam::LoadTruck()            | run spawner:     0.007217 sec
	Beam::LoadTruck()            | log spawner:     0.000093 sec
	Beam::LoadTruck()            | process fixes:   0.000001 sec
	Beam::LoadTruck()            | calc masses:     0.000016 sec
	Beam::LoadTruck()            | soundsources:    0.000018 sec
	Beam::LoadTruck()            | calc node graph: 0.000004 sec
	Beam::LoadTruck()            | bounding boxes:  0.000001 sec
	Beam::LoadTruck()            | groundmodel:     0.000094 sec
	Beam::LoadTruck()            | load dashboards: 0.000001 sec
	BeamFactory::createLocal()   | post-process:    0.000053 sec
	RigLoadingProfiler::Report() | compose report:  0.000010 sec
Widget property 'Text_TextAlign' have type prefix - use 'TextAlign' instead [MapEntity.layout]
Widget property 'Text_FontHeight' have type prefix - use 'FontHeight' instead [MapEntity.layout]
static map icon not found: icon_load.dds
 ===== LOADING VEHICLE: bombinette.load
 == Parsing vehicle file: bombinette.load
 == Parsing done OK, nothing to report

 == Validating vehicle: bombinette
 == Validating done OK
 == Spawning vehicle: bombinette
duplicate resource name 'DefaultDashFont'
[...]

PR #1430, opened 2017-12-18
Miniredmars fixes

only-a-ptr:
The damn bombinette crash took me almost all day, but I finally figured it out: https://github.com/RigsOfRods/rigs-of-rods/issues/1256#issuecomment-352469930

I'm taking tomorrow off. 😆 

UPDATE: I also slipped a minor internal fix here

### DIFF
--- a/source/main/gui/panels/GUI_GameSettings.cpp
+++ b/source/main/gui/panels/GUI_GameSettings.cpp
@@ -35,6 +35,7 @@ void RoR::GUI::GameSettings::Draw()
         {
             App::GetGuiManager()->SetVisible_GameMainMenu(true);
         }
+        ImGui::End();
         return;
     }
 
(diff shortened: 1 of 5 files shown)


tritonas00:
Great! Unfortunately i will be off for 2+ weeks so i can't test it right now, maybe @Michael10055 can

tritonas00:
works here also :+1: 

it's interesting though that the first loading always crashes, but then it works.

tritonas00:
about the first loading crash:

> OGRE EXCEPTION(5:ItemIdentityException): Cannot find a group named General-1 in ResourceGroupManager::isResourceGroupInitialised at /home/babis/Downloads/rigsofrods-next-build-debug/ogre/src/ogre_src_v1-8-1/OgreMain/src/OgreResourceGroupManager.cpp (line 1900)

no segfault

PR #1431, opened 2017-12-18
Made node debug size smaller

Michael10055:
This makes the visual node debug readable on complex n/b's, using the same size values found in 0.36's source (https://github.com/only-a-ptr/ror-legacy-svn-trunk/blob/dcb83de66935cb6b82d153fea2838e166419a7ac/build/main/source/Beam.cpp#L9633)
image


only-a-ptr:
👍

PR #1445, opened 2018-01-24
HOTFIX: jetplane afterburner sound always on

only-a-ptr:
Kudos to @tritonas00 for quickly finding and reporting!


tritonas00:
works fine :+1:

PR #1449, opened 2018-01-26
HOTFIX: Don't save adjusted FOV to RoR.cfg

only-a-ptr:
This should do it.

It builds, but I can't test right now. UPDATE: tested, behaves correctly.


Michael10055:
Awesome 👍

PR #1453, opened 2018-01-30
BUGFIX: Crashes when changing sky type in Settings

only-a-ptr:
It turned out there was a bug in loading resources based on configuration, so this crash could also happen if you toggled other effects. Kudos to @Speciesx for discovering and reporting.


tritonas00:
i'm off this week again :worried:

only-a-ptr:
@tritonas00 No problem 😄

PR #1456, opened 2018-02-04
Added console command spawnobject

AnotherFoxGuy:
The same as #1409, but now with backwards compatibility for console commands

---------------------
This command spawns an odef object at the player's position 
For instance, ```spawnobject opentruckshop``` spawns a opentruckshop at the player's position

This can be useful for quickly adding new objects and moving them with the [terrain editor](http://docs.rigsofrods.org/terrain-creation/editing-terrain-objects/)

### DIFF
--- a/source/main/gui/panels/GUI_GameConsole.cpp
+++ b/source/main/gui/panels/GUI_GameConsole.cpp
@@ -38,6 +38,7 @@
 #include "Scripting.h"
 #include "Settings.h"
 #include "TerrainManager.h"
+#include "TerrainObjectManager.h"
 #include "Utils.h"
 
 #if MYGUI_PLATFORM == MYGUI_PLATFORM_LINUX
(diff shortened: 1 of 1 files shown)


source/main/gui/panels/GUI_GameConsole.cpp
```diff
@@ -235,183 +236,219 @@ void Console::eventCommandAccept(MyGUI::Edit* _sender)
         // discard the empty message
         return;
     }
-        Ogre::StringVector args = StringUtil::split(msg, " ");
```
only-a-ptr: This formatting makes the whole diff unreadable. Again - do a separate 📐 revision for these pls.

source/main/gui/panels/GUI_GameConsole.cpp
```diff
@@ -235,183 +236,219 @@ void Console::eventCommandAccept(MyGUI::Edit* _sender)
...
+        if (!logging)
         {
-            // switch to console logging
-            bool now_logging = !App::diag_log_console_echo.GetActive();
-            const char* msg = (now_logging) ? " logging to console enabled" : " logging to console disabled";
-            this->putMessage(CONSOLE_MSGTYPE_INFO, CONSOLE_SYSTEM_NOTICE, _L(msg), "information.png");
-            App::diag_log_console_echo.SetActive(now_logging);
-            return;
+            putMessage(CONSOLE_MSGTYPE_INFO, CONSOLE_SYSTEM_NOTICE, _L(" logging to console enabled"), "information.png");
+            SETTINGS.setSetting("Enable Ingame Console", "Yes");
```
only-a-ptr: BSETTING won't work anymore because "Enable ingame console" is now transformed to GVar: https://github.com/RigsOfRods/rigs-of-rods/blob/025ff0a335b3be6afde16da677870f42506eb1fa/source/main/Application.cpp#L100 and thus it's deleted from the `Settings` hashmap: https://github.com/RigsOfRods/rigs-of-rods/blob/025ff0a335b3be6afde16da677870f42506eb1fa/source/main/utils/Settings.cpp#L823-L829

I'm thinking you just grabbed your old modified file and pasted it over this one, didn't you?

only-a-ptr:
Let's talk about this here: https://forum.rigsofrods.org/development-discussion/102-game-terrain-editor-mode.html#post591

only-a-ptr:
Allright, now it's fine. But I'd still like you to visit the forum thread and tell me how do you plan to continue supporting the editor mode.

Issue #50, opened 2015-02-13
Water reflection glitched when changing from terrain.


Max98:
The title says it all. Just wanted to put the issue on Github so I can take care of it later. If an other dev can find the solution before me, then it would be great! 

![screenshot_9](https://cloud.githubusercontent.com/assets/4823878/6196217/c1e18b6a-b3d6-11e4-8926-8f6bb6363cd6.png)

PR #1459, opened 2018-02-04
BUGFIX: Fixed HQ water glitch when re-entering simulation

only-a-ptr:
obrazek

Caused by generated textures not being properly unloaded.

Finally resolves a very old ticket by Max98: https://github.com/RigsOfRods/rigs-of-rods/issues/50

### DIFF
--- a/source/main/gfx/HydraxWater.cpp
+++ b/source/main/gfx/HydraxWater.cpp
@@ -166,12 +166,6 @@ void HydraxWater::updateReflectionPlane(float h)
 {
 }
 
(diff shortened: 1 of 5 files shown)


tritonas00:
Doesn't compile here (ogre 1.8)

https://pastebin.com/2Mss1nNv

only-a-ptr:
@tritonas00 This will fix it: https://github.com/RigsOfRods/rigs-of-rods/pull/1461

tritonas00:
hmm still getting the error

Issue #986, opened 2016-05-24
water surface is upside down


Speciesx:
Water type: reflection + refraction (speed and quality)
Reflection and 'water splashes'? are underwater.
https://youtu.be/opRj8fmQTtc

And here how it looks like with water type reflection.
https://youtu.be/ej2qi6rTh2M

ulteq:
Was it different in earlier versions?

Speciesx:
no

adriansnetlis:
In those videos it doesn't seem even 1 billionth bad as in mine. It even doesn't seem upside down. For me it is, though.

only-a-ptr:
I just tested and I couldn't reproduce the upside-down reflection which @adriansnetlis showed in the video. I'd have to see RoR.log from him, from the FPS of the video it seems he's using some weak troublemaker video chip.

ulteq:
Should be fixed with: https://github.com/RigsOfRods/rigs-of-rods/pull/2045

Issue #29, opened 2015-01-17
Water type Reflection + refraction (quality) a lot brighter than Reflection


Speciesx:
Reflection
![ror_2015_01_17_20_44_04_361](https://cloud.githubusercontent.com/assets/1196375/5790126/35f7e0e2-9e8a-11e4-8953-1d6b8b4df60f.png)

Reflection + refraction (quality)

![ror_2015_01_17_20_44_44_739](https://cloud.githubusercontent.com/assets/1196375/5790125/35f458b4-9e8a-11e4-8308-b3f722ca9f8c.png)

DirtGamer301:
![screenshot_597](https://cloud.githubusercontent.com/assets/12358849/12065504/f0dd3de8-afd8-11e5-94c8-e1f7e0141d38.jpg)
![screenshot_598](https://cloud.githubusercontent.com/assets/12358849/12065505/f0e0ce9a-afd8-11e5-84a9-15fb8f98100e.jpg)

PR #1461, opened 2018-02-05
[WIP] Water GFX fixes

only-a-ptr:
UPDATE: I added some extensive code cleanups, do NOT merge yet.

### DIFF
--- a/source/main/gameplay/Character.cpp
+++ b/source/main/gameplay/Character.cpp
@@ -377,7 +377,7 @@ void Character::update(float dt)
 
         if (gEnv->terrainManager->getWater())
         {
-            wheight = gEnv->terrainManager->getWater()->getHeightWaves(position);
+            wheight = gEnv->terrainManager->getWater()->CalcWavesHeight(position);
             if (position.y < wheight - 1.8f)
             {
                 position.y = wheight - 1.8f;
--- a/source/main/gameplay/RoRFrameListener.cpp
+++ b/source/main/gameplay/RoRFrameListener.cpp
@@ -1685,16 +1685,16 @@ bool RoRFrameListener::frameStarted(const FrameEvent& evt)
         IWater* water = gEnv->terrainManager->getWater();
         if (water)
         {
-            water->setCamera(gEnv->mainCamera);
+            water->WaterSetCamera(gEnv->mainCamera);
             if (curr_truck)
             {
-                water->moveTo(water->getHeightWaves(curr_truck->getPosition()));
+                water->SetReflectionPlaneHeight(water->CalcWavesHeight(curr_truck->getPosition()));
             }
             else
             {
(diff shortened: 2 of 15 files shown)


tritonas00:
adding `#include <OgreTexture.h>` to Water.h fixes it

only-a-ptr:
> @tritonas00 adding only #include <OgreTexture.h> to Water.h fixes it 😊

Meaning my revision is also working or not? If not, I guess OGRE 1.8 has the defs in different headers.

tritonas00:
> Meaning my revision is also working or not?

nope doesn't work for me

tritonas00:
just checked this PR again, now it works

tritonas00:
When you set water to reflection+refraction, swim and move the camera below the waterline this happens:
2018-02-07-172543_1366x768_scrot

Notice also the cursor is reflected too. Maybe you want to fix this in this PR.

only-a-ptr:
@tritonas00 Yes I want to fix it, though it's an old thing, I just tried v0.4.6 and it was already hapenning.

Speciesx:
water splashes also underwater https://youtu.be/opRj8fmQTtc?t=31s

only-a-ptr:
@Speciesx Is that new or old issue?

Speciesx:
@only-a-ptr 
Old one like https://github.com/RigsOfRods/rigs-of-rods/issues/986
Maybe you can also fix this one too?
https://github.com/RigsOfRods/rigs-of-rods/issues/29

only-a-ptr:
I'll see what I can do. I think I can fix the splashes, but the brightness issue looks complicated.

only-a-ptr:
Eh, just tested with WeaselBoat on Auriga... my FPS dropped to ~3 🤢 
Profiler showed this:
obrazek
I had "Waves=Yes" in RoR.cfg. I did suspect that code to be inefficient, but this is over the top...

Speciesx:
ror_2018_02_11_22_22_27_105
refraction has a gap above the water

refraction has pretty low resolution, even in quality mode.
https://youtu.be/EXlcGk93cEw

only-a-ptr:
> refraction has a gap above the water

Wasn't it always there? I mean the reflection always drifts somewhat away from the object.

> refraction has pretty low resolution, even in quality mode.

It's hardcoded (comes from OGRE sample), but makes sense to make it configurable.

tritonas00:
Looks good

only-a-ptr:
I can't believe I've let this rot here for *T W O   Y E A R S* 🤢 
My sense of time is genuine nonsense 😳 

I compared this to upstream and found out there is nothing of value here - all the cleanups made it to upstream in the meantime and the reflection code seems to be equivalent, too, minus some experiments that didn't really work out.

Closing and examining conscience.

PR #1466, opened 2018-02-10
Skyx integration

Speciesx:
It works pretty well, except back to menu crashes.


only-a-ptr:
Heh, with all the love SkyX received in the forum, I felt it was time to take it to upstream.

I think we can merge it as-is and I'll deal with the crash immediately afterwards, it's most likely some missing cleanup.

Michael10055:
Awesome 👍 

SkyX should be added to RoRConfig though, right now it gets reset back to Sandstorm.

PR #1473, opened 2018-02-16
Try Linux specific /usr/share resources path earlier

mariushoch:
Try this earlier, before falling back to the generic App::sys_process_dir/..

I didn't test this, but the problem seems obvious on my Fedora 26 install:
App::sys_process_dir/.. will always exist (given that the binary is there), thus
the /usr/share/... path (where the resources are) will never be probed/ selected.


tritonas00:
I will test next week

PR #1479, opened 2018-02-25
Fix NAV autopilot (Instrument Landing System)

atomgott:
This feature was broken for a very long time. If you enable the NAV mode in an airplane it will try to land on the closest airport equipped with locators (poles with red and white stripes).
See also http://docs.rigsofrods.org/gameplay/aircraft-handling/.

### DIFF
--- a/source/main/terrain/TerrainObjectManager.cpp
+++ b/source/main/terrain/TerrainObjectManager.cpp
@@ -105,7 +105,6 @@ TerrainObjectManager::~TerrainObjectManager()
         delete proceduralManager;
     }
     gEnv->sceneManager->destroyAllEntities();
-
 }
 
 void TerrainObjectManager::loadObjectConfigFile(Ogre::String odefname)
--- a/source/main/terrain/TerrainObjectManager.cpp
+++ b/source/main/terrain/TerrainObjectManager.cpp
@@ -105,6 +105,7 @@ TerrainObjectManager::~TerrainObjectManager()
         delete proceduralManager;
     }
     gEnv->sceneManager->destroyAllEntities();
+
 }
 
 void TerrainObjectManager::loadObjectConfigFile(Ogre::String odefname)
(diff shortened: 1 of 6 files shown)


source/main/terrain/TerrainObjectManager.cpp
```diff
@@ -862,34 +862,38 @@ void TerrainObjectManager::loadObject(const Ogre::String& name, const Ogre::Vect
...
+			localizers.push_back(loc);
             continue;
         }
         if (!strcmp("localizer-v", ptline))
         {
-            localizers[free_localizer].position = Vector3(pos.x, pos.y, pos.z);
-            localizers[free_localizer].rotation = rotation;
-            localizers[free_localizer].type = Autopilot::LOCALIZER_VERTICAL;
-            free_localizer++;
+			localizer_t loc;
```
only-a-ptr: Spaces, not tabs, please!

atomgott:
Thanks, I did that. 
I'm still not happy with the relevant code being called from the dashboard ui, so I looked where the other autopilot functions are called. It turns out they are called from updateVisuals in Beam, which doesn't really seem to be make more sense, right? But there at least it stays with other systems like the warning system and does not depend on the dashboard being updated. So should I put the function that checks the localizers there or is there a better solution?

only-a-ptr:
> @atomgott ...not happy with ... code being called from XXX ... which doesn't really seem to ... make sense, right?

Welcome to RoR codebase :laughing: Pretty much everything works by 'getting called at some point' using 'when things happen, update the other things' principle. I'm not happy with it either and I'm slowly moving things around to work more sensibly. The main repo is already polished to some degree and development continues here: https://github.com/only-a-ptr/rigs-of-rods/commits/dev-master-v4 I'd love you to join the effort. I'll start a forum thread and write down where I'm headed with the changes, so far there wasn't another dev who'd read such a thing.

THIS is awesome :heart: https://github.com/RigsOfRods/rigs-of-rods/pull/1479/commits/c6cf24967fee03fc35a198fc343c775e63e163c2

Issue #1468, opened 2018-02-11
Move SkyX out of the source tree


AnotherFoxGuy:
See #1065
If we move Hydrax we also should move SkyX to its own repo

only-a-ptr:
I'd prefer to leave SkyX in the source tree. We're planning to port to OGRE 1.10x in a close future and continue to OGRE 2.1+. This will likely require changes to SkyX. Also our terrain graphics aren't perfect (no dynamic lights etc...) and improving them may again require changes to SkyX. Adding an artificial boundary of separate repos would only complicate this.

SkyX development is fragmented, there's no single active development repo. There's https://bitbucket.org/ogreaddons/skyx/src but it only has C++ code, no shaders or media. The original links at http://www.paradise-sandbox.com/#hydraxskyx.php are dead, I don't even know where @Speciesx got the media files from. GitHub has some individual forks. Maintaining a public repo in this situation would require active development effort, and we don't have the manpower - I have enough to do with RoR alone without separating out subsystems.

PR #1483, opened 2018-03-06
🔧 Moved pagedgeometry into the source tree

AnotherFoxGuy:
See #1468  
Because we currently have Skyx and Hydrax is the source tree, I think is odd to have only one external ogre-plugin


tritonas00:
i use ogre 1.8 and pagedgeometry 1.8 branch https://github.com/RigsOfRods/ogre-pagedgeometry/tree/Ogre-1.8, will this work? 

will test asap

tritonas00:
yep, doesn't work here: https://pastebin.com/aE7wYReJ

AnotherFoxGuy:
@tritonas00 It should be fixed now

tritonas00:
still the same error

tritonas00:
ok now i get: https://pastebin.com/uqsnAWcc

tritonas00:
now at 50% i get: https://pastebin.com/uqsnAWcc

AnotherFoxGuy:
@tritonas00 Coud you try running ```git reset --hard 749603db14981a076d346696aee16c4f7af9e418``` before builing?

tritonas00:
this goes me back to m_reflection_rtt_texture error in water.ccp (fix is in https://github.com/RigsOfRods/rigs-of-rods/pull/1461)

tritonas00:
maybe its a simple statement error? (error: expected ‘;’ before ‘if’)

https://pastebin.com/wU7DBBB3

AnotherFoxGuy:
I forgot a few semi-colons, it should be fixed now

tritonas00:
one staticCast left in GrassLoader.ccp :smiley: https://pastebin.com/kw6EAR5V

tritonas00:
now in WindBatchedGeometry.cpp https://pastebin.com/quNt2C5i

tritonas00:
it works, thank you! :heart:

Speciesx:
>Because we currently have Skyx and Hydrax is the source tree, I think is odd to have only one external ogre-plugin

But https://github.com/OGRECave/ogre-pagedgeometry still more or less maintained, Skyx and Hydrax not.

tritonas00:
well, this makes linux packaging a bit easier, one dependency less

paroj:
alternatively, you can add stuff as git submodules instead of copying it. For e.g. in https://github.com/OGRECave/ogre-imgui imgui is referenced that way.
At this you point to specific commits which allows updating everything in lock step.

tritonas00:
@AnotherFoxGuy @only-a-ptr @ulteq 

Since pagedgeometry is going into ror's source tree, like skyx and imgui, whould be nice to add caleum also?

So after ogre 1.11 update, humble users who just want to compile and play ror will have much less dependencies to deal with:

- sync pagedgeometry from ogrecave and integrate it to source
- sync caleum from ogrecave and integrate it to source
- fix skyx with ogre 1.11 or completely remove it if its dead 
- completely replace mygui with imgui

then ror will only need: angelscript wxgtk ogre and socketw

just a thought

only-a-ptr:
Time has passed and conditions have changed:
* SkyX is not doing so well - poor source code (I looked), bad player response (supposedly bad visual appearence, I didn't check).
* Caelum is doing great - it was recently [adopted by OGRECave](https://github.com/OGRECave/ogre-caelum) and updated to OGRE.11. I looked in the sources and they're way superior to SkyX's. It still uses the zombie-tech Cg shaders, but porting them is a question of time, I looked at it and they're short and self-contained, it should be a straightforward task.
* PagedGeometry was also adopted by OGRECave: https://github.com/OGRECave/ogre-pagedgeometry

Thus I propose a new approach:
* Ditch SkyX from codebase
* Attach OGRECave/Caelum as Git submodule and get it's shaders ported to modern tech.
* Attach OGRECave/PagedGeometry as Git submodule

AnotherFoxGuy:
I already started adding the OGRE plugins as Git submodules, I will create a branch for it later this week

AnotherFoxGuy:
Branch created: https://github.com/AnotherFoxGuy/rigs-of-rods/tree/ogre-submodules

PR #1501, opened 2018-05-10
🔧 Simplified the cmake build system

AnotherFoxGuy:
* CMake no longer depends on Python
* Removed duplicated ```find_package```
* Removed unused Python-scripts


only-a-ptr:
Neat :+1:

PR #1503, opened 2018-05-13
✨ Updated AngelScript to v2.32.0

### LINKED ISSUE #1508: Compilation error when running "game.sh" at 4%
I have run dependencies.sh from https://github.com/RigsOfRods/ror-linux-buildscripts and am now running game.sh, but it exits with the following error:

`[  4%] Linking CXX executable ../../bin/RoRConfig
/home/victoria/rigs-of-rods-source/rigs-of-rods/source/angelscript_addons/scriptstdstring/scriptstdstring.cpp: In function ‘void AngelScript::RegisterStdString_Native(AngelScript::asIScriptEngine*)’:
/home/victoria/rigs-of-rods-source/rigs-of-rods/source/angelscript_addons/scriptstdstring/scriptstdstring.cpp:744:93: error: no matching function for call to ‘AngelScript::asIScriptEngine::RegisterStringFactory(const char [15], AngelScript::asSFuncPtr, AngelScript::asECallConvTypes)’
  r = engine->RegisterStringFactory("const string &", asFUNCTION(StringFactory), asCALL_CDECL); assert( r >= 0 );`
       

How can I make the compilation work? It seems the function "engine->RegisterStringFactory("const string &", asFUNCTION(StringFactory), asCALL_CDECL)" does not exist, even though dependencies are installed.

I am using Debian 9, with the latest version of CMake and the latest version of game.sh

Thanks.

AnotherFoxGuy:
Fixes #1508

### DIFF
--- a/source/angelscript_addons/scriptbuilder/scriptbuilder.h
+++ b/source/angelscript_addons/scriptbuilder/scriptbuilder.h
@@ -1,6 +1,7 @@
 #ifndef SCRIPTBUILDER_H
 #define SCRIPTBUILDER_H
 
+
 // ###########################
 // Rigs of Rods - notes
 //
@@ -95,7 +96,7 @@ public:
 	std::string  GetSectionName(unsigned int idx) const;
 
 #if AS_PROCESS_METADATA == 1
-	// Get metadata declared for class types and interfaces
+	// Get metadata declared for classes, interfaces, and enums
 	const char *GetMetadataStringForType(int typeId);
 
 	// Get metadata declared for functions
@@ -107,7 +108,7 @@ public:
 	// Get metadata declared for class variables
 	const char *GetMetadataStringForTypeProperty(int typeId, int varIdx);
 
-	// Get metadata declared for class functions
+	// Get metadata declared for class methods
 	const char *GetMetadataStringForTypeMethod(int typeId, asIScriptFunction *method);
 #endif
 
@@ -115,7 +116,7 @@ protected:
 	void ClearAll();
 	int  Build();
 	int  ProcessScriptSection(const char *script, unsigned int length, const char *sectionname, int lineOffset);
-	virtual int  LoadScriptSection(const char *filename); // ## Rigs of Rods modification
+	virtual int  LoadScriptSection(const char *filename);
 	bool IncludeIfNotAlreadyIncluded(const char *filename);
 
 	int  SkipStatement(int pos);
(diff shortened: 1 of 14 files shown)


source/main/scripting/CBytecodeStream.cpp
```diff
@@ -35,18 +35,20 @@ CBytecodeStream::~CBytecodeStream()
         fclose(f);
 }
 
-void CBytecodeStream::Write(const void* ptr, AngelScript::asUINT size)
+int CBytecodeStream::Write(const void* ptr, AngelScript::asUINT size)
```
only-a-ptr: This change is not associated with AngelScript update, it's change of our own code and should have been in separate commit. 

Also, why return int when the indicated states are just "success/failure"?

source/angelscript_addons/scriptbuilder/scriptbuilder.h
```diff
@@ -107,15 +101,15 @@ class CScriptBuilder
...
+	// Get metadata declared for class methods
 	const char *GetMetadataStringForTypeMethod(int typeId, asIScriptFunction *method);
 #endif
 
 protected:
 	void ClearAll();
 	int  Build();
 	int  ProcessScriptSection(const char *script, unsigned int length, const char *sectionname, int lineOffset);
-	virtual int  LoadScriptSection(const char *filename); // ## Rigs of Rods modification
+	int  LoadScriptSection(const char *filename);
```
only-a-ptr: You broke the thing, RoR will be unable to load scripts. :slightly_frowning_face: The `virtual` hack was there for a reason. Without it, we'll end up calling the base implementation which only searches local paths, but we need to call our overload which uses OGRE resource system.

source/angelscript_addons/scriptbuilder/scriptbuilder.h
```diff
@@ -1,12 +1,6 @@
 #ifndef SCRIPTBUILDER_H
 #define SCRIPTBUILDER_H
 
-// ###########################
```
only-a-ptr: Don't remove my notes, they're there for a reason

source/main/scripting/CBytecodeStream.cpp
AnotherFoxGuy: From the changelog: ***Both Read and Write in asIBinaryStream can now return an error code to signal failure***
http://www.angelcode.com/angelscript/sdk/docs/manual/classas_i_binary_stream.html

So this change is required otherwise the build fails
```
error C2555: 'CBytecodeStream::Read': overriding virtual function return type differs and is not covariant from 'AngelScript::asIBinaryStream::Read'
```

only-a-ptr: You're right, I just ran into it while building.

PR #1509, opened 2018-05-24
 🐛 Fixed resources directory not being created on Linux

AnotherFoxGuy:


only-a-ptr:
I can't figure out how do those changes affect only Linux - there is no condition there. Can you describe it for me please?

AnotherFoxGuy:
Cmake recursively zips every directory in resources to a tmp folder
On Windows, cmake creates the directory if it doesn't exist, but Linux throws a file not found error

PR #1522, opened 2018-06-19
Fixed unzipping terrain causing materials not to load

### LINKED ISSUE #523: Unzipping terrain causes materials not to load
If you unzip a terrain the materials will not load

zipped

unzipped

AnotherFoxGuy:
Fixes #523

### DIFF
--- a/source/main/main.cpp
+++ b/source/main/main.cpp
@@ -221,6 +221,13 @@ int main(int argc, char *argv[])
 
         App::CreateCacheSystem(); // Reads GVars
 
+        // Initialize "managed materials"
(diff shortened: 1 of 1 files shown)


AnotherFoxGuy:
> @only-a-ptr If yes, please add it as code comment

There is already a code comment explaining it

>  // They must be initialized before any content is loaded, including mod-cache update.
   // Otherwise material links are unresolved and loading ends with an exception

only-a-ptr:
> There is already a code comment explaining it

Right, I forgot cache update actually parses materials

PR #1528, opened 2018-07-13
Hotfix: Multiplayer crash when 2nd client joins

### LINKED ISSUE #1527: Crash when connecting to multiplayer (Entity with name 'Character0_mesh` already exists)
image
Screenshot says it all really. To reproduce, just join the 2.40 Dev server with two RoR clients.

only-a-ptr:
I simply forgot to increment the counter :blush:
Fixes #1527


Michael10055:
:+1:

PR #1460, opened 2018-02-05
[WIP] Improved terrain loading fault tolerance

only-a-ptr:
Improved fault tolerance when loading terrain.

If the terrain can't be loaded for any reason, the player will be returned to menu with a message box "failed to load terrain, see RoR.log for details".

**CAUTION: I didn't test this code yet!**

Issue #1437, opened 2018-01-04
Crash when loading terrain with missing page file


Michael10055:
If a terrain is missing its page file (example: [N-Labs](https://archives.rigsofrods.org/repo/0.4-Terrains/CM%200.4.zip)) RoR will crash.
```
11:05:51:  ===== LOADING TERRAIN GEOMETRY CM.terrn2
11:05:51:   ## Loading Terrain Geometry
11:05:51: OGRE EXCEPTION(6:FileNotFoundException): Cannot locate resource CM-page-0-0.otc in resource group General or any other group. in ResourceGroupManager::openResource at E:\RoR-data\ror-dependencies\build\Source\ogre\OgreMain\src\OgreResourceGroupManager.cpp (line 756)
```
Terrain loads fine after adding the file (even if it is empty).

tritonas00:
same here

tritonas00:
shameless bump :smiley:

only-a-ptr:
Eh, allright then 😄

tritonas00:
shameless bump (version 2 :smile:)

only-a-ptr:
For whatever reason the fix didn't make it to upstream... I'll revisit it.

only-a-ptr:
Uhh, the terrn2 format is so under-designed. Anyway, after thoroughly examining the problem, I concluded:

* [the docs](http://docs.rigsofrods.org/terrain-creation/terrn2-subsystem/#ogre-terrain-page-config-otc) say that there must be a `page-X-Z.otc` file for every page, so let's leave it that way. 
* The workaround using empty `page*.otc` file is actually a bug, but I'll keep it for compatibility and comment it as such in the code.
* If the `page*.otc`  file is missing, RoR will return to menu and display error.
* The N-Labs map was updated to use the workaround, get it here: https://forum.rigsofrods.org/downloads.php?do=file&id=228
* If the modder wishes to disable loading heightmap, he can specify `Flat=true` in terrn2 file.
* Disabling the terrain geometry altogether is not supported and I'm OK with that.

Issue #1356, opened 2017-07-19
Vehicle's HUD in first person view


tritonas00:
Spawn a vehicle and press `c` to enter first person view. HUD disappears and that's fine. I think the user should be able to bring back the HUD by pressing `u` in first person view, like in previous versions, because in latest master that's not the case.

Thanks! :blush:

disloyalpick:
I agree. Would be useful for the vehicles that do not have a  dashboard/ gauges present in 1st person view.

Michael10055:
Do you know which commit prevented the HUD from showing?

tritonas00:
#1271

tritonas00:
is this an easy one for 0.4.8?

It's another RoR's gravestone (#1271) issue :laughing:

tritonas00:
fixed in #1538

PR #1538, opened 2018-08-23
Enhanced parallelism, much cleaner code (dev-master-9)

only-a-ptr:
Highlights:
* GUI improvements with DearIMGUI (top menubar, new skeletonview) showcased here: https://github.com/RigsOfRods/rigs-of-rods/pull/1333
* Tidy-up of physics data structures -> thinner, more sensible
* Project 'AsyncScene' -> major multithreading improvement, updating visual scene (flexbodies, flares, props+animations, cab, dashboard....) now runs in parallel with physics


tritonas00:
> /home/babis/Downloads/rigsofrods-next-build-db-1.8/rigsofrods/src/rigs-of-rods/source/main/gfx/DustManager.cpp: In member function ‘void RoR::GfxScene::InitScene(Ogre::SceneManager*)’:
/home/babis/Downloads/rigsofrods-next-build-db-1.8/rigsofrods/src/rigs-of-rods/source/main/gfx/DustManager.cpp:68:22: error: ‘make_unique’ is not a member of ‘std’
         m_heathaze = std::make_unique<HeatHaze>();

tritonas00:
It doesn't work

tritonas00:
2018-08-23-130010_1920x1080_scrot

:stuck_out_tongue_winking_eye:

AnotherFoxGuy:
@only-a-ptr For gcc you need to add ```set (CMAKE_CXX_STANDARD 14)``` to the main cmakelist.txt, see: https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html

tritonas00:
Apart from some wheels issues, everything seem to work fine

only-a-ptr:
>@tritonas00 Apart from some wheel issues, everything seem to work fine

I must have broken the wheels just yesterday, it always worked before

Anyway considering the scale of changes, there must be more glitches somewhere, look harder 😉

tritonas00:
I'm diggin it

tritonas00:
Another one:

Load simple terrain and spawn agora. Go to top bar see Actors, it shows agora fine.
Remove agora and respawn it. Go to top bar see Actors: empty

only-a-ptr:
I fixed the wheels and added the terrain fault tolerance fix. More to come.

tritonas00:
hmm
> /home/babis/Downloads/rigsofrods-next-build-db-1.8/rigsofrods/src/rigs-of-rods/source/main/gameplay/RoRFrameListener.cpp: In member function ‘bool SimController::CameraManagerMouseMoved(const OIS::MouseEvent&)’:
/home/babis/Downloads/rigsofrods-next-build-db-1.8/rigsofrods/src/rigs-of-rods/source/main/gameplay/RoRFrameListener.cpp:2351:44: error: ‘bool RoR::CameraManager::mouseMoved(const OIS::MouseEvent&)’ is protected within this context
     return m_camera_manager.mouseMoved(_arg);
                                            ^
In file included from /home/babis/Downloads/rigsofrods-next-build-db-1.8/rigsofrods/src/rigs-of-rods/source/main/gameplay/RoRFrameListener.h:25:0,
                 from /home/babis/Downloads/rigsofrods-next-build-db-1.8/rigsofrods/src/rigs-of-rods/source/main/gameplay/RoRFrameListener.cpp:21:
/home/babis/Downloads/rigsofrods-next-build-db-1.8/rigsofrods/src/rigs-of-rods/source/main/gfx/camera/CameraManager.h:148:10: note: declared protected here
     bool mouseMoved(const OIS::MouseEvent& _arg);

only-a-ptr:
@tritonas00 Why does Visual Studio always compile broken code? 🤢

tritonas00:
Update:

gavril reset crash from https://github.com/RigsOfRods/rigs-of-rods/pull/1538#issuecomment-415392614
happens on master also

tritonas00:
this https://github.com/RigsOfRods/rigs-of-rods/issues/1437 is still happening

tritonas00:
Another issue: 

Load Auriga map. Once loaded try to walk/move the ror bot. It doesn't go anywhere. Master works.

tritonas00:
I gathered all issues here: https://github.com/RigsOfRods/rigs-of-rods/pull/1538#issuecomment-415392614

ulteq:
Almost twice as much FPS for me with 20 trucks in the scene! I'm loving it.

Master: 23 FPS
This: 40 FPS

tritonas00:
This PR probably fixes also this: https://github.com/RigsOfRods/rigs-of-rods/issues/1356 because now i can do:

2018-09-01-210914_1366x768_scrot

Nice!

tritonas00:
@only-a-ptr please remove https://github.com/RigsOfRods/rigs-of-rods/pull/1538/commits/322bf6fa5f9393b5071947774312cb705bf669b9 from this PR, it doesn't work and you already stated that is broken: https://github.com/RigsOfRods/rigs-of-rods/pull/1460#issuecomment-365777452

only-a-ptr:
> @ulteq Almost twice as many FPS for me with 20 trucks in the scene! I'm loving it.

Heh, and it wasn't even motivated by performance, but rather code clarity. All sim first, all gfx later. The `bool doUpdate` mess in N/B processing was too much :sunglasses:

only-a-ptr:
I definitely resolved the N-Labs crash, see https://github.com/RigsOfRods/rigs-of-rods/issues/1437#issuecomment-417950544

tritonas00:
Ok at least it shows the error message. But there are other maps also like simple test terrain 12 cars that crash because of this. This https://github.com/RigsOfRods/rigs-of-rods/pull/1549 fixes the issue without modifying the maps. Please include that hot fix also. 

Why break something that worked? :confused:

only-a-ptr:
> @tritonas00 But there are other maps also...

In that case I'll have to take care of it, but know that it's not me who breaks things, it's the modders who create faulty terrains that only work by chance. The documentation is clear on this. @Michael10055 's _empty pagefile_ "fix" or @ulteq 's _1549_ "fix" are just hacks which create terrain page with absolutely bogus default parameters. :nauseated_face:

tritonas00:
fair enough

ulteq:
> with absolutely bogus default parameters.

Can't we provide proper default parameters?

only-a-ptr:
@ulteq Default parameters for terrain? Like 'if there's no heightmap, use [Mount Fuji](https://en.wikipedia.org/wiki/Mount_Fuji)?' 😆

It's our code, we can simply disable the OgreTerrain altogether if the modder doesn't want it. My point is, why think of workarounds when we can create sensible designs?

ulteq:
All we have to do to make the code backwards compatible is to remove those two lines:
https://github.com/only-a-ptr/rigs-of-rods/blob/cccce32df16e0463dbd7688cb565f9ef6cf833a0/source/main/terrain/TerrainGeometryManager.cpp#L331-L332
We can leave a warning in if you like.

ulteq:
Where did those go: https://github.com/RigsOfRods/rigs-of-rods/pull/1538/commits/fe4277795e848b3fe9aa8c903e51da71b8b4b6c2#diff-abbec6ed672585ca1ea0964858716deeL3783

Is it not needed anymore?

only-a-ptr:
@ulteq /unrelated/ it went here: https://github.com/RigsOfRods/rigs-of-rods/commit/fe4277795e848b3fe9aa8c903e51da71b8b4b6c2#diff-0b090e76f9de202df6134e778c74b554R2336

> All we would have to do, to make the code backwards compatible, is to remove this line...
Sure, go ahead, just add comment `// Ignoring error for backwards compatibility`

ulteq:
More bugs:

- Ties are invisble
- The interstate 8st trailer  spawns at (0,0,0)

tritonas00:
> But only if there really are multiple maps with this problem. @tritonas00, are there?

Only two i know.. @Michael10055 would help here...

source/main/gfx/GfxActor.cpp
```diff
@@ -1202,3 +1333,992 @@ void RoR::GfxActor::CalculateDriverPos(Ogre::Vector3& out_pos, Ogre::Quaternion&
...
+    for (prop_t& prop: m_props)
+    {
+        if (!prop.scene_node)
+            continue;
+
+        // -- quick ugly port from `Actor::NotifyCameraChanged()`~ 06/2018
+        const bool mo_visible = (prop.cameramode == -2 || prop.cameramode == m_simbuf.simbuf_cur_cinecam);
+        if (prop.mo != nullptr)
+        {
+            prop.mo->setVisible(mo_visible);
```
ulteq: This used to be `setMeshEnabled`.

source/main/gfx/GfxActor.cpp
```diff
@@ -1202,3 +1333,992 @@ void RoR::GfxActor::CalculateDriverPos(Ogre::Vector3& out_pos, Ogre::Quaternion&
...
+    }
+
+    NodeData* nodes = this->GetSimNodeBuffer();
+
+    for (prop_t& prop: m_props)
+    {
+        if (!prop.scene_node)
+            continue;
+
+        // -- quick ugly port from `Actor::NotifyCameraChanged()`~ 06/2018
```
ulteq: What if the prop is a `scene_node`?

only-a-ptr: > What if the prop is a scene_node?

I don't understand the question.

The `if (prop.scene_node == nullptr)` comes from the original code, somebody needs to fix it - no scenenode means setup failed and the prop should never make it to `m_props`.

source/main/gfx/GfxActor.cpp
only-a-ptr: Perhaps a mistake, I really don't know, the whole `MeshObject` is a piece of ****, see this branch I just made: https://github.com/only-a-ptr/rigs-of-rods/commits/MeshObject-cleanup

source/main/gfx/GfxActor.cpp
ulteq: I think the `// -- quick ugly port from `Actor::NotifyCameraChanged()`~ 06/2018` needs to always run.

only-a-ptr: Eh, that's just one of many things which need to be researched and cleaned up. I work in small iterations, first I cut+paste code with all quirks, then test, then I revisit it later. And everything related to the `AsyncScene` rewrite is in the 'pasted with quirks' stage.

ulteq: Yeah, but those subtle code changes are gonna bite us in the ass later on. :smile:

only-a-ptr: That's a topic for a long forum thread, bugs will always happen, but short commits help pinpointing them, even if it means messier code in the meantime (that's what comments are for). I wouldn't stress causing a few visual glitches, what troubles me in the long run are physics/collisions/scripting.

ulteq:
This does not work with 'Load' actors, which stay `SLEEPING` after being spawned: https://github.com/RigsOfRods/rigs-of-rods/blob/6839b19dd519e6881149c0acea7669ab08809545/source/main/physics/BeamFactory.cpp#L462-L476
Example trucks:
- https://forum.rigsofrods.org/downloads.php?do=file&id=170
- https://forum.rigsofrods.org/downloads.php?do=file&id=402

only-a-ptr:
@ulteq Good spot, the `isPreloadedWithTerrain` should really be `/*is anyhow spawned as SLEEPING*/`. 

I'm trying to get rid of all duplicate/redundant updates in the spawn code. Historically we had a single program flow for any sort of spawn (with map, from script, user action...) which needed to handle all specifics. I'm trying to split it apart.

tritonas00:
Another issue:

Spawn skyvan press F2 to open the back the door. Press F1. It stops. Press F1 again. It still goes down. When it goes full down press F1, it goes up.

Master works at is should: Pressing F2 or F21 instantly opens or closes the door.

ulteq:
`UpdateRods()` needs to account for `bm_disabled` and `bm_broken`.

only-a-ptr:
> @ulteq I think UpdateRods() needs to account for bm_disabled and bm_broken.

Precisely, I just postponed with this note: https://github.com/RigsOfRods/rigs-of-rods/commit/27835c8f08695d567b7a47d1ce69ed29086c21d7#diff-0b090e76f9de202df6134e778c74b554R786

ulteq:
> Precisely, I just postponed it with this note: https://github.com/RigsOfRods/rigs-of-rods/commit/27835c8f08695d567b7a47d1ce69ed29086c21d7#diff-0b090e76f9de202df6134e778c74b554R786

All right, then I'll leave that to you.

only-a-ptr:
What's left to fix/revert here? All the bugs listed above are fixed.

I'm currently working on spawn logic, but the changes are getting big and it's not exactly critical to do right now: https://github.com/only-a-ptr/rigs-of-rods/commits/devmaster9-spawn-codepath

PR #1539, opened 2018-08-24
removed leftover boost from configurator (Linux)

tritonas00:
Built deps and ror with and without boost and boost-libs, tested on Arch.


only-a-ptr:
:+1: I checked - there's no boost used in RoRConfig

PR #1544, opened 2018-08-31
Fix segfault on truck respawn

ulteq:
Fixes the following issue:
> Load gavril omega or skyvan and press Backspace to reset -> segfault. Happens in master also.
https://pastebin.com/DBQqDhzj
https://www.dropbox.com/s/poo57yd8b8e5w1s/gavrilomega.zip?dl=1

Did we remove the custom zero allocator?


tritonas00:
It works! But why appveyor fails?

only-a-ptr:
@ulteq Nice spot.

> Did we remove the custom zero allocator?

The `ZeroedMemoryAllocator` is still in there, but I'm trying to factor it out - it's a quirk and it only works when allocating manually using `new`, while I prefer to allocate on stack.

PR #1546, opened 2018-09-02
Fix '-Wreturn-type' warning

ulteq:
Not sure if this is how it is supposed to be, but it fixes the warning.


only-a-ptr:
I'm quite sure it's supposed to be this way: https://github.com/RigsOfRods/rigs-of-rods/pull/1538/commits/52dd2a57d9e3176771d37cb3b6029c71176f3bf9 :sunglasses:

PR #1552, opened 2018-09-04
Hotfix: Crash on loading map

only-a-ptr:
See #1541 
Needs testing.


only-a-ptr:
Confirmed as working" https://github.com/RigsOfRods/rigs-of-rods/issues/1541#issuecomment-418663273

PR #1557, opened 2018-09-06
Fixed segfault in 'SetPlayerActor'

ulteq:
Pressing 'COMMON_ENTER_NEXT_TRUCK' (default: CTRL+]) segfaults if you press it before spawning any trucks.


only-a-ptr:
👍

PR #1564, opened 2018-09-16
Fixes plugins.cfg generation on UNIX

ulteq:
Will this break Conan? Is there a simple way to make this work with both CMake and Conan?

### DIFF
--- a/source/main/plugins.cfg.in
+++ b/source/main/plugins.cfg.in
@@ -1,7 +1,7 @@
 # Defines plugins to load
 
 # Define plugin folder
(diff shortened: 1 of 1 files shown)


AnotherFoxGuy:
But what is exactly broken on master at the moment?

ulteq:
```
cmake -DCMAKE_PREFIX_PATH=~/ror-dependencies/Dependencies_Linux/ ./
make zip_and_copy_resources
```
-> plugins.cfg does not contain the `DCMAKE_PREFIX_PATH` path.

PR #1565, opened 2018-09-16
 🐛 Fixed location of plugins.cfg && Fixed crash on shutdown

AnotherFoxGuy:


only-a-ptr:
:+1: The MyGUI bug looks like something I did :blush:

PR #1566, opened 2018-09-18
Fixed 'IsNodeValid()' helper function

ulteq:
This fixes the orbit camera angle on trucks with `ar_camera_node_pos[0] == 0`.


only-a-ptr:
Good spot, pardon a little mind dump:
* we need to clean up the blatantly repetitive 'is cameranode valid' checks... it should be done once on spawn.
* we need to add something like `reference_box` to Truckfile to avoid abusing camera#0 as universal reference box.
* Ditto for using node#0 as universal position/velocity source.
* There is already some `getPositionAndStuff()` API in place, can you check it out?

ulteq:
> can you check it out?

Yes, but let's work through those first: https://github.com/only-a-ptr/rigs-of-rods/pulls

PR #1571, opened 2018-09-20
[WIP] Greatly simplified GVars and logging

only-a-ptr:
I couldn't resist doodling future scripting concepts, and I ran into actual need to access GVars by name from script/console (I absolutely plan that when I remake console with ImGUI). Also, the code which maps RoR.cfg entries to GVars is silly.

While thinking how to do it, it struck me how bloated the GVars code is for no good reason, except that I sucked at C++ at the time.

### DIFF
--- a/source/main/Application.cpp
+++ b/source/main/Application.cpp
@@ -59,109 +59,109 @@ static MumbleIntegration* g_mumble;
 static TerrainManager*  g_sim_terrain;
 
 // App
- GVarEnum_AP<AppState>    app_state               ("app_state",               nullptr,                     AppState::BOOTSTRAP,     AppState::MAIN_MENU);
- GVarStr_A<100>           app_language            ("app_language",            "Language",                  "English");
- GVarStr_A<50>            app_locale              ("app_locale",              "Language Short",            "en");
- GVarPod_A<bool>          app_multithread         ("app_multithread",         "Multi-threading",           true);
- GVarStr_AP<50>           app_screenshot_format   ("app_screenshot_format",   "Screenshot Format",         "jpg",                   "jpg");
+ GVar_AP<AppState>      app_state               ("app_state",               nullptr,                     AppState::BOOTSTRAP,     AppState::MAIN_MENU);
+ GVar_A<std::string>    app_language            ("app_language",            "Language",                  "English");
+ GVar_A<std::string>    app_locale              ("app_locale",              "Language Short",            "en");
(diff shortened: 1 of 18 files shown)


only-a-ptr:
@tritonas00 Buggy string type conversions, I'll fix it.

only-a-ptr:
@siradam7th Could you take a look at this for me?

I have string conversions of `RoR::Str<>` all wrong, the conversion operators sometimes kick in and sometimes not and I don't exactly understand why. I mostly worked it around with explicit `.ToCStr()` calls.

Also, as pointed out by @tritonas00, some `c_str()` conversions are missing.

only-a-ptr:
I couldn't resist...

Idea: instead of mixed legacy logging methods
* LOG() pseudomacro
* Log(), LogFormat() global functions
* LOGSTREAM macros

let's use following practice:
* We log to OGRE log and prefix our messages with [RoR|$TITLE] or at least [RoR] marker
* Code to log single message: `LogStream("logger-test") << "This" << " logger" << " works";`
* Helper function in major classes:
```
LogStream Log() { return LogStream("ClassName"); } // Usage: this->Log() << "stuff";
```

ulteq:
Can we please at least keep the old syntax?

only-a-ptr:
> Can we please at least keep the old syntax?

Which one, and for what area/purpose exactly?

I'd like the RoR.log to contain `[RoR]` entries from us and unprefixed entries from OGRE. Currently it's mixed, I added the prefixes ad-hoc to places, there are outdated prefix formats... Why not unify it?

I was thinking @siradam7th might help us cleaning this up.

ulteq:
> We can keep RoR::Log(const char*) and modify it to add [RoR] prefix automaticallly.

:+1:

AnotherFoxGuy:
> @only-a-ptr UPDATE: Or maybe we could be 😎 and use http://fmtlib.net

It is available on bintray: https://bintray.com/bincrafters/public-conan/fmt%3Abincrafters
So you can add `fmt/5.2.0@bincrafters/stable` to `conanfile.txt`,
run `conan install`
and add this to `source/main/CMakeLists.txt`:
```CMake
# --- fmt - Small, safe and fast formatting library ---
if(CONAN_FMT_ROOT)
  target_link_libraries(${BINNAME} PRIVATE CONAN_PKG::fmt)
  target_include_directories(${BINNAME} PRIVATE CONAN_PKG::fmt)
  target_compile_definitions(${BINNAME} PRIVATE USE_FMT)
else ()
```

And you can start implementing fmt

Issue #1577, opened 2018-09-25
Segfault when loading volvo-fmx-effer


tritonas00:
Load http://archives.rigsofrods.net/repo/files/repofiles-1st-batch/0f86053526958080c5313f88d092bdc377f7fddb_volvo-fmx-effer-1355.zip
```
Thread 6 "RoR" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffdeef8700 (LWP 22642)]
0x00005555558a7da2 in Actor::calcShocks2(int, float, float&, float&, bool) ()
    at /home/babis/Downloads/ror-dependencies-1.8/rigs-of-rods/source/main/physics/Beam.cpp:2531
2531	                        if (ar_beams[scount].shock && (ar_beams[scount].shock->flags & SHOCK_FLAG_ISTRIGGER)) // don't mess anything up if the user set the number too big
```
```
(gdb) bt
#0  0x00005555558a7da2 in Actor::calcShocks2(int, float, float&, float&, bool) ()
    at /home/babis/Downloads/ror-dependencies-1.8/rigs-of-rods/source/main/physics/Beam.cpp:2531
#1  0x00005555558bd211 in Actor::calcBeams(bool) ()
    at /home/babis/Downloads/ror-dependencies-1.8/rigs-of-rods/source/main/physics/BeamForcesEuler.cpp:1335
#2  0x00005555558beb42 in Actor::calcForcesEulerCompute(int, int) ()
    at /home/babis/Downloads/ror-dependencies-1.8/rigs-of-rods/source/main/physics/BeamForcesEuler.cpp:69
#3  0x00005555558aea51 in RoR::ActorManager::<lambda()>::operator() (__closure=0x7fffc8000b70)
    at /home/babis/Downloads/ror-dependencies-1.8/rigs-of-rods/source/main/physics/BeamFactory.cpp:1208
#4  std::_Function_handler<void(), RoR::ActorManager::UpdatePhysicsSimulation()::<lambda()> >::_M_invoke(const std::_Any_data &) (__functor=...) at /usr/include/c++/8.2.1/bits/std_function.h:297
#5  0x00005555558b32e4 in std::function<void ()>::operator()() const (this=0x7fffc8000b40)
    at /usr/include/c++/8.2.1/bits/std_function.h:682
[...]

PR #1579, opened 2018-09-26
:bug: Fixed segfault in 'calcShocks2'

ulteq:
Probably caused by: https://github.com/RigsOfRods/rigs-of-rods/commit/fe74eeb8e13e8924a19f31b12605ab49c6261154

Fixes: https://github.com/RigsOfRods/rigs-of-rods/issues/1577

### DIFF
--- a/source/main/physics/Beam.cpp
+++ b/source/main/physics/Beam.cpp
@@ -2568,7 +2568,7 @@ void Actor::calcShocks2(int beam_i, Real difftoBeamL, Real& k, Real& d, Real dt,
                     ar_beams[i].shock->trigger_switch_state = 0.0f;
                 if (ar_beams[i].shock->flags & SHOCK_FLAG_TRG_BLOCKER) // this is an enabled blocker and past boundary
                 {
-                    for (int scount = i + 1; scount <= i + ar_beams[i].shock->trigger_cmdshort; scount++) // (cycle blockerbeamID +1) to (blockerbeamID + beams to lock)
+                    for (int scount = i + 1; scount <= std::min(i + ar_beams[i].shock->trigger_cmdshort, ar_num_beams - 1); scount++) // (cycle blockerbeamID +1) to (blockerbeamID + beams to lock)
                     {
                         if (ar_beams[scount].shock && (ar_beams[scount].shock->flags & SHOCK_FLAG_ISTRIGGER)) // don't mess anything up if the user set the number too big
                         {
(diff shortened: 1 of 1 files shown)


tritonas00:
https://media.firebox.com/product/8301/column_grid_10/love-heart-neon-light_29822.jpg

only-a-ptr:
👎 Thanks for investigating, but this workaround is unacceptable as-is. 
* First off, it's very likely to introduce some subtle issues with triggers. I don't understand the code you changed and I already dislike shocks2 for being convoluted. If you know what's going on in there, comment it.
* Second, those values are set in Spawner and never change (only swap between cmdlong/cmdshort), so it's wrong to check & fix them in each physics step.

ulteq:
I'll try to find a better solution.

> Second, those values are set in Spawner and never change (only swap between cmdlong/cmdshort), so it's wrong to check & fix them in each physics step.

I also thought about detecting this in the RigSpawner, but I thought it wouldn't be possible.

> First off, it's very likely to introduce some subtle issues with triggers.

No, this is an exact replica of the behavior before: https://github.com/RigsOfRods/rigs-of-rods/commit/fe74eeb8e13e8924a19f31b12605ab49c6261154

> // don't mess anything up if the user set the number too big

Probably means don't crash by trying to dereference a `nullptr`.

source/main/physics/Beam.cpp
```diff
@@ -2568,7 +2568,7 @@ void Actor::calcShocks2(int beam_i, Real difftoBeamL, Real& k, Real& d, Real dt,
                     ar_beams[i].shock->trigger_switch_state = 0.0f;
                 if (ar_beams[i].shock->flags & SHOCK_FLAG_TRG_BLOCKER) // this is an enabled blocker and past boundary
                 {
-                    for (int scount = i + 1; scount <= i + ar_beams[i].shock->trigger_cmdshort; scount++) // (cycle blockerbeamID +1) to (blockerbeamID + beams to lock)
+                    for (int scount = i + 1; scount <= std::min(i + ar_beams[i].shock->trigger_cmdshort, ar_num_beams - 1); scount++) // (cycle blockerbeamID +1) to (blockerbeamID + beams to lock)
```
only-a-ptr: The  added `min()` is a hack which hasn't been here before: https://github.com/RigsOfRods/rigs-of-rods/blame/b45238e29512858b4f8e487c127698d8fa31b5d3/source/main/physics/Beam.cpp#L2606

This is what I had in mind when I talked about introducing subtle trigger bugs

ulteq: `ar_beams` used to be an array with a fixed number of elements. Hence no crash if you tried to access a beam beyond `ar_num_beams`.

`ar_beams[scount].shock` used to return false in those cases now it just crashes.

Using `std::min` is equivalent to `ar_beams[scount].shock == false`. The if block is not executed.

only-a-ptr: I'm positive the `ar_beams[scount].shock != nullptr` check doesn't mean "Am I still within beam count?" but simply "is this beam a shock?". 

RoR's physics are all done in means of "loop everything, switch over item type and process" and in this case, we're only filtering out shocks from all the beams. An optimal solution would be to have a separate array of `shock_t` with added beam index - then we could simply iterate over that. Also, separating actual shock absorbers from triggers would be nice.

ulteq: > was never intended to check "Am I still within beam count?"

We need to replicate the old behavior nonetheless.

> An optimal solution

I'm really only looking for a proper fix right now. We have way to many pending issues to even think about refactoring `calcShocks`.

> loop everything

If the truck file is setup correctly then this loop only loops over shocks.
The if check is only there as a "failsafe".

PR #1584, opened 2018-09-27
Fixed segfault in 'calcShocks2'

ulteq:
Fixes: #1577
Closes: #1579

### DIFF
--- a/source/main/physics/RigSpawner.cpp
+++ b/source/main/physics/RigSpawner.cpp
@@ -459,6 +459,17 @@ void ActorSpawner::FinalizeRig()
         m_actor->ar_engine->SetAutoMode(App::sim_gearbox_mode.GetActive());
     }
     
+    // Sanitize trigger_cmdshort and trigger_cmdlong
+    for (int i=0; i<m_actor->ar_num_beams; i++)
+    {
+        shock_t* shock = m_actor->ar_beams[i].shock;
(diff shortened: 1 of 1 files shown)


source/main/physics/RigSpawner.cpp
```diff
@@ -459,6 +459,17 @@ void ActorSpawner::FinalizeRig()
         m_actor->ar_engine->SetAutoMode(App::sim_gearbox_mode.GetActive());
     }
     
+    // Sanitize trigger_cmdshort and trigger_cmdlong
+    for (int i=0; i<m_actor->ar_num_beams; i++)
+    {
+        shock_t* shock = m_actor->ar_beams[i].shock;
+        if (shock && ((shock->flags & SHOCK_FLAG_TRG_BLOCKER) || (shock->flags & SHOCK_FLAG_TRG_BLOCKER_A)))
+        {
+            shock->trigger_cmdshort = std::min(shock->trigger_cmdshort, m_actor->ar_num_beams - i - 1);
```
only-a-ptr: is it (num_beams - (i - 1)), or ((num_beams - i) - 1)? From top of my head (ignoring context) I guess the latter, but please add parentheses to make it clear.

ulteq: The condition we need to meet is: `i + trigger_cmdshort < m_actor->ar_num_beams`

ulteq: I don't think that parentheses would increase readability. But feel free to add them.

PR #1585, opened 2018-09-29
Revamped Brakes, ABS and TC

ulteq:
Less code and:
- Almost completely removes brake induced wheel yitter
- Greatly improves braking on slopes ("Slopebrake")
- Greatly improves the anti lock system (ABS)
- Improves the traction control system (TC)

### DIFF
--- a/source/main/physics/BeamForcesEuler.cpp
+++ b/source/main/physics/BeamForcesEuler.cpp
@@ -568,6 +568,7 @@ void Actor::calcForcesEulerCompute(int step, int num_steps)
             ar_wheel_speed += newspeeds[i] / (float)m_num_proped_wheels;
         }
     }
+    ar_avg_wheel_speed = ar_avg_wheel_speed * 0.995 + ar_wheel_speed * 0.005;
 
     if (ar_engine && ar_num_wheels && ar_wheels[0].wh_radius > 0.0f)
     {
--- a/source/main/gameplay/LandVehicleSimulation.cpp
+++ b/source/main/gameplay/LandVehicleSimulation.cpp
@@ -120,7 +120,7 @@ void LandVehicleSimulation::UpdateCruiseControl(Actor* vehicle, float dt)
     }
     if (RoR::App::GetInputEngine()->getEventBoolValue(EV_TRUCK_CRUISE_CONTROL_READJUST))
     {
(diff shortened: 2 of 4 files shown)


only-a-ptr:
This is the kind of stuff you're better at. I'm always worried I'll subtly destabilize some mod I've never heard of.

ulteq:
> How much testing do you do before you draw the conclusions?

Hours of testing sometimes days. But I've also worked on that part of the code before. That makes it a lot easier. There is one last issue, then it's ready for final review.

source/main/physics/BeamForcesEuler.cpp
```diff
@@ -421,76 +423,58 @@ void Actor::calcForcesEulerCompute(int step, int num_steps)
...
                         brake_coef = 0.0f;
-                    }
-                }
-                else
-                {
-                    ar_wheels[i].firstLock = true;
                 }
 
-                if (ar_wheels[i].wh_speed > 0)
+                if (ar_wheels[i].wh_avg_speed > 0)
```
ulteq: This is the key to why this new code works so much better.

source/main/physics/BeamForcesEuler.cpp
```diff
@@ -401,6 +401,8 @@ void Actor::calcForcesEulerCompute(int step, int num_steps)
             total_torque = (m_num_axles == 0) ? engine_torque : intertorque[i];
         }
 
+        float wheel_slip = fabs(ar_wheels[i].wh_speed - curspeed) / std::max(1.0f, curspeed);
```
ulteq: Our old wheel slip calculation was trash.

source/main/physics/BeamForcesEuler.cpp
```diff
@@ -421,76 +423,58 @@ void Actor::calcForcesEulerCompute(int step, int num_steps)
...
-                        ar_wheels[i].wh_avg_speed = 0.0f;
-                        ar_wheels[i].firstLock = false;
-                    }
-                    // anti-jitter
-                    if (fabs(ar_wheels[i].wh_avg_speed) < 2.0f)
-                    {
-                        brake_coef = pow(fabs(ar_wheels[i].wh_speed), 2.0f);
-                    }
-                    else
+                    if (alb_pulse_state)
```
ulteq: The old ABS pulse logic was trash.

PR #1586, opened 2018-09-29
Unified actor-spawn/change API (using queues)

only-a-ptr:
Our spawning logic is very convoluted, there are lots of special cases and redundant reset requests that likely cancel each other out. My plan:
* Add queues for actor changes {spawn/modify/remove}
* Add 'pending player vehicle change' mechanism
* Process the queues + pending changes in one place -> less callbacks and ad-hoc state updates.
* Create unified 'spawn request' ticket which keeps all input data.

Known bugs:
* Terrain-preloaded actors are visible in top menubar. EDIT: this is upstream glitch, actually
* Not very thoroughly tested - expect glitches

### DIFF
--- a/source/main/gameplay/RoRFrameListener.cpp
+++ b/source/main/gameplay/RoRFrameListener.cpp
@@ -116,6 +116,7 @@ using namespace RoR;
 SimController::SimController(RoR::ForceFeedback* ff, RoR::SkidmarkConfig* skid_conf) :
     m_player_actor(nullptr),
     m_prev_player_actor(nullptr),
+    m_pending_player_actor(nullptr),
     m_actor_manager(),
     m_character_factory(),
     m_dir_arrow_pointed(Vector3::ZERO),
@@ -1106,7 +1107,7 @@ void SimController::UpdateInputEvents(float dt)
                         }
                         else
                         {
-                            this->SetPlayerActor(rescuer);
+                            this->SetPendingPlayerActor(rescuer);
                         }
                     }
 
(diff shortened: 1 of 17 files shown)


tritonas00:
> Terrain-preloaded actors are visible in top menubar. EDIT: this is upstream glitch, actually

imo not a glitch, you have pre spawned vehicles, they should be visible in the vehicles list so you can enter them easily from anywhere

only-a-ptr:
> @tritonas00 please don't change this behavior

I like it too, but it's not how it "worked before", which always makes me nervous.

tritonas00:
there is a `SPAWNER_PROFILE_SCOPED();` left in RigSpawner.cpp

only-a-ptr:
@tritonas00 >  vehicles spawn a bit below the ground

Tell me the full testing scenario please, or where it's best seen. Can you also pinpoint which commit causes it?

tritonas00:
Load nhelens spawn any vehicle

ulteq:
> Can you also pinpoint which commit causes it?

This one: https://github.com/RigsOfRods/rigs-of-rods/pull/1586/commits/0586d2fe7588903a2bd5be857229a825dd94e833

only-a-ptr:
> This belongs into the Actor constructor if you ask me:

* The slidenode locking? How, when slidenodes (or nodes for that matter) arent init in ctor?
* The NET info? Same thing
* eh, the GUI?

Constructors are overrated. Initially the Actor ctor invoked everything, including parsing truckfile + the actual spawn, and if it went wrong, a zombie object was still created (how else do you handle errors in ctor?) and later deleted. Ctors are for simple inits which won't easily fail. I'm allergic to their overuse.

ulteq:
Right, `ToggleSlideNodeLock()` does a lot more than I thought. I'm convinced.

PR #1588, opened 2018-09-30
Improved collision bounding boxes

only-a-ptr:
* New prediction method - OLD: "current aabb padded by node 0 velocity", NEW: "for each node, account current + future position (add velocity)".
* Removed duplicate code in ActorSpawner, Created new unified Actor method.
* Clarified 'bounding box ID' node attribute, added INVALID_BBOX constant.
* Improved spawning partial collision boxes - instead of clearing the std::vector and re-inserting AABBs every step, only insert on spawn and then reset+rebuild in-place.


ulteq:
> then reset+rebuild in-place.

Did you benchmark it against the old method?

ulteq:
> Finally containers spawn correctly
Weren't they already fixed by: https://github.com/RigsOfRods/rigs-of-rods/commit/c44f96fb25cfa30c04d5418ff41b5b2818fdfd86

tritonas00:
My memory has problems xD

PR #1594, opened 2018-10-01
Fixed debug view visibility in pause mode

ulteq:


only-a-ptr:
Don't be lazy and fork the repo instead of creating dev braches up here 😆

ulteq:
I don't even know why I switched the repository. I'll go back to mine on the next PR.

PR #1596, opened 2018-10-01
Final slope start tweak

ulteq:


source/main/physics/BeamForcesEuler.cpp
```diff
@@ -418,11 +418,9 @@ void Actor::calcForcesEulerCompute(int step, int num_steps)
...
                 ar_wheels[i].wh_tc_coef = pow(ar_wheels[i].wh_tc_coef, tc_ratio);
             }
-            if (!ar_engine || ar_engine->GetAcceleration() > ar_engine->getAccToHoldRPM(ar_engine->GetEngineRpm()))
-            {
-                total_torque *= pow(ar_wheels[i].wh_tc_coef, 1.0f / std::max(1.0f, std::min(wheel_slip, 10.0f)));
-                m_tractioncontrol = std::max(m_tractioncontrol, ar_wheels[i].wh_tc_coef < 0.9);
-            }
+            float tc_coef = pow(ar_wheels[i].wh_tc_coef, std::min(std::max(0.2f, curspeed), 1.0f));
+            total_torque *= pow(tc_coef, 1.0f / std::max(1.0f, std::min(wheel_slip, 10.0f)));
+            m_tractioncontrol = std::max(m_tractioncontrol, ar_wheels[i].wh_tc_coef < 0.9);
```
only-a-ptr: What is `m_tractioncontrol`? My impression is that it means "is TC currently in effect?", but I'm not sure. Please comment it.

`tc_mode`:  Please change  to bool (if it's used as such) or fix the comment in Beam.h

source/main/physics/BeamForcesEuler.cpp
```diff
@@ -418,11 +418,9 @@ void Actor::calcForcesEulerCompute(int step, int num_steps)
                 ar_wheels[i].wh_tc_coef = curspeed / fabs(ar_wheels[i].wh_speed);
                 ar_wheels[i].wh_tc_coef = pow(ar_wheels[i].wh_tc_coef, tc_ratio);
             }
-            if (!ar_engine || ar_engine->GetAcceleration() > ar_engine->getAccToHoldRPM(ar_engine->GetEngineRpm()))
-            {
-                total_torque *= pow(ar_wheels[i].wh_tc_coef, 1.0f / std::max(1.0f, std::min(wheel_slip, 10.0f)));
-                m_tractioncontrol = std::max(m_tractioncontrol, ar_wheels[i].wh_tc_coef < 0.9);
-            }
+            float tc_coef = pow(ar_wheels[i].wh_tc_coef, std::min(std::max(0.2f, curspeed), 1.0f));
```
only-a-ptr: just an idea: introduce helper 
```
namespace RoR {
template<typename T>
inline T Clamp(T min, T val, T max) { return std::max(min, std::min(val, max)); }
} // namespace RoR
```

ulteq: `Ogre::Math::Clamp` :man_facepalming:

source/main/physics/BeamForcesEuler.cpp
ulteq: -> https://github.com/RigsOfRods/rigs-of-rods/pull/1589

PR #1598, opened 2018-10-02
Simplified threading logic

ulteq:
The physics worker thread is now handled by the thread pool:
- Gets rid of the dummy thread pool which was used as a background worker thread
- Makes the `DisableThreadPool` setting obsolete
- Renamed `NumThreadsInThreadPool` into `NumWorkerThreads`
- Introduces `GVarPod_APS<int> app_num_workers`
- Converts `GVarPod_A<bool> app_multithread` into `GVarPod_APS<bool> app_async_physics`

- [x] Decide if we should rename `RoR::App::app_multithread` into `RoR::App::app_async_physics`

Reference: https://github.com/only-a-ptr/rigs-of-rods/pull/22

### DIFF
--- a/source/main/physics/BeamFactory.cpp
+++ b/source/main/physics/BeamFactory.cpp
@@ -96,34 +96,18 @@ ActorManager::ActorManager()
     , m_simulation_speed(1.0f)
     , m_actor_counter(0)
 {
+    // Create thread pool
+    int logical_cores = std::thread::hardware_concurrency();
+    LOG("BEAMFACTORY: " + TOSTRING(logical_cores) + " logical CPU cores" + " found");
 
-    if (RoR::App::app_multithread.GetActive())
+    int thread_pool_workers = ISETTING("NumThreadsInThreadPool", 0);
+    if (thread_pool_workers < 1 || thread_pool_workers > logical_cores)
     {
-        // Create worker thread (used for physics calculations)
-        m_sim_thread_pool = std::unique_ptr<ThreadPool>(new ThreadPool(1));
-
-        // Create general-purpose thread pool
-        if (!BSETTING("DisableThreadPool", false))
-        {
-            int logical_cores = std::thread::hardware_concurrency();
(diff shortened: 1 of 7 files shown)


only-a-ptr:
> Decide if we should rename RoR::App::app_multithread into RoR::App::app_async_physics

How about `GVarPod_APS<int> app_num_workers` - when 0, auto-detected and updated (left as 0 in RoR.cfg).

ulteq:
That variable would then be used for changing the amount of workers on the fly?

ulteq:
I should have used `GVarPod_APS` as you suggested. :man_facepalming:

ulteq:
Ready, from my point of view.

ulteq:
Ready, from my point of view.

ulteq:
@only-a-ptr anything you want to add?

source/main/physics/BeamFactory.cpp
```diff
@@ -1023,18 +1008,14 @@ void ActorManager::UpdateActors(Actor* player_actor, float dt)
...
-    {
-        this->UpdatePhysicsSimulation();
-    }
+    auto func = std::function<void()>([this]()
+        {
+            this->UpdatePhysicsSimulation();
+        });
+    m_sim_task = gEnv->threadPool->RunTask(func);
+
+    if (!RoR::App::app_async_physics.GetActive())
```
only-a-ptr: This reminds me of the "AsynchronousPhysics" setting in 0.4.0.7 which I removed as broken by design. Unless I'm missing something, it should cause a race condition under heavier physics workload. `join()`ing the sim. tasks here guards you against overlapping physics updates - without it, you may start another round of updates before the previous one finishes, eventually causing the same actor (nodes, beams, engine...) to be modified by 2 threads, the current step and the previous step. Correct me if I'm wrong.

only-a-ptr:
I didn't find the time to test, so I just skimmed over the code. I'm confident it will run just fine.

only-a-ptr:
I wanted to have a thorough look on this, but finally I didn't find the time, so I just skimmed over the code now. I'm confident it will run just fine.

PR #1599, opened 2018-10-02
Fixed 'Unimog U5000 Dakar' segfaults

ulteq:
I'm unable to find a better solution. But this seems to work fine.

Truck file: http://archives.rigsofrods.net/repo/files/repofiles-1st-batch/58dcfc39b7024d3eacb3d70d8dca3bc9709ed7ae_UnimogU5000Dakar.zip

Underlying issue:
```
error loading mesh: dashboard1.mesh: OGRE EXCEPTION(2:InvalidParametersException):
Couldn't read 16 bit header value from input stream.
```

### DIFF
--- a/source/main/gfx/GfxActor.cpp
+++ b/source/main/gfx/GfxActor.cpp
@@ -2535,9 +2535,9 @@ void RoR::GfxActor::SetCastShadows(bool value)
     // Props
     for (prop_t& prop: m_props)
     {
-        if (prop.mo != nullptr)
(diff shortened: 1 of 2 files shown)


ulteq:
> kill the prop altogether

The question is, do we lose truck features that would otherwise not be lost if we do it that way? I can't answer that by just looking at the code.

ulteq:
> Then delete the code and look at the rendered result 

I would have had to delete the entire `ProcessProp()` method.

PR #1601, opened 2018-10-03
Fixed survey / teleport map

ulteq:
Reason for the bug: `mMapTextureCreator->update()` was called too early.
- Removed obsolete methods
- Removed `mMapTextureNeedsUpdate` logic


only-a-ptr:
I never understood why the overview map was implemented as top-down render; icons would have been much more readable.

PR #1605, opened 2018-10-04
Interactive truck reset fix

ulteq:
- Fixes position/rotation creep during interactive truck reset
- You no longer get stuck in the permanent interactive truck reset mode unintended


only-a-ptr:
:+1:

PR #1611, opened 2018-10-05
New commandline parameters: '-pos', '-rot'

ulteq:
New command line parameters which allow you to override the default spawn position and rotation.

Example: `./RoR -map 31 -pos "619 10 580" -rot 90`


only-a-ptr:
> New commandline parameters: '-pos', '-rot' 

:+1: 

> Skip the camera animation on spawn

What camera animation? //Hack comment is a good start, but it should say what exactly it tries to work around. I think the whole commit should be replaced by `m_camera_manager->Reset()` which clears the movement cache of camera, making it jump to the current target immediately. Test it pls.

> Make '-truck' and '-enter' transient 

Not the way to do it. There's GVar type 'APS' for this, read comments in Application.h

ulteq:
I tried all of those:
```
m_camera_manager.CameraBehaviorOrbitReset();
m_camera_manager.NotifyContextChange();
this->ResetCamera();
```
And none of them skips the camera animation after the terrain loading is finished.

PR #1621, opened 2018-10-07
Fixed the prev player logic

ulteq:
The old 'm_prev_player_actor' update logic broke when the 'm_actor_remove_queue' was introduced


only-a-ptr:
Clever.

PR #1630, opened 2018-10-10
Hotfix: Terrain gravity definition (terrn2)

only-a-ptr:
The simplest mistake - value from terrn2 file was never propagated to live terrain 😊


tritonas00:
yep, it works fine

PR #1631, opened 2018-10-11
Improved 'Stall RPM' option parsing

ulteq:
Fixes trucks like the `international-4400-single-axle-rollback.truck` which define a stall rpm higher than the idle rpm.

### DIFF
--- a/source/main/gameplay/BeamEngine.cpp
+++ b/source/main/gameplay/BeamEngine.cpp
@@ -211,7 +211,7 @@ void EngineSim::SetEngineOptions(float einertia, char etype, float eclutch, floa
     m_post_shift_time = std::max(0.0f, m_post_shift_time);
     m_clutch_time = Math::Clamp(m_clutch_time, 0.0f, 0.9f * m_shift_time);
 
-    m_engine_stall_rpm = Math::Clamp(m_engine_stall_rpm, 0.0f, m_idle_rpm);
+    m_engine_stall_rpm = Math::Clamp(m_engine_stall_rpm, 0.0f, 0.9f * m_idle_rpm);
 
     if (etype == 'c')
     {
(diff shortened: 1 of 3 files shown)


source/main/gameplay/BeamEngine.cpp
```diff
@@ -211,7 +211,7 @@ void EngineSim::SetEngineOptions(float einertia, char etype, float eclutch, floa
     m_post_shift_time = std::max(0.0f, m_post_shift_time);
     m_clutch_time = Math::Clamp(m_clutch_time, 0.0f, 0.9f * m_shift_time);
 
-    m_engine_stall_rpm = Math::Clamp(m_engine_stall_rpm, 0.0f, m_idle_rpm);
+    m_engine_stall_rpm = Math::Clamp(m_engine_stall_rpm, 0.0f, 0.9f * m_idle_rpm);
```
only-a-ptr: Doesn't this just mask the problem for cases where Stall-RPM is only slightly higher than Idle-RPM?

ulteq: `m_engine_stall_rpm` >= `m_idle_rpm` will always break the current simulation.

And it doesn't make sense either. We might as well pick 80% of the idle rpm.

only-a-ptr: Right. Stall RPM should ideally be defined as percentage of idle RPM.

In the current state, if it's ill-defined, we may just revert to some sane default with a warning in log. Either way your workaround titled "fix" is ugly, improve it. 😄

ulteq: `m_engine_stall_rpm >= m_idle_rpm` basically makes you unable to drive the truck. Which I would definitely consider a bug. That's why this commit is fixing a bug. Older versions (pre 0.4.0.7) did not discern between idle and stall rpm.

> we may just revert to some sane default with a warning in log

We could choose 80% of the idle rpm (which is reasonable) and output a warning in the log.

only-a-ptr: > 90% of the idle rpm is a reasonable default value.

You're right, I misunderstood the code.

ulteq:
> improve it

I'll try to detect obvious error in the `ActorSpawner::ProcessEngoption` and output an appropriate warning.

ulteq:
Done: https://github.com/RigsOfRods/rigs-of-rods/pull/1631/commits/b7ea07695ea594a31e567336bd9ffea75df41384

Issue #1623, opened 2018-10-07
improve engine startup simulation


Speciesx:
https://youtu.be/9KdzWy6QrlE

Ignition is on, first gear is engaged, manual clutch released and the engine has reached min rpm to startup. The engine should start, but it won't until 'S' (STARTER) is pressed.

PR #1637, opened 2018-10-12
Advanced engine startup

ulteq:
- More natural rev up
- Allows to pull start the engine

Fixes: https://github.com/RigsOfRods/rigs-of-rods/issues/1623


Speciesx:
>Allows to pull start the engine

Works only with forwards gears

only-a-ptr:
:+1: Perfect.

PR #1639, opened 2018-10-12
Much improved mesh collision performance

ulteq:
Performance was literally lying on the ground waiting to be picked up. :man_facepalming:

This simply hides the costly matrix multiplication behind a bounding box check.

| Before | After |
| ------------- | ------------- |
| before | after |


only-a-ptr:
:100:

PR #1640, opened 2018-10-13
Fixed the terrain initialization order

ulteq:
Both the collision and the script subsystem need to be initialized after the terrain geometry is loaded.

This fixes the segfault on the CPU Benchmark terrain.


only-a-ptr:
:+1:

PR #1641, opened 2018-10-14
Improved mesh collision lookup performance

ulteq:
Decent performance gain by skipping entire cells based on the maximum height of their collision tris/boxes.


only-a-ptr:
:+1: awesome.

PR #1642, opened 2018-10-14
Fixed collision debug visualization

ulteq:
The debug visualization should only be visible if the cell_id matches.


only-a-ptr:
:+1: I didn't believe that thing even worked :smile:

ulteq:
It worked fine. :smile:

PR #1643, opened 2018-10-14
Increased node vs. ground collision accuracy (10x)

ulteq:
- Fixes ground collision issues of nodes which are not wheel nodes
- Makes the 'HighResWheelNodeCollisions' setting obsolete

Goodbye 'collTestTimer'!


only-a-ptr:
> Goodbye 'collTestTimer'!

We'll will so not miss you :grin:

Issue #1627, opened 2018-10-09
leaving trains on load spawner causing performance drops


Speciesx:
https://youtu.be/Hhvves64D5Y
https://youtu.be/BDFebwbMzO8

Terrain
https://forum.rigsofrods.org/downloads.php?do=file&id=109
Trains
https://forum.rigsofrods.org/downloads.php?do=file&id=189
https://forum.rigsofrods.org/downloads.php?do=file&id=188

only-a-ptr:
This is most definitely caused by scripting, specifically firing the "character is on spawner!" event. There's already a good deal of hackery and sorry comments in the code related to this.

Actually I'm a bit surprised you don't end up in an endless loop of displaying the Selector GUI. That's how I ended up getting out from Agora bus on Load spawner in Auriga.

vido89:
Im not sure that we should stick "hackey" code around and support it beyond some point in time.
Second thing: I did sitting animation along with other animations should we have more that one animation in sitting pose since we have truck, train, car and airplane ?

only-a-ptr:
@vido89 > Im not sure that we should stick "hackey" code around

I'm already scheming to get rid of it: https://github.com/RigsOfRods/rigs-of-rods/issues/1558

@vido89 > Second thing

Offtopic, start an issue ticket or thread in Dev forum

ulteq:
> This is most definitely caused by scripting

Yes it is caused by this logic: https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/physics/collision/Collisions.cpp#L788-L793

Which fails on overlapping collision boxes.

only-a-ptr:
@ulteq Exactly. We need to use something like `std::unordered_set<int> m_occupied_collision_boxes`

PR #1645, opened 2018-10-14
Fixed FPS drop on train spawners

ulteq:
Shamelessly based on the idea of @only-a-ptr:
> Exactly. We need to use something like std::unordered_set<collision_box_t*> m_occupied_collision_boxes

Fixes: https://github.com/RigsOfRods/rigs-of-rods/issues/1627

### DIFF
--- a/source/main/physics/collision/Collisions.cpp
+++ b/source/main/physics/collision/Collisions.cpp
@@ -118,7 +118,6 @@ Collisions::Collisions():
     , free_eventsource(0)
     , hashmask(0)
     , landuse(0)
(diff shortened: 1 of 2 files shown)


only-a-ptr:
I didn't test, but I think this will only work right if the spawnboxes overlap perfectly, because IIRC exiting either will invoke `clearEventCache()` and erase info about the other. Theoretically this should duplicate the "spawnerbox entered" event, but I'm not exactly sure about these.

Thanks for fixing anyway.

ulteq:
It fixes the performance issues without introducing new issues.

PR #1647, opened 2018-10-14
Added a 'remove all vehicles' button

ulteq:


ulteq:
I would not have done it, but meh, why not ...

PR #1648, opened 2018-10-14
An attempt to improve the sleeping logic

ulteq:
Does this fix any of the current sleeping logic issue?
- Attached items seem to sometimes fall off when things go sleeping?
- Items seem to fall through trailers when things go sleeping?


only-a-ptr:
What sleeping logic issues? Do we have issue tickets or anything written?

Speaking of improving the sleep logic, I suggest remaking it to count physics steps instead of seconds

ulteq:
> I suggest remaking it to count physics steps instead of seconds

Sounds good.

ulteq:
The sleeping issue seems to be gone already. I'll create a new PR for the suggested changes.

PR #1649, opened 2018-10-15
Classic case of premature optimization

ulteq:
This is really not worth the extra cluttering

### DIFF
--- a/source/main/physics/BeamFactory.cpp
+++ b/source/main/physics/BeamFactory.cpp
@@ -1046,15 +1046,12 @@ void ActorManager::UpdatePhysicsSimulation()
     }
     for (int i = 0; i < m_physics_steps; i++)
     {
-        bool have_actors_to_simulate = false;
-
         {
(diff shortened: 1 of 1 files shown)


only-a-ptr:
👍 for the code, I hate the title tho - it's actually just duplicate logic, a loop executed 0x is the same thing as branch not taken, in both principle and CPU instructions. The author was only thinking he optimizes something.

ulteq:
> a loop executed 0x is the same thing as branch not taken

However, the original code should run faster, at least theoretically.

only-a-ptr:
> However, the original code should run faster, at least theoretically.

after looking at the code for 2nd time, I agree, it is an optimization attempt which just isn't worth it.

I think the `ar_update_physics` ( and by extension `ar_sim_state`  and  `m_ongoing_reset` ) is sub-optimal design to begin with. Since switches between states are not frequent, I'd prefer to have a separate  `std::vector<> m_locally_simulated_actors` and  `std::vector<> m_network_controlled_actors`

PR #1651, opened 2018-10-15
Removed unnecessary helper method

ulteq:


ulteq:
I'll think about it.

ulteq:
I don't think that it makes sense to update it more often.

Issue #1348, opened 2017-07-06
Weird Vehicle Behavior. (See Attached Pastebin)


disloyalpick:
https://pastebin.com/3Z29zciL
This is probably the only way to fully undestand what I'm trying to get at with this issue.

It was basically found out that the hydros are causing weird strain.

How to reproduce. Get one of gabester's vehicles. 
Then, start moving slightly. 
Then, turn left or right. Which ever way will make the vehicle start rolling by itself with absolutely no throttle input. 
Then shut the engine off and switch to neutral.
the vehicle will continue to keep rolling.

@L-Tec has been helping me pinpoint the exact issue recently. And Lifter has come to the conclusion that is is definitely hydro related.

If you need any more info like a video of it happening, let me know

Michael10055:
Pasting this message here from me on Discord:

> the issue is really noticeable on this vehicle https://forum.rigsofrods.org/thread-102.html which uses a very simple steering setup. it does not happen on 0.3x:
https://gfycat.com/WiltedCheapBarasinga
0.4.7.0+
https://gfycat.com/BraveAchingDrongo

only-a-ptr:
I just reproduced it using vehicle https://forum.rigsofrods.org/thread-102.html - it happens when I steer full left while the vehicle had some prior motion.

Weird thing.

disloyalpick:
@only-a-ptr Some time ago Lifter helped to tace it down to the hydros causing it to happen. Try sending him a DM about it. He may have a bit more info from his testings.

only-a-ptr:
May be related: https://github.com/RigsOfRods/rigs-of-rods/pull/1652

tritonas00:
this is probably fixed now, mike and i can't reproduce it anymore

PR #1652, opened 2018-10-15
Brought seemingly necessary hack back

ulteq:
Prevents the Gavril Omega Six from sliding on ground on the simple2 spawn.

Reference: https://github.com/RigsOfRods/rigs-of-rods/commit/d11a88142f737528638bd357c38d717c85cebba6#diff-4003254e55aec2c60d21228f375f2a2dL1153

Please don't ask me what this actually does.


ulteq:
> Might be related: #1348

I doubt that they are related.

> Idea: choose a new origin every frame

But this is an awesome idea and will potentially fix the issue.

PR #1654, opened 2018-10-15
Proper submesh friction for inter vehicle collisions

ulteq:
You can get the build here: http://prdownloads.sourceforge.net/rigs-of-rods/RoR-0.4.8.0-truck-truck-collision-pof.zip

For now this is just a proof of concept. The physics are not correct.

### DIFF
--- a/source/main/physics/Beam.h
+++ b/source/main/physics/Beam.h
@@ -474,6 +474,7 @@ private:
     bool              m_antilockbrake;         //!< GUI state
     bool              m_tractioncontrol;       //!< GUI state
     bool              m_ongoing_reset;         //!< Hack to prevent position/rotation creep during interactive truck reset
+    bool              m_first_physics_step;    //!< Hack to prevent submesh destruction
 
     bool m_hud_features_ok:1;      //!< Gfx state; Are HUD features matching actor's capabilities?
     bool m_slidenodes_locked:1;    //!< Physics state; Are SlideNodes locked?
--- a/source/main/physics/BeamForcesEuler.cpp
+++ b/source/main/physics/BeamForcesEuler.cpp
@@ -34,9 +34,11 @@
 #include "CmdKeyInertia.h"
 #include "Collisions.h"
 #include "Differentials.h"
+#include "DynamicCollisions.h"
 #include "DustPool.h"
 #include "FlexAirfoil.h"
 #include "InputEngine.h"
+#include "PointColDetector.h"
 #include "Replay.h"
 #include "RoRFrameListener.h"
 #include "ScrewProp.h"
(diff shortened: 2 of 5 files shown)


source/main/physics/Beam.h
```diff
@@ -474,6 +474,7 @@ class Actor : public ZeroedMemoryAllocator
     bool              m_antilockbrake;         //!< GUI state
     bool              m_tractioncontrol;       //!< GUI state
     bool              m_ongoing_reset;         //!< Hack to prevent position/rotation creep during interactive truck reset
+    bool              m_first_physics_step;    //!< Hack to prevent submesh destruction
```
only-a-ptr: Good comment, but what exactly is at risk of destruction?

source/main/physics/BeamForcesEuler.cpp
```diff
@@ -55,14 +57,34 @@ void Actor::calcForcesEulerCompute(int step, int num_steps)
     IWater* water = App::GetSimTerrain()->getWater();
     const bool is_player_actor = (this == RoR::App::GetSimController()->GetPlayerActor());
 
+    if (!m_first_physics_step)
```
only-a-ptr: This would be a good place to comment why we need this guard. I have no clue.

source/main/physics/collision/DynamicCollisions.cpp
```diff
@@ -136,6 +136,47 @@ void ResolveCollisionForces(const float penetration_depth,
     no.Forces -= gamma * forcevec;
 }
 
+/// Calculate collision forces and apply them to the collision node and the three vertex nodes of the collision triangle.
+void ResolveInterCollisionForces(const float penetration_depth,
```
only-a-ptr: I don't like where this is going. After `calcBeamsInterTruck()` this is another fragmentation of logic between "intra-actor" and "inter-actor". But by principle, those should be perfectly the same thing - so by extension, a well designed code will only do it in one place. RoR separates it for historical reasons (rollon setting - convenient but a practical hack) and also threading in it's current design (process actors separately, then together), but IMO we should advance towards "process nodes[all actors], beams [all actors], collisions[all actors]" design.

ulteq:
The motivation is to make submesh friction work as one would expect: http://docs.rigsofrods.org/vehicle-creation/fileformat-truck/#submesh_groundmodel

There is a completely lack of static friction between contacters and cabs on master.
The topic is complicated and can't be explained in two sentences.
I don't plan on merging this pull request soon. It needs more testing and more work.
Treat it as a proof of concept for now.

This is how submesh friction behaves on master: https://streamable.com/hvdc4
This is how submesh friction behaves with with pull request: https://streamable.com/lqhh7

source/main/physics/collision/DynamicCollisions.cpp
ulteq: > this is another fragmentation of logic between "intra-actor" and "inter-actor"

I tried hard to prevent further fragmentation, but I don't think it is possible to fix submesh friction while maintaining backwards compatibility without fragmentation. I've already spent many hours to make this happen without any performance penalty and full backwards compatibility.

source/main/physics/Beam.h
ulteq: The inter-vehicle-collision code needs to run after `CalcBeams` and before `CalcNodes`. In order to achieve that without a massive performance penalty I had to reorder the execution of them, which breaks the intra-vehicle-collision code (for some vehicles) without this hack.

only-a-ptr:
No problem at all. I know this old submesh friction bug and I thought you were dealing with exactly that, I was just nervous about the sparingly commented PR. You've been doing a lot of small PRs lately, this seemed like one in a row maybe.

I'm going to be travelling for the next ~3 weeks, so you'll have a lot of space to re-arrange things. Feel free to start your own `dev-frenzy` branch 😉

ulteq:
> so you'll have a lot of space to re-arrange things

I was hoping to be done with the rearranging. I'm already growing grey hairs from this. :smile:

ulteq:
> I see :sparkles: but I don't see the improvement

The community is going to love this one.

only-a-ptr:
Community is going to be thrilled no doubt, I wish you good luck. But grey hair from re-arranging too much? I grow them from not re-arranging enough, all my accomplishments were based on redesigning and reordering things 😄

ulteq:
> But grey hair from re-arranging too much?

Figuring out why the old code did not work took a lot of re-arranging.

ulteq:
> all my accomplishments were based on redesigning and reordering things

No problem. I knew that I won't win a design award for this. :smile:

source/main/physics/collision/DynamicCollisions.cpp
ulteq: I will try my best to unify those two methods once we know that it really works as intended.

PR #1657, opened 2018-10-16
Improved the hide gui logic

ulteq:
It's now working somewhat like you would expect and fixes the dashboard visibility logic,
which broke here: https://github.com/RigsOfRods/rigs-of-rods/commit/1b50c147a73be01c316d7df237f567207789d77e#diff-8a886433b2467e133b491db2b4ec51ffL1761

### DIFF
--- a/source/main/gameplay/RoRFrameListener.cpp
+++ b/source/main/gameplay/RoRFrameListener.cpp
@@ -1834,22 +1834,15 @@ void SimController::HideGUI(bool hidden)
     }
 #endif // USE_SOCKETW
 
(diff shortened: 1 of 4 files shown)


only-a-ptr:
👍 thanks

ulteq:
We might want to introduce a dedicated button for the vehicle dashboard in the future. That would both simplify the code a lot and improve the improve the user experience.

only-a-ptr:
> @ulteq  We might want to introduce a dedicated button for the vehicle dashboard in the future. That would both simplify the code a lot and improve the improve the user experience.

Agreed

PR #1658, opened 2018-10-16
Fixed nasty bug in the idle mixture calculation

ulteq:
The bug was hidden by the previous idle mixture logic (https://github.com/RigsOfRods/rigs-of-rods/commit/912be9ec1ecce946707db36232b944cad6408831)

This indirectly fixes the idle mixture calculation, which results in a correct idle rpm, which fixes the broken engine triggers. It also makes the rpm hold feature behave better.

In addition it prevents bad interaction between manual inputs and engine trigger inputs.


only-a-ptr:
:+1:

PR #1660, opened 2018-10-17
🚚 Replaced AppImage with portable build

AnotherFoxGuy:
Test the new builds here: https://anotherfoxguy.itch.io/rigs-of-rods
The password is `RoR-Dev`

## TODO

- [x] Create redist folder on windows
- [x] Fix caelum entry in plugins.cfg


tritonas00:
after applying this on master, there is no caelum entry on plugins.cfg

PR #1690, opened 2018-10-28
Added default content from RigsOfRods/content

AnotherFoxGuy:
* Removed content from `resources/skeleton/packs/`
* `zip_and_copy_resources` will now fetch some default content from [RigsOfRods/content](https://github.com/RigsOfRods/content) and place it into `bin/content`
* Detached `zip_and_copy_resources` from the `RoR` target


only-a-ptr:
What motivated you to do this? Was there any discussion on webforum/Discord?

PR #1692, opened 2018-10-29
Separation of inter-wheel and inter-axle differentials

ulteq:
- Cleanly separates the existing differential code into inter-wheel and inter-axle differentials
- Makes the open diff code ready to be used for inter-axle differentials
- Simplifies the handling detached wheels in the differential code


only-a-ptr:
👍 perfect

PR #1697, opened 2018-10-30
Fixed possible nullptr dereference

ulteq:
Reported by: Coverity.


only-a-ptr:
I don't understand why Coverity reported it and why this change fixes it - in both the old and new code, we have the `gm != nullptr` and `gm->type` in one expression and we rely on left-to-right asociativity of operators `&&` and `||`: https://github.com/MicrosoftDocs/cpp-docs/blob/master/docs/cpp/cpp-built-in-operators-precedence-and-associativity.md

ulteq:
We called `n->nd_last_collision_slip.length();` before testing `n != nullptr`.

PR #1701, opened 2018-10-31
Fixed debug view desync in pause menu

ulteq:


only-a-ptr:
ouch, I was kind of hoping we can avoid exactly this sort of hacks. But it's OK for now I guess

ulteq:
I don't see any other way of solving the issue, unless we want to build the debug view on top of the sim buffer.

only-a-ptr:
> ...unless we want to build the debug view on top of the sim buffer.

We indeed want that, it's  _visual_ -ization, that's the principle 😄 But this workaround is still OK for now.

PR #1715, opened 2018-11-02
Contacter cleanup

ulteq:


ulteq:
I can't think of any extra info that a contacter node could have over a regular node.

PR #1716, opened 2018-11-03
Fixed overview map glitches

ulteq:
Fixes:
- View obstruction by fog
- Water flickering when the map is opened
- Ugly water rendering when using the speed optimized water rendering option


only-a-ptr:
Mixed, but whatever 👍

PR #1734, opened 2018-11-05
Fixed submesh-collision behavior of slidenodes

ulteq:
The slide-node forces need to be updated after the intra-truck collisions are handled and before the inter-truck collisions are handled.

### DIFF
--- a/source/main/physics/Beam.cpp
+++ b/source/main/physics/Beam.cpp
@@ -2261,6 +2261,23 @@ void Actor::CalcAnimators(const int flag_state, float& cstate, int& div, Real ti
     }
 }
 
+void Actor::CalcContacters()
+{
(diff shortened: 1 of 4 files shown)


only-a-ptr:
Those `// must do before|after` comments are multiplying 😟 

> Updating them before the intra-truck collisions are handled would break tracked vehicles.
I'd prefer having comments like these directly in the code, so we can eventually analyze and straighten the update cycle. Living with all those 'before/after' rules would be hell.

ulteq:
> Those // must do before|after comments are multiplying

I made them least restrict as possible. Ignoring them will definitely break things.

> Could you concentrate them in `CalcForcesEulerCompute()`

I guess you are right. We should clarify why those restrictions are necessary via comments.

Issue #816, opened 2016-03-30
Future: proper multithreaded architecture


only-a-ptr:
This is a tech proposal document.

At the moment, RoR's multithreading is flawed. Read description of milestone (future-proper-mt)[https://github.com/RigsOfRods/rigs-of-rods/milestones] for details and general design goals.

## General design goals (recap)
* Main thread should exclusively manage OGRE objects and do rendering and visual-only tasks. It should never affect simulation. Timing - dynamic (FPS)
* Sim thread(s) should never touch OGRE and run on fixed 2khz timestep.

## Overview and TODO list
<table>
<thead><tr><th>State</th><th>Area</th><th>Current</th><th>Desired</th></tr></thead>
<tbody>
<tr>
    <td>OK</td>
    <td>Collisions - static objects</td>
    <td>On map load, a physics-only collision triangles are created</td>
    <td>~</td>
</tr>
<tr>
    <td>TODO</td>
    <td>Collisions - heightmap</td>
    <td>Direct ray-queires to OGRE terrain</td>
    <td>Separate collision-only heightmap (+bump map?)</td>
</tr>
<tr>
    <td>TODO</td>
    <td>Sleeping vehicle - engine</td>
    <td>Updated on main thread</td>
    <td>Updated on sim thread</td>
</tr>
<tr>
    <td>OK</td>
    <td>Active vehicle - engine</td>
    <td>Updated on sim thread</td>
    <td>~</td>
</tr>
<tr>
    <td>TODO</td>
    <td>Active vehicle - (auto)gearbox</td>
    <td>Updated on main thread (sync?!)</td>
    <td>Updated on sim thread</td>
</tr>
<tr>
    <td>TODO</td>
    <td>Active vehicle - brakes</td>
    <td>Updated on main thread (sync?!)</td>
    <td>Updated on sim thread</td>
</tr>
<tr>
    <td>TODO</td>
    <td>Active vehicle - TC/ALB</td>
    <td>Updated on main thread (sync?!)</td>
    <td>Updated on sim thread</td>
</tr>

</tbody>
</table>

only-a-ptr:
Related workarounds and rants: https://github.com/RigsOfRods/rigs-of-rods/pull/1752/files/080ed5e49774c03337b63c59303be150b520773f#diff-11319c6cae8caeca2be4dde205bd243b

ulteq:
> Triggers (IIRC main thread, should be sim)

[...]

PR #1752, opened 2018-11-11
Various threading fixes

ulteq:
Fixes: https://github.com/RigsOfRods/rigs-of-rods/pull/1598
Replaces: https://github.com/RigsOfRods/rigs-of-rods/pull/1747

Prevents deadlocks and segfaults.

### DIFF
--- a/source/main/gfx/GfxActor.cpp
+++ b/source/main/gfx/GfxActor.cpp
@@ -114,6 +114,9 @@ RoR::GfxActor::GfxActor(Actor* actor, std::string ogre_resource_group,
 
 RoR::GfxActor::~GfxActor()
 {
+    this->FinishWheelUpdates();
+    this->FinishFlexbodyTasks();
+
     // Dispose videocameras
     this->SetVideoCamState(VideoCamState::VCSTATE_DISABLED);
     while (!m_videocameras.empty())
--- a/source/main/physics/BeamFactory.cpp
+++ b/source/main/physics/BeamFactory.cpp
@@ -95,14 +95,17 @@ ActorManager::ActorManager()
     , m_simulation_speed(1.0f)
     , m_actor_counter(0)
 {
-    // Create thread pool
+    // Create worker thread (used for physics calculations)
+    m_sim_thread_pool = std::unique_ptr<ThreadPool>(new ThreadPool(1));
+
+    // Create general-purpose thread pool
     int logical_cores = std::thread::hardware_concurrency();
     LOG("BEAMFACTORY: " + TOSTRING(logical_cores) + " logical CPU cores" + " found");
 
     int thread_pool_workers = RoR::App::app_num_workers.GetActive();
-    if (thread_pool_workers < 2 || thread_pool_workers > logical_cores)
+    if (thread_pool_workers < 1 || thread_pool_workers > logical_cores)
     {
-        thread_pool_workers = std::max(2, logical_cores - 1);
+        thread_pool_workers = logical_cores - 1;
         RoR::App::app_num_workers.SetActive(thread_pool_workers);
     }
 
@@ -1051,7 +1054,7 @@ void ActorManager::UpdateActors(Actor* player_actor, float dt)
         {
             this->UpdatePhysicsSimulation();
         });
-    m_sim_task = gEnv->threadPool->RunTask(func);
+    m_sim_task = m_sim_thread_pool->RunTask(func);
 
     if (!RoR::App::app_async_physics.GetActive())
         m_sim_task->join();
--- a/source/main/physics/BeamForcesEuler.cpp
+++ b/source/main/physics/BeamForcesEuler.cpp
@@ -472,6 +472,7 @@ void Actor::CalcWheels(bool doUpdate, int num_steps)
         ar_engine->SetWheelSpin(ar_wheel_spin * RAD_PER_SEC_TO_RPM); // Update the driveshaft speed
     }
 
+#if 0
     if (doUpdate)
     {
         if (!m_antilockbrake)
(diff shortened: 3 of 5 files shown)


source/main/physics/BeamForcesEuler.cpp
```diff
@@ -472,6 +472,7 @@ void Actor::CalcWheels(bool doUpdate, int num_steps)
         ar_engine->SetWheelSpin(ar_wheel_spin * RAD_PER_SEC_TO_RPM); // Update the driveshaft speed
     }
 
+#if 0
```
only-a-ptr: NEEDINFO - what is the issue here?

source/main/gfx/GfxActor.cpp
```diff
@@ -114,6 +114,9 @@ RoR::GfxActor::GfxActor(Actor* actor, std::string ogre_resource_group,
 
 RoR::GfxActor::~GfxActor()
 {
+    this->FinishWheelUpdates();
+    this->FinishFlexbodyTasks();
```
only-a-ptr: TDev-style workarounds ☹️ We have to schedule the disposal the same way we schedule spawn.

source/main/physics/BeamFactory.cpp
```diff
@@ -95,14 +95,17 @@ ActorManager::ActorManager()
     , m_simulation_speed(1.0f)
     , m_actor_counter(0)
 {
-    // Create thread pool
+    // Create worker thread (used for physics calculations)
+    m_sim_thread_pool = std::unique_ptr<ThreadPool>(new ThreadPool(1));
```
only-a-ptr: ?? I'm guessing you removed the old `ThreadPool(1)` in some recent commit and now you're re-introducing it, otherwise I don't understand this.

ulteq: We removed it a few months ago during the major multi-threading cleanup: https://github.com/RigsOfRods/rigs-of-rods/pull/1598

But it turned out that it can deadlock (as I always suspected): https://github.com/only-a-ptr/rigs-of-rods/pull/22#issuecomment-423998730

See: https://github.com/ulteq/rigs-of-rods/commit/256b8813d3c2f21ad4b2880addae4de36a2617d1#diff-e11c82258dd7c1910ddc4dbc51cafa60

ulteq: `m_sim_thread_pool` is just a lazy way to launch our physics thread in the background.

The alternative would be something like `std::async`.

source/main/gfx/GfxActor.cpp
ulteq: Like this?
```
for (auto t: m_flexwheel_tasks) { t->join(); }
for (auto t: m_flexbody_tasks) { t->join(); }
```

source/main/physics/BeamForcesEuler.cpp
ulteq: It segfaults in some cases. I guess I should have taken the time to move it into the GfxActor?

source/main/gfx/GfxActor.cpp
only-a-ptr: No, I mean the tasks should be started and stopped [only in a single location](https://github.com/RigsOfRods/rigs-of-rods/blob/742bff1ce076d02fa617b1be274ff6c20d6f699a/source/main/gfx/DustManager.cpp#L228-L233) and GfxActor should be destroyed while they're _definitely_ not running. Otherwise the design is flawed.

source/main/physics/BeamFactory.cpp
only-a-ptr: @ulteq Please visit https://github.com/RigsOfRods/rigs-of-rods/issues/816 and explain why it should deadlock the way you describe. I can't think of any reason, other than flawed architecture. I know at the beginning (v0.4.0x) you had to work your way thru all the legacy logic, but I think now we can just do it right without compromises.

> https://github.com/ulteq/rigs-of-rods/commit/256b8813d3c2f21ad4b2880addae4de36a2617d1#diff-e11c82258dd7c1910ddc4dbc51cafa60

🤢

source/main/physics/BeamForcesEuler.cpp
only-a-ptr: > in some cases.

On shutdown or exiting to main menu I bet. Otherwise it wouldn't make sense. OpenAL by principle runs it's own thread, thus it's sound-update functions are lightweight and can be called anywhere anytime. It doesn't belong to GfxActor. If it segfaults, it's our bug.

ulteq: No, on truck activation.

ulteq: > can be called anywhere anytime

But we shouldn't do any sound updates from inside the physics loop in my opinion.

source/main/gfx/GfxActor.cpp
ulteq: I think I failed. I cannot reproduce the segfault anymore on latest master and I also don't understand how it could segfault there in the first place. I will revert the change.

source/main/physics/BeamFactory.cpp
ulteq: The single thread pool approach fails if there is just one worker, because there is no other thread to trigger the condition variable again: https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/threadpool/ThreadPool.h#L63, if there is already a waiting task (in our case the sim thread task, which is also waiting at this point).

The alternative solution is to not allow the creation of thread pools with less than 2 workers: https://github.com/RigsOfRods/rigs-of-rods/commit/256b8813d3c2f21ad4b2880addae4de36a2617d1

source/main/physics/BeamForcesEuler.cpp
ulteq: -> https://github.com/RigsOfRods/rigs-of-rods/commit/8e0c8ac3c487ec31d8d05afd4856aeaab7110f1b

PR #1743, opened 2018-11-09
✨ Introduced shocks3

ulteq:
Similar to `shocks2`, but the damping is now based off the compression / extension velocity.

Idea, design and truck file format by @Dohzer.

Example:
```
shocks3
;node1, node2, springin(N/m), dampin(N/m/s), slowdampin, splitin(m/s), fastdampin, springout(N/m), dampout(N/m/s), slowdampout, splitout(m/s), fastdampout, shortbound, longbound, precomp, options 
;F
40, 72, 23430, 1500, 0.5, 1.2, 0.2, 23430, 500, 0.5, 1.2, 0.2, 0.62, 0.00, 1.2, n
41, 73, 23430, 1500, 0.5, 1.2, 0.2, 23430, 500, 0.5, 1.2, 0.2, 0.62, 0.00, 1.2, n
;R
46, 55, 65000, 5800, 0.5, 1.2, 0.2, 65000, 1080, 0.5, 1.2, 0.2, 0.62, 0.00, 1.2, n
45, 54, 65000, 5800, 0.5, 1.2, 0.2, 65000, 1080, 0.5, 1.2, 0.2, 0.62, 0.00, 1.2, n
```

shocks3

PR #1755, opened 2018-11-12
Improved the shocks3 logic

ulteq:
- Fixed the shock velocity calculation
- Fixed the damping unit (needs to be N/(m/s))
- Added boundary limits similar to 'shocks1'
- Clamped the shock velocity to 0.1 m/s - 20 m/s.
- Simplified the extension / compression detection
- Improved the shocks debug view output

### DIFF
--- a/source/main/physics/Beam.cpp
+++ b/source/main/physics/Beam.cpp
@@ -2436,21 +2436,34 @@ void Actor::CalcShocks2(int i, Real difftoBeamL, Real& k, Real& d)
 
 void Actor::CalcShocks3(int i, Real difftoBeamL, Real &k, Real& d, Real v)
 {
-    if (ar_beams[i].shock->lastpos < difftoBeamL) // Extension
+    if (difftoBeamL > ar_beams[i].longbound * ar_beams[i].L)
     {
+        float interp_ratio =  difftoBeamL - ar_beams[i].longbound  * ar_beams[i].L;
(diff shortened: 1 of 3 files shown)


source/main/physics/Beam.cpp
```diff
@@ -2441,12 +2441,21 @@ void Actor::CalcShocks3(int i, Real difftoBeamL, Real &k, Real& d, Real v)
...
     }
     else // Compression
     {
         k = ar_beams[i].shock->springin;
         d = ar_beams[i].shock->dampin  * ar_beams[i].shock->dslowin  * std::min(v,        ar_beams[i].shock->splitin ) +
             ar_beams[i].shock->dampin  * ar_beams[i].shock->dfastin  * std::max(0.0f, v - ar_beams[i].shock->splitin );
+        d /= v;
+    }
+
+    if (difftoBeamL > ar_beams[i].longbound * ar_beams[i].L || difftoBeamL < -ar_beams[i].shortbound * ar_beams[i].L)
```
only-a-ptr: Maybe put this condition first and `return` early

only-a-ptr:
Oh, so we have `shocks3` now? I just had a look: https://github.com/RigsOfRods/rigs-of-rods/pull/1743#issuecomment-437811159

Looks good.

ulteq:
> Oh, so we have shocks3 now?

They're still in the making. The goal is to get them production ready with this pull request.

PR #1760, opened 2018-11-13
Keep trucks within the forwardcommands-range activated

ulteq:
This will automatically activate trucks which are listening to import commands as soon as a truck that is sending forward commands comes into range and keep them activated as long as that truck is nearby.


source/main/physics/BeamFactory.cpp
```diff
@@ -718,6 +718,12 @@ void ActorManager::ForwardCommands(Actor* source_actor)
                     (actor->getPosition().distance(source_actor->getPosition()) <
```
only-a-ptr: Nitpick: We should use `.squaredDistance()` here

ulteq:
> We should use `squaredDistance()` here

`ForwardCommands` is no longer part of the tight physics loop. See: https://github.com/RigsOfRods/rigs-of-rods/pull/1677

PR #1763, opened 2018-11-14
✨ Multithreaded sha1 calculation during cache regen

ulteq:
This cuts the cache regeneration time in half on my system.

The alternative would be to switch to something like [xxHash64](https://github.com/Cyan4973/xxHash), [MurmurHash3](https://github.com/aappleby/smhasher/wiki/MurmurHash3) or [CityHash](https://github.com/google/cityhash).

### DIFF
--- a/source/main/resources/CacheSystem.cpp
+++ b/source/main/resources/CacheSystem.cpp
@@ -42,6 +42,7 @@
 #include "SoundScriptManager.h"
 #include "TerrainManager.h"
 #include "Terrn2Fileformat.h"
+#include "ThreadPool.h"
 #include "Utils.h"
 
 #include "GUI_LoadingWindow.h"
@@ -2205,9 +2206,6 @@ void CacheSystem::loadSingleZip(Ogre::FileInfo f, bool unload, bool ownGroup)
 
 void CacheSystem::loadSingleDirectory(String dirname, String group, bool alreadyLoaded)
 {
-    char hash[256];
-    memset(hash, 0, 255);
-
     LOG("Adding directory " + dirname);
 
     rgcounter++;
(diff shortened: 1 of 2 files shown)


source/main/resources/CacheSystem.cpp
```diff
@@ -2331,6 +2334,35 @@ void CacheSystem::loadAllZipsInResourceGroup(String group)
...
+                FileInfo f = (FileInfo)*it;
+                String zippath = f.archive->getName() + "/" + f.filename;
+                String realzipPath = getRealPath(zippath);
+                String virtualPath = getVirtualPath(zippath);
+                char hash[256] = {};
+                RoR::CSHA1 sha1;
+                sha1.HashFile(const_cast<char*>(realzipPath.c_str()));
+                sha1.Final();
+                sha1.ReportHash(hash, RoR::CSHA1::REPORT_HEX_SHORT);
+                zipHashes[virtualPath] = hash;
```
only-a-ptr: This should be guarded by mutex.

ulteq: Right.

ulteq:
And to detect content mismatches in multiplayer, or not?

only-a-ptr:
> And to detect content mismatches in multiplayer, or not?

Umm, maybe, but then it would make sense to hash the actual truckfile/terrnfile when it's introduced instead of rehashing everything everytime.

> We already do that on incremental cache updates.

Somewhat, yes, but I wasn't very convinced by that code when I last looked at it.

ulteq:
Multiplayer mismatch detection seems to be entirely based on the filename: https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/physics/BeamFactory.cpp#L456 :thinking:

> but then it would make sense to hash the actual truckfile/terrnfile when it's introduced

That's how the hashing currently works. It's done once for each file.

PR #1768, opened 2018-11-15
ForwardDeclarations cleanup

ulteq:


source/main/ForwardDeclarations.h
```diff
@@ -182,7 +176,6 @@ class ProceduralManager;
 class RailSegment;
 class RailGroup;
 class Replay;
-class RigsOfRods;
```
only-a-ptr: I can't believe this was still around 😆

ulteq: Should we leave it in for some extra style points?

only-a-ptr: Nah, it's still just deadcode.

PR #1769, opened 2018-11-15
Optimized ComputeFlexbody execution order

ulteq:
This exploits the fact that not all flexbodies require the same amount of computation time.

The result is a decent FPS improvement (5-7%) on low core count (2-4) CPUs on trucks like the `Tatra-T813-Dakar`.

### DIFF
--- a/source/main/gfx/DustManager.cpp
+++ b/source/main/gfx/DustManager.cpp
@@ -75,8 +75,8 @@ void RoR::GfxScene::UpdateScene(float dt_sec)
     // Actors - start threaded tasks
     for (GfxActor* gfx_actor: m_live_gfx_actors)
     {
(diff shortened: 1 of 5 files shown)


only-a-ptr:
You're looking in a very promising direction - we're doing the same work (updating nodes, beams, flexbody vertices) in various-size chunks (N/B per actor, vertices per flexbody) and separately per actor. I think it would be much more machine-friendly if we did it in bulk, like all nodes of all active actors split into %NUM_CORES% threads.

However I don't understand why this specific code change makes any difference. We're still doing 1:1 tasks with flexbodies. The performance gain seems like a coincidence, perhaps the larger flexbodies with more N/B bindings 'warm up' the CPU caches better, but that's a speculation.

ulteq:
> You're looking in a very promising direction

I'm looking back. It was done this way in 0.4.07. :smile:

PR #1770, opened 2018-11-15
Fixed multiplayer selector window logic

ulteq:
Fixes the logic which controls the effects of the 'Network enable' and 'Preselected Map' settings.

The terrain selector window should only be shown if the server does not enforce a terrain and we do not have a pre-selected terrain.

### DIFF
--- a/source/main/main.cpp
+++ b/source/main/main.cpp
@@ -304,14 +304,13 @@ int main(int argc, char *argv[])
         AppState prev_app_state = App::app_state.GetActive();
         App::app_state.SetPending(AppState::MAIN_MENU);
 
-        if (! App::diag_preset_terrain.IsActiveEmpty())
-        {
-            App::app_state.SetPending(AppState::SIMULATION);
-        }
-
         if (App::mp_join_on_startup.GetActive() == true)
         {
             App::mp_state.SetPending(RoR::MpState::CONNECTED);
+        }
+        else if (!App::diag_preset_terrain.IsActiveEmpty())
+        {
+            App::app_state.SetPending(AppState::SIMULATION);
         }
 
         while (App::app_state.GetPending() != AppState::SHUTDOWN)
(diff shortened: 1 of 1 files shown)


source/main/main.cpp
```diff
@@ -304,15 +304,14 @@ int main(int argc, char *argv[])
...
-        if (! App::diag_preset_terrain.IsActiveEmpty())
-        {
-            App::app_state.SetPending(AppState::SIMULATION);
-        }
-
         if (App::mp_join_on_startup.GetActive() == true)
         {
             App::mp_state.SetPending(RoR::MpState::CONNECTED);
         }
+        else if (!App::diag_preset_terrain.IsActiveEmpty())
```
only-a-ptr: This was indeed broken.

source/main/MainMenu.cpp
```diff
@@ -224,6 +224,11 @@ void MainMenu::MainMenuLoopUpdate(float seconds_since_last_frame)
             String terrain_name = Networking::GetTerrainName();
             if (terrain_name != "any")
             {
+                if (gui->GetMainSelector()->IsVisible())
```
only-a-ptr: This looks like a workaround. We shouldn't show the selector window in the first place. I can't even figure out how could we end up in such state.

ulteq: `SkipMainMenu=Yes` combined with a server that enforces a certain terrain.

ulteq: The selector windows is opened before we know if the server enforces a certain terrain.

only-a-ptr: Then we have 2 options:
* Fix `SkipMainMenu` to only show selector early in singleplayer, and to wait for server info in multiplayer. That's sensible design IMO.
* Add comment `// Config option [SkipMainMenu=Yes] opens the terrain selector before we know the server's map`

ulteq: You were right, we don't need to account for `SkipMainMenu` at all in multiplayer.

ulteq: And the clean code team won once again. :smile:

PR #1771, opened 2018-11-15
Fixed broken actor stream setup

ulteq:
The actor stream setup was never executed due to a flaw in the `ar_uses_networking` logic.

### DIFF
--- a/source/main/physics/Beam.cpp
+++ b/source/main/physics/Beam.cpp
@@ -4313,7 +4313,6 @@ Actor::Actor(
     , ar_autopilot(nullptr)
     , ar_is_police(false)
     , m_disable_default_sounds(false)
-    , ar_uses_networking(rq.asr_origin == RoR::ActorSpawnRequest::Origin::NETWORK)
     , ar_engine(nullptr)
     , ar_driveable(NOT_DRIVEABLE)
     , m_skid_trails{} // Init array to nullptr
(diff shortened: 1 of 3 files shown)


only-a-ptr:
I never really understood what `ar_uses_networking` is, thanks for investigating, but now with `//...might have...`  I think it's bad design.

⌛️ Looking at it...

ulteq:
`//...might have...` since there won't be any remote actors if you play alone on multiplayer.

only-a-ptr:
Sure, but both the variable name and comment look weird.

I suggest `bool ar_net_transmitting:1; //!< We have a working remote instance`. The desire to setup stream should be implicit when GVar `mp_state == CONNECTED`

ulteq:
I also thought about renaming the variable, but I failed at coming up with a good name.

`ar_net_transmitting` sounds about right.

only-a-ptr:
> Or we remove it and just check the mp_state.

That would be ideal, but can we do that? Don't we need to distinguish ready-to-go actors from stream-setup-in-progress ones?

only-a-ptr:
> what if we wanted to support...

Then we'll code it, like anything else we want to support. 😄 

> ...static actors which don't need to be synced across the net

Sounds like one of many possible distant future continuities 👾 But as long as all RoR's actors are "trucks", "loads" and this sort of stuff, we're fine.

PR #1772, opened 2018-11-16
Fixed broken actor stream register

ulteq:
-> https://github.com/RigsOfRods/ror-server/pull/55


only-a-ptr:
WTF? The struct was equivalent in client and server codebases, how was anything 'broken'?

ulteq:
The first bytes of the `ActorStreamRegister` needs to be identical to the first bytes of `StreamRegister`. The last RoRNet update broke it.

only-a-ptr:
I wouldn't have to ask if you had put the exact same text in 
```
struct ActorStreamRegister
{
    // The first bytes need to be identical to StreamRegister`
    int32_t type; //!< stream type
```
etc...

PR #1784, opened 2018-11-19
Fixed a few truck reset issues

ulteq:
Properly resets the:
- Wheel torque
- Turn indicators
- Input shaft RPM
- (Locked) Differentials


only-a-ptr:
👍

PR #1786, opened 2018-11-20
✨ Godlike brakes

ulteq:
What this does:
- It completely removes the backspin during braking
- It completely removes any creeping when parking on a slope
- It makes all the anti-jitter / anti-skidding magic obsolete

How it does that:
- It calculates the mass of the rotating parts of the wheel
- It records the external forces (`retorque`) that act on the wheel
- It calculates the force required to counteract those and the momentum

We can built upon the `retorque` concept in other parts of the code (limited slip differentials, true clutch slipping, ...).

### DIFF
--- a/source/main/physics/BeamForcesEuler.cpp
+++ b/source/main/physics/BeamForcesEuler.cpp
@@ -297,8 +297,6 @@ void Actor::CalcWheels(bool doUpdate, int num_steps)
         if (ar_wheels[i].wh_is_detached)
             continue;
 
-        ar_wheels[i].wh_avg_speed = ar_wheels[i].wh_avg_speed * 0.995 + ar_wheels[i].wh_speed * 0.005;
-
         float relspeed = ar_nodes[0].Velocity.dotProduct(getDirection());
         float curspeed = fabs(relspeed);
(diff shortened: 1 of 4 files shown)


source/main/physics/BeamForcesEuler.cpp
```diff
@@ -341,41 +339,33 @@ void Actor::CalcWheels(bool doUpdate, int num_steps)
                 dbrake = ar_brake_force * abs(ar_hydro_dir_state);
             }
 
-            if ((abrake != 0.0 || dbrake != 0.0 || hbrake != 0.0) && m_num_braked_wheels != 0 )
+            if (abrake != 0.0 || dbrake != 0.0 || hbrake != 0.0)
             {
-                float brake_coef = 1.0f;
-                float antilock_coef = 1.0f;
+                float adbrake = abrake + dbrake;
```
only-a-ptr: Abrakadabraka FTW 😆

ulteq: If `abrake` and `dbrake` come together you get `adbrake`. Seemed logical. :smile:

PR #1788, opened 2018-11-20
Fixed out of bounds access

ulteq:
Coverity Scan reports:
```
overrun-buffer-arg: Overrunning buffer pointed to by bufferContent of 16384 bytes
by passing it to a function which accesses it at byte offset 16399
using argument len (which evaluates to 16384).
```


only-a-ptr:
👍 Good to have Coverity

PR #1814, opened 2018-11-24
Moved UpdateEngineAudio out of the tight physics loop

ulteq:


only-a-ptr:
👍 The less `doUpdate`, the better

PR #1815, opened 2018-11-24
'node_t' cleanup

ulteq:
The initial node positions (`initial_pos`) are only used during truck reset and don't need to be part of the node struct.

The `wheelid` became obsolete during the drive train refactoring.

The `iswheel` documentation was incorrect.


only-a-ptr:
👍 That's the way to go.

PR #1821, opened 2018-11-26
✨ Introduced MurmurHash3 hashing

ulteq:
Much faster than SHA1. Perfect fit for content hashing (detection of duplicates, ...).


paroj:
also available via Ogre::FastHash.

ulteq:
:man_facepalming:

ulteq:
Wait a second, FastHash is not using MMH3 or something of similar quality, is it?

Not that it would matter, but I kinda like MMH3.

paroj:
https://github.com/OGRECave/ogre/blob/master/OgreMain/include/OgreCommon.h#L47-L50

ulteq:
Convinced :smile:

PR #1823, opened 2018-11-26
✨ Much faster content hashing

ulteq:
With 8.5 GiB content 18 instead of 28 seconds for a full cache regen.


only-a-ptr:
Eh, not so good...
obrazek

ulteq:
Maybe because of this: https://github.com/RigsOfRods/rigs-of-rods/commit/a5081ddea4b98bdba89543b5fd682df18bd58ca5 ?

only-a-ptr:
Maybe, and my compiler keeps crashing...

PR #1824, opened 2018-11-26
🐛 Fixed linking Caelum on Linux

AnotherFoxGuy:


tritonas00:
works fine on top of master

only-a-ptr:
:+1: Looks very good

PR #1825, opened 2018-11-27
Fixed wheels2 spawn pressure

ulteq:
Also improved the pressure change logic.


ulteq:
It delays the call of `RoR::App::GetOverlayWrapper()->showPressureOverlay(false);`

PR #1827, opened 2018-11-27
Fixed crash to desktop after regen

ulteq:
How to reproduce:

1. Rebuild the cache
2. Load [Aspen Grove](https://github.com/RigsOfRods/rigs-of-rods/files/2619701/AspenGrove.zip) without restarting the game

```

Technical details below: 

ItemIdentityException: ParticleSystem template with name 'tracks/TurbopropSmoke' already exists. in ParticleSystemManager::createTemplate at /home/archie/ror-dependencies/Source/ogre/OgreMain/src/OgreParticleSystemManager.cpp (line 199)
```
@only-a-ptr I don't know if this is the correct way of fixing it. I also experimented around with `destroyParticleSystem`, but that didn't work.

### DIFF
--- a/source/main/resources/CacheSystem.cpp
+++ b/source/main/resources/CacheSystem.cpp
@@ -2206,7 +2206,7 @@ void CacheSystem::loadSingleDirectory(String dirname, String group, bool already
 #ifdef USE_OPENAL
             SoundScriptManager::getSingleton().clearNonBaseTemplates();
 #endif //OPENAL
-            //ParticleSystemManager::getSingleton().removeTemplatesByResourceGroup(rgname);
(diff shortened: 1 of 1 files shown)


only-a-ptr:
I didn't see the code yet, but the base problem here is that our cache-regen performs "initializeResourceGroup()" which processes scripts (materials, soundscripts, particles) for no reason, really. All we need is listing files from the RG's which doesn't need the init.

This crash should theoretically occur also if you go to the map, back to menu and to the map again.

ulteq:
> This crash should theoretically occur also if you go to the map, back to menu and to the map again.

It does not.

only-a-ptr:
> It does not.

Makes sense, why would a map define TurbopropSmoke? It's some mod, cache regen leaves it's resource hanging around, simulation spawning doesn't.

PR #1833, opened 2018-11-28
Fixed broken overview map icons

ulteq:
What a mess ...


ulteq:
I went all in and fixed everything. :blush:

Issue #1837, opened 2018-11-28
Phast Phantom rear wheels spawn broken


tritonas00:
As the title says, spawn [Phast_Phantom.zip](https://github.com/RigsOfRods/rigs-of-rods/files/2626359/Phast_Phantom.zip)

![pha](https://user-images.githubusercontent.com/2660424/49293363-9ffb3380-f4b8-11e8-998a-f645614e696e.png)

only-a-ptr:
🙏 RoR.log

tritonas00:
Sorry i forgot: [RoR.log](https://github.com/RigsOfRods/rigs-of-rods/files/2628823/RoR.log)

ulteq:
Three physics frames into the simulation:

![shocks](https://user-images.githubusercontent.com/9437207/50740744-527a9080-11f3-11e9-9bd7-42e378a8ccb2.jpg)

tritonas00:
mod issue

PR #1840, opened 2018-11-29
Removed unnecessary double id from the node struct

ulteq:
Positive side effect: The true wheel nodes indices made their way back into the node details debug overlay.

@only-a-ptr Do we really have to keep track of which nodes are auto-generated an which are not? I don't think so, but I might be wrong.

### DIFF
--- a/source/main/gameplay/SceneMouse.cpp
+++ b/source/main/gameplay/SceneMouse.cpp
@@ -171,7 +171,7 @@ bool SceneMouse::mouseMoved(const OIS::MouseEvent& _arg)
 
             for (std::vector<hook_t>::iterator it = grab_truck->ar_hooks.begin(); it != grab_truck->ar_hooks.end(); it++)
             {
-                if (it->hk_hook_node->id == minnode)
+                if (it->hk_hook_node->pos == minnode)
                 {
                     grab_truck->ToggleHooks(it->hk_group, MOUSE_HOOK_TOGGLE, minnode);
                 }
--- a/source/main/physics/Beam.cpp
+++ b/source/main/physics/Beam.cpp
@@ -3261,8 +3261,8 @@ void Actor::ToggleTies(int group)
                         if (!itr->multilock && itr->in_use)
                             continue;
 
-                        //skip if tienode is ropable too (no hk_selflock)
-                        if (itr->node->id == it->ti_beam->p1->id)
+                        // skip if tienode is ropable too (no selflock)
+                        if (this == actor && itr->node->pos == it->ti_beam->p1->pos)
                             continue;
 
                         // calculate the distance and record the nearest ropable
@@ -3390,7 +3390,7 @@ void Actor::ToggleHooks(int group, hook_states mode, int node_number)
     // iterate over all hooks
     for (std::vector<hook_t>::iterator it = ar_hooks.begin(); it != ar_hooks.end(); it++)
     {
-        if (mode == MOUSE_HOOK_TOGGLE && it->hk_hook_node->id != node_number)
+        if (mode == MOUSE_HOOK_TOGGLE && it->hk_hook_node->pos != node_number)
         {
             //skip all other nodes except the one manually toggled by mouse
             continue;
@@ -3459,7 +3459,7 @@ void Actor::ToggleHooks(int group, hook_states mode, int node_number)
                             continue;
 
                         // exclude this truck and its current hooknode from the locking search
-                        if (this == actor && i == it->hk_hook_node->id)
+                        if (this == actor && i == it->hk_hook_node->pos)
                             continue;
 
                         // a lockgroup for this hooknode is set -> skip all nodes that do not have the same lockgroup (-1 = default(all nodes))
--- a/source/main/datatypes/node_t.h
+++ b/source/main/datatypes/node_t.h
@@ -46,11 +46,10 @@ struct node_t
     Ogre::Real surface_coef;
     Ogre::Real volume_coef;
 
-    char iswheel; //!< 0=NOWHEEL, 1=WHEEL_DEFAULT, 2=WHEEL_2  3=WHEEL_FLEXBODY
-    char nd_coll_bbox_id; //!< Optional attribute (-1 = none) - multiple collision bounding boxes defined in truckfile
-    short nd_lockgroup;
-    short pos;     //!< This node's index in Actor::ar_nodes array.
-    short id;      //!< Numeric identifier assigned in truckfile (if used), or -1 if the node was generated dynamically.
+    int16_t iswheel;          //!< 0=NOWHEEL, 1=WHEEL_DEFAULT, 2=WHEEL_2  3=WHEEL_FLEXBODY
+    int16_t pos;              //!< This node's index in Actor::ar_nodes array.
+    int16_t nd_coll_bbox_id;  //!< Optional attribute (-1 = none) - multiple collision bounding boxes defined in truckfile
+    int16_t nd_lockgroup;     //!< Optional attribute (-1 = default, 9999 = deny lock) - used in the hook lock logic
 
     Ogre::Real      nd_avg_collision_slip;   //!< Physics state; average slip velocity across the last few physics frames
     Ogre::Vector3   nd_last_collision_slip;  //!< Physics state; last collision slip vector
--- a/source/main/gfx/GfxActor.cpp
+++ b/source/main/gfx/GfxActor.cpp
@@ -779,7 +779,7 @@ void RoR::GfxActor::UpdateDebugView()
                 {
                     ImVec2 pos_xy(pos.x, pos.y);
                     Str<25> id_buf;
-                    id_buf << nodes[i].id;
+                    id_buf << nodes[i].pos;
                     drawlist->AddText(pos_xy, NODE_TEXT_COLOR, id_buf.ToCStr());
 
                     if (m_debug_view != DebugViewType::DEBUGVIEW_BEAMS)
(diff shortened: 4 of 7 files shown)


ulteq:
> We can't do this because numbers in truckfile may not match final array positions.

Right! :man_facepalming:

> since they will only ever be useful in the diagnostic view.

They're also used for lockgroups. Guess I'll have to think about it some more.

only-a-ptr:
> They're also used for lockgroups. Guess I'll have to think about it some more.

That's a bug, all IDs need to be resolved to positions here: https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/resources/rig_def_fileformat/RigDef_SequentialImporter.cpp#L380

ulteq:
Then I don't understand the issue anymore. Why do we need an extra `id` if it's not supposed to be used anyways?

only-a-ptr:
It's all legacy cruft waiting to be analyzed and eliminated. Everything should work using node indices and IDs should be only available for debug view.

ulteq:
That means we need to post process the auto generated ids after everything was parsed?

Basically just replace '-1' with the actual position in the node array?

ulteq:
> Your code is fine except it displays indices in log + diag view where IDs should be.

Maybe display pos + id + weight and hide the id if it matches the pos?

only-a-ptr:
> Maybe display pos + id + weight and hide the id if it matches the pos?

❓ That sounds complicated, hardly useful and weird. I'm for straightforward "use only indices, display only IDs"

ulteq:
The problem is that many trucks rely on the fact that you can create beams between auto generated nodes and regular nodes. But in order to do that you need to know the node index of the auto generated node.

And I can't think of any reasonable solution other than displaying the actual node index in the debug info.

How did the old debug overlay do it?

only-a-ptr:
> How did the old debug overlay do it?

It generated sequential node numbers on the go, starting with 0, so it effectively displayed indices.

I understand your point, but if it's a "diagnostic view", then it must display stuff the way modders typed it in truckfiles. A view displaying internal indices is not diagnostic, it's an editor utility. I think the right way to do it would be to have such 2 views.

ulteq:
I'll have to think about it some more.

ulteq:
> I understand your point, but if it's a "diagnostic view", then it must display stuff the way modders typed it in truckfiles.

The old parser did not allow non consecutive indices unless `nodes2` were used: https://github.com/only-a-ptr/ror-legacy-svn-trunk/blob/master/source/main/physics/input_output/SerializedRig.cpp#L1420-L1429

So what are the cases where the `id` does not match the `pos`? Aren't they extremely rare?

I don't think that the separate `id` adds any kind of value. It only adds confusion.

ulteq:
But they do happen and they're going to bite us heavily in the ass, no? Won't this change in behavior break any existing mod that relies on the section order to create beams between wheel nodes and regular nodes?

ulteq:
> That's a bug, all IDs need to be resolved to positions here: https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/resources/rig_def_fileformat/RigDef_SequentialImporter.cpp#L380

But won't this break triggers / lockgroups and such?

only-a-ptr:
This had been in place for 4 years now: https://github.com/RigsOfRods/rigs-of-rods/blame/master/source/main/resources/rig_def_fileformat/RigDef_SequentialImporter.h#L60-L71 The only mod known to be still broken is Lotus Esprit 🔫 007

> But won't this break triggers / lockgroups and such?

I assumed that you tested these thoroughly since you're proposing this change 😄

only-a-ptr:
Please read  https://forum.rigsofrods.org/threads/future-how-to-define-n-b-in-truckfile.617/ To understand my train of thought.

ulteq:
> The only mod known to be still broken is Lotus Esprit gun 007

Not true: https://github.com/RigsOfRods/rigs-of-rods/issues/1837

> triggers are already handled: https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/resources/rig_def_fileformat/RigDef_SequentialImporter.cpp#L637-L641>

I can't imagine that those two lines handles trigger blockers correctly:
> B: Blocks other triggers when triggered, a number in shortkey represent the number of trigger**s** to block at shortbound , a number in longkey the number of trigger**s** to release at longbound.

The trigger codes iterates across the node indices: https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/physics/Beam.cpp#L2435

ulteq:
> OK, we have 2 broken mods:

My RoR gut prediction tells me that there are hundreds of mods which are affected by this.

source/main/physics/Beam.cpp
```diff
@@ -3459,7 +3459,7 @@ void Actor::ToggleHooks(int group, hook_states mode, int node_number)
                             continue;
 
                         // exclude this truck and its current hooknode from the locking search
-                        if (this == actor && i == it->hk_hook_node->id)
+                        if (this == actor && i == it->hk_hook_node->pos)
```
ulteq: But we can't merge this without updating the parser first, right?

only-a-ptr:
> My RoR gut prediction tells me that there are hundreds of mods which are affected by this.

So what do you suggest?

I'm not extra worried about breaking mods, users report that quickly and loudly, and RoR's license says `... made in a hope to be useful but WITHOUT ANY WARRANTY`. However, I'm very much worried about the creeping stagnation caused by too many arcane and fragile features blocking further development.

source/main/physics/Beam.cpp
only-a-ptr: I think we should walk feature-by-feature, always checking if the importer works right, then replacing IDs with positions, testing and merging.

ulteq: Sounds perfect.

ulteq:
> So what do you suggest?

Drop support for stupid things like wheels sections between node sections in all newly created tools / diagnostic views / etc. But at the same time make the parser smart enough to handle those cases for backwards compatibility. Basically what you already did.

> I'm not extra worried about breaking mods,

I am, because we have ~3 active content creators and 10000+ existing mods.
RoR without support for existing content is pointless.

only-a-ptr:
> Basically what you already did.

Not basically, that's precisely what I've been after all this time.

> I am, because we have ~3 active content creators and 10000+ existing mods.

Oh, I think we're not understanding each other on "breaking mods". I was talking about accidental breakage, i.e. we're not sure how it works, so we clarify the code to work one way, and if we get it wrong, we just try again. I want to reach a state where we know exactly what features we have and what's a broken truckfile. I think (potential) accidental breakage is lesser evil than limiting development by worrying what we might break.

ulteq:
> that's precisely what I've been after all this time.

Perfect. So all we have to do for now (0.4.8) is to verify that the parser correctly handles lockgroups and trigger blockers?

source/main/gameplay/SceneMouse.cpp
```diff
@@ -171,7 +171,7 @@ bool SceneMouse::mouseMoved(const OIS::MouseEvent& _arg)
 
             for (std::vector<hook_t>::iterator it = grab_truck->ar_hooks.begin(); it != grab_truck->ar_hooks.end(); it++)
             {
-                if (it->hk_hook_node->id == minnode)
+                if (it->hk_hook_node->pos == minnode)
```
ulteq: I think `pos` is correct here.

source/main/physics/Beam.cpp
```diff
@@ -3390,7 +3390,7 @@ void Actor::ToggleHooks(int group, hook_states mode, int node_number)
     // iterate over all hooks
     for (std::vector<hook_t>::iterator it = ar_hooks.begin(); it != ar_hooks.end(); it++)
     {
-        if (mode == MOUSE_HOOK_TOGGLE && it->hk_hook_node->id != node_number)
+        if (mode == MOUSE_HOOK_TOGGLE && it->hk_hook_node->pos != node_number)
```
ulteq: I think `pos` is correct here.

source/main/physics/Beam.cpp
```diff
@@ -3262,7 +3262,7 @@ void Actor::ToggleTies(int group)
                             continue;
 
                         //skip if tienode is ropable too (no hk_selflock)
-                        if (itr->node->id == it->ti_beam->p1->id)
+                        if (itr->node->pos == it->ti_beam->p1->pos)
```
ulteq: I'm not entirely sure, but it seems as if it wouldn't matter if we use `id` or `pos` here.

source/main/physics/Beam.cpp
ulteq: After thinking about it some more I think that `id` was wrong, and `pos` is the way to go here.

source/main/gameplay/SceneMouse.cpp
only-a-ptr: Since `minnode` is an index, `pos` is definitely correct here.

source/main/physics/Beam.cpp
only-a-ptr: I can't figure this out..
`it` is a local tie.
`itr` is a neighbour ropable.
`if (itr->node->pos == it->ti_beam->p1->pos)` - makes no sense, we compare indices of neighbour ropable node and our tie-beam p1. But those live in different arrays. Tie has p1 always on local truck, p2 attaches to the neighbour.

source/main/physics/Beam.cpp
only-a-ptr: ✔️ Yes because value of `node_number` is an index: https://github.com/RigsOfRods/rigs-of-rods/blob/cf08a7dd0b930c7f6da8a7ffb23a5dc689e55aed/source/main/gameplay/SceneMouse.cpp#L176

source/main/physics/Beam.cpp
only-a-ptr: ✔️

source/main/physics/Beam.cpp
ulteq: Same as: https://github.com/RigsOfRods/rigs-of-rods/pull/1840#pullrequestreview-180252993

PR #1844, opened 2018-11-30
✨ Added blender 2.79 import/export plugins

ulteq:
Features:
- Full vertex group support
- Automatic column alignment
- Sophisticated vertex position rounding
- Full preservation of the original truck file
- Automatic beam sorting based on the vertex groups
- All node, beam and cab options are preserved during import / export
- All `set_node_defaults`, `set_beam_defaults` and `set_beam_defaults_scale` entries are preserved

Repository entry: https://www.rigsofrods.org/forum/resources/blender-import-export-plugin.25/

### DIFF
--- a/tools/blender/ror_export.py
+++ b/tools/blender/ror_export.py
@@ -1,0 +1,155 @@
+# <pep8-80 compliant>
+
+bl_info = {
+    "name": "RoR Exporter",
+    "author": "ulteq",
+    "version": (0, 0, 1),
+    "blender": (2, 79, 0),
+    "category": "RoR",
+}
+
+import bpy
+import json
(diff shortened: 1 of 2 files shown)


only-a-ptr:
Wow, you've done Python before, right? I obviously have gaps, I'll need a while to check this, but it looks great.

ulteq:
> Wow, you've done Python before, right?

Yes.

> I'll need a while to check this

The most interesting part is how we use custom data layers to store information about the n/b structure on the actual blender objects. This makes them survive any kind of modification (reordering, insertion, deletion) inside of blender.

We can easily extend this concept beyond basic node / beam support.

PR #1851, opened 2018-12-01
Added 'GetMaxHeight' helper

ulteq:
Might come in handy at some point.


only-a-ptr:
:scream: Have you just committed deadcode?

PR #1854, opened 2018-12-02
Removed 'GetMaxHeight' helper

ulteq:
@only-a-ptr, you're right. It was a bad idea to add code which might be useful in the future.

Reverts: https://github.com/RigsOfRods/rigs-of-rods/pull/1851

### DIFF
--- a/source/main/terrain/TerrainGeometryManager.cpp
+++ b/source/main/terrain/TerrainGeometryManager.cpp
@@ -170,14 +170,6 @@ void TerrainGeometryManager::getTerrainPositionAlign(Real x, Real y, Real z, Ter
     };
 }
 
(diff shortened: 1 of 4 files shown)


only-a-ptr:
Don't get me wrong, but "it might be useful in the future" is a no. 1 reason for bloating codebases. There's even a named anti-pattern about it: https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it

ulteq:
1. I implemented it the overview map texture generator
2. I realized that I don't need it
3. I didn't want to delete it <- this is were I fucked up :smile:

only-a-ptr:
Lecture just for you: https://www.youtube.com/watch?v=Ev7lM7SWVHE

PR #1862, opened 2018-12-04
Simplified terrain normal calculation

ulteq:


ulteq:
> Also I think the code will produce skewed results because it only samples X+1 and Z+1

That would have been the next step: https://github.com/ulteq/rigs-of-rods/commit/2c73bcdd72aac4a7dc31269451ca23c80bab68d4

paroj:
did not look too close at the code, but is `plane.normal` in `getHeightAtTerrainPosition` not what you are actually after?

ulteq:
That was also my first idea, but it doesn't work. Our physics simulation seems to depend on the smoothing effect that comes off the multi sampling.

PR #1875, opened 2018-12-08
🔧 Create missing folders

AnotherFoxGuy:
RoR will now create any missing folders automatically

### DIFF
--- a/source/main/utils/PlatformUtils.h
+++ b/source/main/utils/PlatformUtils.h
@@ -40,13 +40,15 @@ namespace RoR {
 
 extern char PATH_SLASH;
 
-bool FileExists(const char* path);   //!< Path must be UTF-8 encoded.
+bool FileExists(const char *path);   //!< Path must be UTF-8 encoded.
 bool FolderExists(const char* path); //!< Path must be UTF-8 encoded.
 void CreateFolder(const char* path); //!< Path must be UTF-8 encoded.
+void CreateFolderIfNotExists(const char* path);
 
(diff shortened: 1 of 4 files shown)


source/main/utils/PlatformUtils.h
```diff
@@ -40,13 +40,15 @@ namespace RoR {
 
 extern char PATH_SLASH;
 
-bool FileExists(const char* path);   //!< Path must be UTF-8 encoded.
+bool FileExists(const char *path);   //!< Path must be UTF-8 encoded.
```
ulteq: Clang-Format

tritonas00:
Nice! I tried with no .rigsofrods dir to see what happens. It seems in that case, cache is not generated at all:

ca

ulteq:
> I tried with no .rigsofrods dir

No `.rigsofrods` dir mean no content, right?

ulteq:
Right!, @AnotherFoxGuy

ulteq:
Let's merge this and fix the content issue with another pull request.

AnotherFoxGuy:
> cache is not generated at all

That is an issue with missing .cfg files, but that is probably for another PR
Since some of the .cfg files are meant to be read-only so I don't see the point in having them
https://github.com/RigsOfRods/rigs-of-rods/blob/master/resources/skeleton/config/categories.cfg

ulteq:
The problem is that we forgot to automatically unpack the `skeleton.zip` if needed.

PR #1879, opened 2018-12-09
More GVars

ulteq:


source/main/physics/RigSpawner.cpp
```diff
@@ -386,7 +382,7 @@ void ActorSpawner::InitializeRig()
 
     m_placeholder_managedmat = RoR::OgreSubsystem::GetMaterialByName("rigsofrods/managedmaterial-placeholder"); // Built-in
 
-    m_apply_simple_materials = BSETTING("SimpleMaterials", false);
+    m_apply_simple_materials = App::gfx_simple_materials.GetActive();
```
only-a-ptr: This is more of `diag_simple_materials`

PR #1883, opened 2018-12-10
Removed CrashRpt leftovers

ulteq:


only-a-ptr:
When did we drop support for CrashRpt?

PR #1884, opened 2018-12-11
PathCombine helper

ulteq:


source/main/utils/PlatformUtils.h
```diff
@@ -44,6 +44,9 @@ inline bool FileExists(std::string const& path)   { return FileExists(path.c_str
 inline bool FolderExists(std::string const& path) { return FolderExists(path.c_str()); }
 inline void CreateFolder(std::string const& path) { CreateFolder(path.c_str()); }
 
+inline std::string PathCombine(std::string a) { return a + PATH_SLASH; };
```
only-a-ptr: I'm not so sure about this, are you using it anywhere?

ulteq: Twice, but we can remove it. I also don't like it that much.

PR #1886, opened 2018-12-11
Fixed the render system selection logic

ulteq:
As suggested by @AnotherFoxGuy, the `RoR.cfg` settings must be loaded first.


source/main/main.cpp
```diff
@@ -126,6 +126,11 @@ int main(int argc, char *argv[])
             return -1;
         }
 
+        Settings::getSingleton().LoadRoRCfg(); // Main config file - path obtained from GVars
+#if OGRE_PLATFORM != OGRE_PLATFORM_APPLE //MacOSX adds an extra argument in the form of -psn_0_XXXXXX when the app is double clicked
```
only-a-ptr: Looks archaic. I'd remove this until we actually start supporting Mac.

ulteq: Feel free to remove it on master. I will fix the merge conflict.

PR #1888, opened 2018-12-12
Removed RoRConfig

ulteq:
Maybe a bit early, but I think it's time.

- [x] Notify the player when he changes settings that require a restart https://github.com/RigsOfRods/rigs-of-rods/commit/1edaf1d0f68c675f897ac7699f85657be3207f77
- [x] Fix the language selection drop-down list box: #1925
- [x] Add a button to clear the cache #1944


tritonas00:
>resolution/vsync/fullscreen settings

Same as RoRConfig iirc. Render system is auto detected by RoR, if not it opens an ogre dialog for the user to choose. Resolution is the minimum found on the list.

PR #1889, opened 2018-12-12
🔧 Fixed debug build on windows

### LINKED ISSUE #1572: Fix Debug build
At the moment the debug build is disabled here: https://github.com/RigsOfRods/rigs-of-rods/blob/master/CMakeLists.txt#L4-L7
It would be nice if we got it working again because Conan only has prebuilt release and debug dependencies available

AnotherFoxGuy:
Fixes #1572


only-a-ptr:
It works! Thanks!

One minor glitch tho - plugins.cfg must be hand edited to add `_d` suffix to plugins. We may ressurect the support for 'plugins_d.cfg'.

Issue #1269, opened 2017-01-25
Localization is done wrong.


only-a-ptr:
Our localization works in the following manner:
 1. A developer writes English phrases into the source, for example `_L("Loader")`
 2. The `_L()` macro detects user's language and if it's non-english, it looks up the translation for the english phrase and displays that, instead.

This is plain wrong because languages don't work this way. The fact that single english phrase/word suits multiple places doesn't mean the same will be true for any other language. The phrase "Loader" above was not chosen randomly - when I first started 0.38, it used my native language (czech) and the loader UI window was titled "Nakladač" (= "Cargo loader"). I immediately knew where's the problem.

Our options:
* [UPDATE Dec 2018] Use gettext contexts: seems to be exactly what we need to resolve the ambiguities: https://www.gnu.org/software/gettext/manual/html_node/Contexts.html
 * use ID labels like MAINMENU_BUTTON_OPTIONS. Feature: if a translation is missing, this label will be displayed directly. This will make everyone notice it's wrong, which is both good and bad thing depending on context. Advantage: If someone needs to rephrase the English text, (s)he only needs to edit the translation, not the ID-label. Advantage2: Since the ID-label is not intended for display, programmer is free to give it any name, hence no clash needs to occur.
* Use english text + some kind of text-group ID. Both groupID+text would be used to look up the translation. Advantage: If the translation is missing, the english text is used. Disadvantage: If the text needs to be rephrased, ALL translations need to be updted because the text is the key, literally. Disadvantage2: If you get 2 equal english words close to each other, for example on the same UI panel, you'll probably have single groupID reserved for the panel. In that case, you either violate the uniqueness principle (backward!), create more groups (hack!) or somehow phrase the english to become unique (silly).

I favor the former option (MAINMENU_BUTTON_OPTIONS).

Opinions?

only-a-ptr:
Re-making of the GUI has begun: https://github.com/RigsOfRods/rigs-of-rods/pull/1333 - so it's time to start re-making the localization, too.

We need a syntax for the message ID strings. Here's my draft:

```
// Syntax: AREA]/WINDOW[/CONTAINER]/[TextPreview]
// Areas:
//  m = main menu
//  s = simulation
//  x = (multiple areas)
//  r = rig editor     (near future)
//  t = terrain editor (future)

// Examples:

m/menu/SinglePlayer         // Main menu
m/menu/MultiPlayer
m/menu/Settings
m/menu/About
[...]

PR #1892, opened 2018-12-12
🎌 Localized settings menu

AnotherFoxGuy:


only-a-ptr:
I'm not sure this is how we want to localize - when two places use the same text in English, it doesn't mean the same text will fit in other languages. I elaborated about that here: https://github.com/RigsOfRods/rigs-of-rods/issues/1269

ulteq:
But that's something for 0.4.9, right?

PR #1893, opened 2018-12-12
Fix prop anim. for event:TRUCK_TOGGLE_CONTACT

only-a-ptr:
This is a minor issue, but I couldn't resist fixing it because it presented a challenge.

Animated props only checked actual controller input, so `event:TRUCK_TOGGLE_CONTACT` didn't work when vehicle was spawned with started engine.

I've laid out an architecture which handles both cases.

### DIFF
--- a/source/main/gfx/GfxActor.cpp
+++ b/source/main/gfx/GfxActor.cpp
@@ -1377,7 +1377,8 @@ void RoR::GfxActor::UpdateSimDataBuffer()
         m_simbuf.simbuf_gear            = m_actor->ar_engine->GetGear();
         m_simbuf.simbuf_autoshift       = m_actor->ar_engine->getAutoShift();
         m_simbuf.simbuf_engine_rpm      = m_actor->ar_engine->GetEngineRpm();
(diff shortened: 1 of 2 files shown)


ulteq:
I don't want to disappoint you, but it works fine in master and this PR breaks it.

tritonas00:
It works on master if you turn off the engine. If you turn it on again it the disappears.

ulteq:
`m_starter_has_contact` is a state which is flipped when you press `EV_TRUCK_TOGGLE_CONTACT`.

only-a-ptr:
This is tricky, the docs for animated props says `event:` specifies a keypress event. The correct way of handling it would be to add a new `source: starter_contact` for animated props and update the Tourliner, otherwise the dashboard will always appear broken to users.

ulteq:
> what if the respective vehicle states are changed internally?

Emulate a key press and be done with it? The plan is to buffer input events anyways, right?

only-a-ptr:
> Emulate a key press and be done with it? The plan is to buffer input events anyways, right?

Emulating keypress will need the buffeer + monitoring logic which un-sets it again, but it makes the most sense.

PR #1901, opened 2018-12-14
Minor code cleanup

ulteq:


ulteq:
`HashFile` is only ever called if `ft == 0`, just like before, no?

ulteq:
Maybe add a little comment:
```
// check file time, if that fails, fall back to hash comparison
```

PR #1911, opened 2018-12-16
Use custom font for IMGUI

AnotherFoxGuy:


ulteq:
Needs a little getting used to, but I think it's much better than before.

PR #1913, opened 2018-12-16
✨ Overview map icon revival

ulteq:
* Fixed the rotation of the player icon
* Fixed the size of the race icons
* Removed the teleport window
* Streamlined the terrain initialization

aspen-map


only-a-ptr:
Fantastic 👍 

The races functionality seems to be different now - is there any reference content?

ulteq:
> The races functionality seems to be different now

The new code works with all terrains I have. The old code didn't work with any of them.

> is there any reference

https://docs.rigsofrods.org/terrain-creation/scripting/#how-do-races-work

Terrain code: `...terrn.as` <---> AngleScript API: `races.as` <---> RoR terrain parser

PR #1920, opened 2018-12-18
Attempt to improve the ImGUI font sharpness

ulteq:


only-a-ptr:
I needed a close look but yes, there is an improvement:
image

PR #1928, opened 2018-12-19
Minor map icon size adjustment

ulteq:
nhelens-minimap


AnotherFoxGuy:
The icons are way to small on my 1600X900 laptop

ulteq:
Maybe we should double the size on the big map?

Issue #1909, opened 2018-12-15
Compiling error on linux: `'App' has not been declared`


pkillnine:
I'm trying to compile Rigs Of Rods (latest git master as of writing this bug report) on GuixSD. At the bottom of this bug report is the output of cmake which shows which optional dependencies I have used.

I get the following compilation error:

```
Scanning dependencies of target generate_version
[  0%] Generate files containing version information
true
true
[  0%] Built target generate_version
Scanning dependencies of target version_info
[  1%] Building CXX object source/version_info/CMakeFiles/version_info.dir/version_info.cpp.o
[  1%] Linking CXX static library ../../bin/libversion_info.a
[  1%] Built target version_info
[  1%] Generating CXX unity source source/main/cotire/RoR_CXX_unity.cxx
[  2%] Generating CXX prefix source source/main/cotire/RoR_CXX_prefix.cxx
-- Result 1 scanning includes of /tmp/rigs-of-rods/source/main/cotire/RoR_CXX_unity.cxx.
-- 35 line(s) skipped, see /tmp/rigs-of-rods/source/main/cotire/RoR_CXX_prefix.cxx.log
[  2%] Generating CXX prefix header source/main/cotire/RoR_CXX_prefix.hxx
[  2%] Building CXX precompiled header source/main/cotire/RoR_CXX_prefix.hxx.gch
Scanning dependencies of target RoR
[  2%] Building CXX object source/main/CMakeFiles/RoR.dir/main.cpp.o
In file included from /tmp/rigs-of-rods/source/main/gui/GUIManager.h:28,
                 from /tmp/rigs-of-rods/source/main/main.cpp:27:
/tmp/rigs-of-rods/source/main/gui/imgui/OgreImGui.h:79:11: warning: ‘OgreImGui::ImGUIRenderable’ declared with greater visibility than the type of its field ‘OgreImGui::ImGUIRenderable::mRenderOp’ [-Wattributes]
     class ImGUIRenderable : public Ogre::Renderable
           ^~~~~~~~~~~~~~~
/tmp/rigs-of-rods/source/main/main.cpp: In function ‘int main(int, char**)’:
/tmp/rigs-of-rods/source/main/main.cpp:77:9: error: ‘App’ has not been declared
         App::sys_process_dir.SetActive(RoR::GetParentDirectory(exe_path.c_str()).c_str());
         ^~~
[...]

Issue #1908, opened 2018-12-15
build fails with own deps


Speciesx:
Windows 10 Pro x64
VS Community 2017 15.9.4

https://github.com/RigsOfRods/ror-dependencies

```
1>------ Build started: Project: ZERO_CHECK, Configuration: Release x64 ------
1>Checking Build System
1>CMake does not need to re-run because D:/ror/dev/rigs-of-rods_clean/rigs-of-rods/build_x64/CMakeFiles/generate.stamp is up-to-date.
1>CMake does not need to re-run because D:/ror/dev/rigs-of-rods_clean/rigs-of-rods/build_x64/source/CMakeFiles/generate.stamp is up-to-date.
1>CMake does not need to re-run because D:/ror/dev/rigs-of-rods_clean/rigs-of-rods/build_x64/source/version_info/CMakeFiles/generate.stamp is up-to-date.
1>CMake does not need to re-run because D:/ror/dev/rigs-of-rods_clean/rigs-of-rods/build_x64/source/main/CMakeFiles/generate.stamp is up-to-date.
1>CMake does not need to re-run because D:/ror/dev/rigs-of-rods_clean/rigs-of-rods/build_x64/source/angelscript_addons/CMakeFiles/generate.stamp is up-to-date.
1>CMake does not need to re-run because D:/ror/dev/rigs-of-rods_clean/rigs-of-rods/build_x64/doc/CMakeFiles/generate.stamp is up-to-date.
2>------ Build started: Project: generate_version, Configuration: Release x64 ------
3>------ Build started: Project: angelscript_addons, Configuration: Release x64 ------
4>------ Skipped Build: Project: clean_cotire, Configuration: Release x64 ------
4>Project not selected to build for this solution configuration 
5>------ Skipped Build: Project: zip_and_copy_resources, Configuration: Release x64 ------
5>Project not selected to build for this solution configuration 
2>Generate files containing version information
2>true
3>Generating CXX unity source D:\ror\rigs-of-rods\build_x64\source\angelscript_addons\Release\cotire\angelscript_addons_CXX_unity.cxx
2>true
3>Building Custom Rule D:/ror/dev/rigs-of-rods_clean/rigs-of-rods/source/angelscript_addons/CMakeLists.txt
3>CMake does not need to re-run because D:/ror/dev/rigs-of-rods_clean/rigs-of-rods/build_x64/source/angelscript_addons/CMakeFiles/generate.stamp is up-to-date.
[...]

Issue #1420, opened 2017-12-11
Arch Linux build issue


gdm413229:
# My issue with building Rigs of Rods

## Tick-box summary of what goes through the linker/librarian without trouble

 - [x] AngelScript Addons
 - [x] RoR Configurator
 - [x] RoR Version Information
 - [x] SocketW
 - [ ] Rigs of Rods Executable

## Description of the build issue

Everything compiled and assembled just fine, but the proper Rigs of Rods executable didn't get built due to an error related to undefined references that came out of the linker.  Rigs of Rods requires SocketW to build and run, and if you can make the source in a way that SocketW is optional or not required for network play, that will be appreciated.  Linker error messages can be found below the description of this issue.

Other Arch Linux users can post their fixes for linking against SocketW, just so I don't have to use WINE, Windows, Ubuntu or Macintosh machines for RoR.

### Linker error messages

```
/usr/lib/gcc/x86_64-pc-linux-gnu/7.2.1/../../../../lib/libSocketW.so: undefined reference to `X509_set_notBefore'
/usr/lib/gcc/x86_64-pc-linux-gnu/7.2.1/../../../../lib/libSocketW.so: undefined reference to `X509_set_notAfter'
/usr/lib/gcc/x86_64-pc-linux-gnu/7.2.1/../../../../lib/libSocketW.so: undefined reference to `SSLv23_method'
/usr/lib/gcc/x86_64-pc-linux-gnu/7.2.1/../../../../lib/libSocketW.so: undefined reference to `SSL_load_error_strings'
/usr/lib/gcc/x86_64-pc-linux-gnu/7.2.1/../../../../lib/libSocketW.so: undefined reference to `SSL_library_init'
```

# Machine Specifications
## Hardware Specifications
Component | Name
----------|---------
CPU | AMD FX-8350
Graphics Adapter | NVIDIA GeForce GTX 1070
RAM | 16GB DDR3
Motherboard | ASUS 970 PRO GAMING/AURA
Internal Storage | ST2000DX001-1NS164 [reported by SATA host adapter]
## Software Specifications
Software | Version or Name
------------|----------
GNU GCC | 7.2.1
Linux Kernel | 4.14
Linux Kernel Architecture | x86-64
Linux Distribution | Arch Linux

tritonas00:
this is because socketw cant build with openssl 1.1

[...]

PR #1950, opened 2018-12-21
🔧 Massive build system cleanup

### LINKED ISSUE #1991: Add 'install' target for `make` (so you can run `make install`)
It would be good to be able to run `make install` so that it copies `RoR` into `<installation prefix>/bin`, etc. This makes it easier for packaging, and is fairly standard.

AnotherFoxGuy:
* Removed `FindXXX.cmake` files (See: https://cgold.readthedocs.io/en/latest/rejected/find-modules.html)
* Made Conan required
* Fixed #1991


ulteq:
* Removed compile definitions
-> No (we should at least keep `USE_ANGELSCRIPT`, `USE_SOCKETW` and `USE_OPENAL`)
* Made Conan required
-> No (the existing (Conan free) build path is working far to well to completely ditch it)

AnotherFoxGuy:
> @ulteq  No (the existing (Conan free) build path is working far to well to completely ditch it)

https://vector-of-bool.github.io/2018/10/15/pmm.html#fear-of-commitment

tritonas00:
> Made Conan required

no please

ulteq:
* What happens if you leave the project? Nobody of us knows how to maintain Conan / PMM.

* What happens if there are upstream issues with Conan / PMM? Just look at the issues @only-a-ptr already had with Conan so far ... I don't plan on having these.
How many new external tools do we depend on if this gets merged?
bintray? conan.io? pmm? python? ...

* I massively dislike it when tools automatically download and execute code from sources I don't have control of.

* How easy is it to manually recompile the dependencies with slight adoptions if needed?

Don't get me wrong. I do like the concept behind PMM (with Conan).

What's the issue with the existing FindXXX scripts? They don't hurt anyone do they? We can't we keep the old build path for a couple of months, then reevaluate?

And if we really decide to pull the trigger on this, then please after 0.4.8 is out.

AnotherFoxGuy:
> * What happens if you leave the project? Nobody of us knows how to maintain Conan / PMM.

I don't see why you wouldn't be able to learn how Conan recipes work, because Conan has better documentation than our homebrew solution(https://docs.conan.io/en/latest/creating_packages.html)

> * What happens if there are upstream issues with Conan / PMM? Just look at the issues only-a-ptr already had with Conan so far ... I don't plan on having these.

When we switched to the cmake meta-project we also had issues with building RoR.
And the cmake meta-project also had build issues (See #1909 #1908 #1420)

>   How many new external tools do we depend on if this gets merged?
>   bintray? conan.io? pmm? python? ...

PMM for managing conan, and Conan for managing the deps.

> * I massively dislike it when tools automatically download and execute code from sources I don't have control of.

All core deps are maintained by me (https://github.com/AnotherFoxGuy/conan-OGRE https://github.com/AnotherFoxGuy/conan-MyGUI etc...)
Also, the cmake meta-project depends on external cmake build systems.

> * How easy is it to manually recompile the dependencies with slight adoptions if needed?

You still can do that with Conan; just download the recipe, make the changes and run `conan create`

> What's the issue with the existing FindXXX scripts? They don't hurt anyone do they?

The issue with FindXXX scripts is that the grab random packages that may not work with RoR, also defining the correct version is extremely difficult with FindXXX scripts

ulteq:
> #1909 #1908

Both of them were not related with cmake.

> The issue with FindXXX scripts is that they grab random packages that may not work with RoR

That's why we have ror-dependencies, no?

> You still can do that with Conan; just download the recipe, make the changes and run `conan create`

And that will build all dependencies locally?

AnotherFoxGuy:
> That's why we have ror-dependencies, no?

FindXXX scripts prefer system-wide packages over external installed packages (https://github.com/Kitware/CMake/blob/master/Modules/FindOpenAL.cmake#L64-L70)

> And that will build all dependencies locally?

No, only the modified dependency, the others will still be downloaded

tritonas00:
@AnotherFoxGuy is `ninja` a requirement for this?

tritonas00:
i get https://pastebin.com/8ZYQ1GKf at OGRE

AnotherFoxGuy:
@tritonas00 I have never tested RelWithDebInfo, i will check

tritonas00:
Ok i did
```
cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo .
make -j8
make zip_and_copy_resources
```
when i `cd bin` then run `./RoR` it works but when i run outside of that dir (`bin/RoR` for example) it fails (cant find ogre plugins) because `PluginFolder=.` in plugins.cfg

tritonas00:
>The libCaelum issue should be fixed now

Indeed :+1:

AnotherFoxGuy:
@only-a-ptr What do you think about this?

AnotherFoxGuy:
> TL;DR Instead of forcing everyone to use a new package manager, improve the documentation on what is needed to build RoR, and improve the CMake files so they find these requirements.

The issue is that all CMake files are homebrew scripts, with no documentation on how to use/maintain it.

Also, with Conan you won't have issues like this anymore: #1979 #1958 since the deps are built on travis/appveyor and downloaded when you run `conan install`

pkillnine:
That wouldn't work for me because my distro doesn't use the standard filesystem hierarchy (/usr, /bin, etc), it stores each package in it's own directory, and links files directly to other packages' directories (It's GuixSD which uses similar system as NixOS).

The latter issue was fixed in the source code anyways, and the first issue has gone for me for some reason since putting it into a package.

PR #1961, opened 2018-12-22
🎮 Advanced contacter logic

ulteq:
Automatically turns any ground contactable node into a contacter node.

### DIFF
--- a/source/main/datatypes/node_t.h
+++ b/source/main/datatypes/node_t.h
@@ -57,7 +57,8 @@ struct node_t
     ground_model_t* nd_last_collision_gm;    //!< Physics state; last collision 'ground model' (surface definition)
 
     // Bit flags
-    bool            nd_contacter:1;          //!< Attr; This node is part of collision triangle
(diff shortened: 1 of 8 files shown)


only-a-ptr:
> Automatically turns any ground contactable node into a contacter node.

You mean: istead of relying on the modder to mark all cab-nodes as 'contacters', I added another flag 'is_cab_node' which is filled automatically.

I never really understood the complex break-ability evaluation that uses those flags. Can you educate me why we have that?

ulteq:
I renamed the old `nd_contacter` flag, because it wasn't related with [contacters](https://docs.rigsofrods.org/vehicle-creation/fileformat-truck/#contacters) at all.

The `nd_cab_node` flag is set if the node is part of a submesh triangle (aka `cab`). That flag is used to determine if a beam can break at all or not.

ulteq:
This simple patch really brings the submesh collisions to the next level. Feels so much better with existing content (which weren't created with submesh collisions in mind).

PR #1967, opened 2018-12-23
🔧 Replaced FetchContent with git submodules

AnotherFoxGuy:
This should also improve the cmake configure time 

Closes #1966

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,6 @@
+[submodule "content"]
+	path = content
+	url = https://github.com/RigsOfRods/content.git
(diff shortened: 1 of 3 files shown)


paroj:
I was already wondering why you switched to FetchContent in the first place..

tritonas00:
https://github.com/RigsOfRods/rigs-of-rods/pull/1723

dejavu?

AnotherFoxGuy:
Because switching from a branch with submodules to one without and back breaks git and I was doing that a lot with the upgrade from 1.9 to 1.11, so I replaced it with FetchContent because that was easier at that time

PR #1971, opened 2018-12-24
🐛 Fixed broken race event logic

ulteq:
A little Christmas present: Races are now once again enabled by default.


AnotherFoxGuy:
Truck triggers don't work anymore and avatar triggers only trigger once

ulteq:
What do you mean with truck (avatar) triggers? Example?

ulteq:
Should now be fixed.

PR #1988, opened 2018-12-27
🔧 Create missing folders

AnotherFoxGuy:


ulteq:
Just add:
```
CreateFolder(App::sys_cache_dir.GetActive());
CreateFolder(App::sys_screenshot_dir.GetActive());
```
and be done with it.

source/main/main.cpp
```diff
@@ -124,6 +124,10 @@ int main(int argc, char *argv[])
             return -1;
         }
 
+        CreateFolderIfNotExists(App::sys_cache_dir.GetActive());
+        CreateFolderIfNotExists(App::sys_config_dir.GetActive());
```
ulteq: This will break the skeleton extraction logic.

ulteq:
It's better now, but I still think that #1994 is the way to go here.

PR #1989, opened 2018-12-27
Download cotire.cmake to source directory instead

pkillnine:
This is a little friendlier to my particular package manager which doesn't allow downloading during the build process. (specifically, it can't copy cotire.cmake to the build directory because I need to create the build directory first in order to copy cotire.cmake into it, but if the build directory already exists then the package manager's build process fails because it tries to create a directory that already exists)

I don't think this change will break the build process in any way, and I think it makes more sense anyway to download something that is part of the source to the source directory.


tritonas00:
from discord:

>Cmake should not write any files to the source directory
 The source directory should be read-only

PR #1992, opened 2018-12-27
Add note that the build instructions are out of date for the dev build.

pkillnine:


ulteq:
I think we should just fix the build instructions -> https://github.com/RigsOfRods/rigs-of-rods/issues/2079

PR #1662, opened 2018-10-22
:sparkles: Added microprofile

AnotherFoxGuy:
Replaced high-performance-cplusplus-profiler with [microprofile](https://github.com/jonasmr/microprofile)

PR #2005, opened 2018-12-30
📐 Profiler folder removal

ulteq:
This should have already been done in: #1662


source/main/terrain/TerrainGeometryManager.cpp
```diff
@@ -223,6 +223,7 @@ float TerrainGeometryManager::getHeightAtTerrainPosition(Real x, Real y)
 
 float TerrainGeometryManager::getHeightAt(float x, float z)
 {
+    return 0.0f;
```
AnotherFoxGuy: Wut?

ulteq: Accident.

PR #2020, opened 2019-01-03
🔧 Disabled the warning texture by default

ulteq:


source/main/main.cpp
```diff
@@ -188,6 +188,15 @@ int main(int argc, char *argv[])
             Ogre::ResourceGroupManager::DEFAULT_RESOURCE_GROUP_NAME, Ogre::TEX_TYPE_2D, 512, 512, 0,
             Ogre::PF_R8G8B8, Ogre::TU_RENDERTARGET);
 
+        if (!App::diag_warning_texture.GetActive())
+        {
+            // We overwrite the default warning texture (yellow stripes) with something unobtrusive
+            Ogre::PixelBox pixels(1, 1, 1, Ogre::PF_R5G6B5);
```
paroj: with PF_BYTE_RGB, you could just use `uchar pixels[3] = {0};`. blitFromMemory does format conversion.

PR #2029, opened 2019-01-04
🎮 Reimplemented the cache regen logic

ulteq:
* Brings back the `-checkcache` command line argument
* Adds a new `Trigger cache update (and quit)` button to the (diag) settings

In an ideal world we would be able to update the cache without having to restart the game.


only-a-ptr:
> In an ideal world we would be able to update the cache without having to restart the game.

Ideal world? I'm on the way! https://github.com/RigsOfRods/rigs-of-rods/pull/2083

Issue #2058, opened 2019-01-11
Daf Skiploader Ropes no longer work


RustyMonster:
Would love to see this fixed, trying to update older content to more modern standards.

CuriousMike56:
Vehicle link: http://archives.rigsofrods.net/repo/files/repofiles-3rd-batch/e8941f35fa486db9b12bfccf78f61b0bcd516f90_skiploader.zip

The rope end nodes seem to move to node 0 at spawn, causing them to break.

PR #2062, opened 2019-01-11
🐛 Fixed the 'ropes' feature

ulteq:
fixed_a_bug
Fixes: https://github.com/RigsOfRods/rigs-of-rods/issues/2058


ulteq:
@only-a-ptr we really need to unify hooks / ties / ropes / ... at some point in time.

So much duplicate code. :man_facepalming:

PR #2068, opened 2019-01-13
Revert "🐛 Fixed material issues during truck reload (from disk)"

ulteq:
Reverts RigsOfRods/rigs-of-rods#2065 (which was a really bad idea)


ulteq:
I didn't want to revert the other commits. :thinking:

PR #2075, opened 2019-01-13
🔧 Fixed the wheel entries in the cache

ulteq:
Requires a full cache regen to take effect.


source/main/resources/CacheSystem.cpp
```diff
@@ -934,11 +933,38 @@ void CacheSystem::fillTruckDetailInfo(CacheEntry& entry, Ogre::DataStreamPtr str
...
+    for (const auto& w : def->root_module->flex_body_wheels)
+    {
+        entry.wheelcount++;
+        if (w.propulsion != RigDef::Wheels::PROPULSION_NONE)
+            entry.propwheelcount++;
+    }
+
+    if (!def->root_module->axles.empty())
+    {
+        entry.propwheelcount = static_cast<int>(def->root_module->axles.size() * 2);
```
only-a-ptr: Wheels can have both 'propulsion' attr set and be part of an axle, axle then overrides the attr, so this needs an extra check.

ulteq: `=`, not `+=` though :thinking:

ulteq: > Does anybody even read that value in the selector?

It was requested by the community.

PR #2083, opened 2019-01-15
[WIP] In-game cache regen (main menu only)

only-a-ptr:
**Test thoroughly, I coded it at 3AM!**

Settings panel / Diagnostics tab now contains button "Regen cache in-game" which forces an incremental check, but performs full regen if needed (i.e. user manually deletes cache files).

Not elegant in code but very convenient.

### DIFF
--- a/source/main/Application.cpp
+++ b/source/main/Application.cpp
@@ -86,7 +86,7 @@ static TerrainManager*  g_sim_terrain;
  GVarStr_A<100>           app_rendersys_override  ("app_rendersys_override",  "Render system",             "");
  GVarStr_A<300>           app_extra_mod_path      ("app_extra_mod_path",      "Extra mod path",            "");
  GVarPod_A<bool>          app_force_cache_udpate  ("app_force_cache_udpate",  "Force cache update",        false);
-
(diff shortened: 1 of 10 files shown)


tritonas00:
Yes

ulteq:
Could we do the same with the 'clear cache' button?

ulteq:
> Yes, but what should it do exactly?

Delete `mod.cache` -> Restart Ogre (or RoR) -> Recreate `mod.cache`

only-a-ptr:
@tritonas00: I kept it simple, just display the window (no updates), do the work and hide it again, most of the time it will just take a split second anyway but may also stall for a few seconds. Updating progressbar would just slow it down a bit.

> @ulteq It should basically replicate a game restart after deleting the mod.cache:

That's a definition of a full regen - should there be  [incremental update] [full update] buttons  then?

ulteq:
> should there be [incremental update] [full update] buttons then?

Yes.

tritonas00:
>Eventually I'll create a nicer and smarter IMGUI progress window.

that would be nice

PR #2089, opened 2019-01-18
✨ Basic savegame functionality (with quick- & autosave)

ulteq:
Adds:
* One `autosave` slot (allows you to resume the previous session from main menu)
* One `quicksave` slot slot per terrain
* Five global `savegame` slots (can also be used to quick access your favorite terrain)
  * Which support automatic terrain switching
  * Whose shortcuts are also available in main menu
* New `-resume` command line argument (automatically restores the last `autosave`)
  * Can be combined with the `-map` command line argument (to restore the `quicksave`)

Removes:
* The 'truck local' position storage feature (which only saved the node positions)


ulteq:
temp

PR #2107, opened 2019-01-22
🔧 Improved the collision avoidance performance

ulteq:


source/main/physics/Beam.cpp
```diff
@@ -927,11 +920,8 @@ Vector3 Actor::calculateCollisionOffset(Vector3 direction)
                         nb->AbsPosition - collision_offset,
                         actor->ar_collision_range * 3.0f);
 
-                    if (!m_intra_point_col_detector->hit_list.empty())
-                    {
-                        collision = true;
+                    if (collision = !m_intra_point_col_detector->hit_list.empty())
```
only-a-ptr: I don't like this sort of C-gymnastics. Does it affect performance at all (since it's in a 'performance' commit)?

ulteq: It doesn't affect the performance.

ulteq:
> Which of these changes actually make difference in performance?

Moving the calls to `UpdateIntraPoint` and `UpdateInterPoint` out of `calculateCollisionOffset`.

And using the result of previous `calculateCollisionOffset` calls as upper threshold for the following `calculateCollisionOffset` calls.

The combination of the two makes the call to `resolveCollisions` up to 70 times faster.

PR #2147, opened 2019-01-28
🔧 SkyX improvements

AnotherFoxGuy:
* Proper names for the light objects 
* Improved sun and moonlight colours 
* Added Time of day slider for SkyX

### DIFF
--- a/resources/SkyX/SkyXDefault.skx
+++ b/resources/SkyX/SkyXDefault.skx
@@ -4,7 +4,7 @@
 SkyXVersion=0.4.0
 
 #Main options field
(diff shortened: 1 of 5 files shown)


tritonas00:
Nice! Can you do something about those clouds?
cl

It "fixes" itself when you start driving

ulteq:
I would move the `USE_CAELUM` check into the first if block: https://github.com/RigsOfRods/rigs-of-rods/blob/d2acf9bf1dde12245f3740c1fcf4f4ac97e32c85/source/main/gui/panels/GUI_TopMenubar.cpp#L426-L447

PR #2167, opened 2019-01-31
🐛 Fixed cache write failure on 'Inf' floats

ulteq:


ulteq:
@AnotherFoxGuy Any idea what I do wrong?

source/main/resources/CacheSystem.cpp
```diff
@@ -53,7 +53,7 @@
 #include <rapidjson/document.h>
 #include <rapidjson/istreamwrapper.h>
 #include <rapidjson/ostreamwrapper.h>
-#include <rapidjson/prettywriter.h>
+#include <rapidjson/writer.h>
```
only-a-ptr: @ulteq Why this? Performance / file size?

PR #2184, opened 2019-02-02
🔧 Subtle mouse-drag force scaling

ulteq:


source/main/physics/Beam.cpp
```diff
@@ -1202,7 +1202,7 @@ void Actor::ResetPosition(Vector3 target)
 void Actor::HandleMouseMove(int node, Vector3 pos, float force)
 {
     m_mouse_grab_node = node;
-    m_mouse_grab_move_force = force;
+    m_mouse_grab_move_force = force * std::pow(m_total_mass / 3000.0f, 0.75f);
```
only-a-ptr: Are you sure it's subtle? Looks like quite heavy scale-down to me.

ulteq: Y-Axis: Scale
X-Axis: Mass in kg
temp

only-a-ptr: My math is worse than I feared. 😆

PR #2210, opened 2019-02-07
🔧 Various networking reverts

ulteq:
An attempt to reduce the network latency by reverting some of the less well thought out networking logic to how it used to be.


ulteq:
https://github.com/RigsOfRods/rigs-of-rods/commit/473fa7269bffbb54f95143333ddf793faceaacde did the trick.

PR #2225, opened 2019-02-10
🐛 Fixed duplicate zips in subfolders

AnotherFoxGuy:


source/main/resources/CacheSystem.cpp
```diff
@@ -969,6 +969,9 @@ void CacheSystem::ParseZipArchives(String group)
     int i = 0, count = files->size();
     for (const auto& file : *files)
     {
+		if (file.filename.find ('/') != std::string::npos)
```
ulteq: Try:
```
if (!file.path.empty())
```

PR #2229, opened 2019-02-10
🐛 Fixed bad resource handling of subdirectories

ulteq:
* Prevents duplicate cache entries when having zip archives in subdirectories
* Fixes support for content in subdirectories of zip archives

Closes: https://github.com/RigsOfRods/rigs-of-rods/pull/2225


ulteq:
And it makes the `file.path.empty()` check (https://github.com/RigsOfRods/rigs-of-rods/commit/d72faf6743d19568d0343e782d15c3595262a585) obsolete, which broke support for content in subdirectories of zip archives.

PR #2246, opened 2019-02-13
🎮 Resurrected the race results submission

ulteq:


source/main/physics/Savegame.cpp
```diff
@@ -38,9 +38,6 @@
 #include "SkyManager.h"
 #include "TerrainManager.h"
 
-#ifdef Bool // Conflicts with RapidJSON, see https://github.com/Tencent/rapidjson/issues/628
```
only-a-ptr: https://github.com/RigsOfRods/rigs-of-rods/commit/8f4eee3db8e7be8c6fe10222e3b203ee39bf5d2f
Why did you remove it?

ulteq: It seems to only be needed when you use the `PrettyWriter`, which we don't use anymore, due to other issues with it.

PR #2279, opened 2019-02-18
🔧 Simplified invalid actor logic and revamped actor spawn info

ulteq:
We can just delete remote trucks the moment we detect that they're no compatible with what we have installed locally.

| Missing content | Content mismatch |
| :-: | :-: |
| temp2 | temp1 |


CuriousMike56:
Would it be possible to log in the chat if a remote vehicle failed to load?

ulteq:
Possible yes, easy no.

PR #2276, opened 2019-02-17
Truckfile now supports multiple `minimass` sections.

only-a-ptr:
image

Test load: 
[MultiMiniMass_Test_Load.zip](https://github.com/RigsOfRods/rigs-of-rods/files/2875354/MultiMiniMass_Test_Load.zip)

Section minimass got options:
```
; option n = filler, does nothing
minimass
10.0, n

; option d = sets new default value, like `set_node_defaults`
minimass
4.0, d
```

Issue #2255, opened 2019-02-15
Issues with defining node mass


DODGEcharger:
I'm currently working on a project that makes use of lifter's Flexbodywheels section, one quirk of this section is that the wheel rim nodes are not included in the defined wheel mass, rather their mass is controlled directly by the minimass setting. Whether this is intentional or not, I'm not sure, but it has proved useful for tuning, so perhaps it's an expected behavior.

However I'm running into an issue where I need to make nodes within the body structure lighter than those defined by the minimass, but since the minimass is a global setting, the node masses are reset on spawn.

Possible solutions would be allowing "L" defined node masses to be lighter than the minimass, or making minimass a non-global setting like set_beam_defaults, where the settings are only applied until the next tag is reached.

CuriousMike56:
Fully agree with allowing multiple minimass sections. Would make weight tuning easier.

ulteq:
> allowing "L" defined node masses to be lighter than the minimass

> multiple minimass sections

How would we solve possible backwards compatibility issues?

only-a-ptr:
Suggestion:
```
; global, can be overriden with set_beam_defaults
minimass

; to reset minimass to global, use minimass=-1
set beam defaults -1,-1,-1,-1, minimass=1.23
```
I already coded the parser extension which supports these NAME=VALUE optional parameters, it's a part of https://github.com/RigsOfRods/rigs-of-rods/pull/1415/commits/fd4763ea1ddf3c5a4fdb5fc386acc0490063f30f

DODGEcharger:
> > allowing "L" defined node masses to be lighter than the minimass
> 
> > multiple minimass sections
> 
> How would we solve possible backwards compatibility issues?

Any previous mods wouldn't be affected, since what we're talking about was never allowed.

only-a-ptr:
* _"L nodes" lighter than minimass_ should be backwards compatible, eventually just breaking mods which had accidentally already set lighter "L" and relied on the auto-clamping. We could add a new parameter `allow_lighter: true/false` to `minimass`.
* _multiple `minimass` sections_ is not possible since `minimass` is position-independent, it could be defined at the end of truckfile and still affect everything. My parser rewrite silently lifted all the "must be defined before/after" rules. We could introduce a new directive like `set_node_default_minimass`

ulteq:
How about `minimass2`? Simple solution for all backwards compatibility issues.

DODGEcharger:
[...]

PR #2289, opened 2019-02-20
Truckfile: improved minimass handling

only-a-ptr:
* Added directive `set_default_minimass 1.23` which replaces `minimass` with `d` flag. The directive can be used in-between defining nodes, like `set_node_defaults`.
* Added `minimass` option `l` which will disable the minimass check for all nodes with 'l' (load weight) option. Suggested by @DODgecharger: https://github.com/RigsOfRods/rigs-of-rods/issues/2255#issue-410565608; https://github.com/RigsOfRods/rigs-of-rods/pull/2276#issuecomment-465442036 (at bottom)

[SetDefaultMinimass_TestLoad.zip](https://github.com/RigsOfRods/rigs-of-rods/files/2885636/SetDefaultMinimass_TestLoad.zip)

### DIFF
--- a/source/main/physics/Beam.cpp
+++ b/source/main/physics/Beam.cpp
@@ -724,9 +724,9 @@ void Actor::RecalculateNodeMasses(Real total)
     //update mass
     for (int i = 0; i < ar_num_nodes; i++)
     {
-        //LOG("Nodemass "+TOSTRING(i)+"-"+TOSTRING(ar_nodes[i].mass));
-        //for stability
-        if (!ar_nodes[i].nd_tyre_node && ar_nodes[i].mass < ar_minimass[i])
+        if (!ar_nodes[i].nd_tyre_node &&
+            !(m_definition->minimass_skip_loaded_nodes && ar_nodes[i].nd_loaded_mass) &&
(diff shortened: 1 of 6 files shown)


ulteq:
It feels wrong that the `l` flag "escapes" the scope of `d` flag `minimass` directives.

only-a-ptr:
> It feels wrong that the `l` flag "escapes" the scope of `d` flag `minimass` directives.

To begin with, it feels wrong that a section (`minimass`) should also do work of a directive (option `d`).
But such is the community request, so I'm not getting smart about it.

ulteq:
Maybe we should have used `set_mass_defaults` instead of going with the options?

only-a-ptr:
> Agreed, but we only do it like that to maintain backwards compatibility.

I don't believe so, we could have just introduced i.e. `set_default_minimass 1.23` and we'd be fine, too. I thought the smaller syntax change, the better. I'm not a modder and community requests are ad-hoc by nature, so how can I future-proof?

ulteq:
`set_default_minimass 1.23` sounds perfect to me.

ulteq:
Seems to work great! Better than master, when it comes to backwards compatibility.

PR #2298, opened 2019-02-25
✨ Node / Beam tuning utility

ulteq:
The code is relatively ugly, but the feature is so amazing that we should have it in 0.4.8.

Features:
- Live editing of actor mass, and beams / shocks / wheels springiness and damping.
- Automated search for optimal springiness and damping values.

temp


CuriousMike56:
That's awesome! Does it allow you to adjust individual shocks/wheels?

ulteq:
> Does it allow you to adjust individual shocks/wheels?

Not yet.

PR #2307, opened 2019-03-02
do not crash in ScriptEngine::addFunction when CompileFunction fails

paroj:


ulteq:
Can't we just replace all six `sprintf` calls with `Ogre::StringUtil::format`?

only-a-ptr:
> Should we not just replace all six sprintf calls with Ogre::StringUtil::format?

I'll bring in `libfmt` soon enough.

PR #2310, opened 2019-03-03
Revert "📐 Removed CrashRpt leftovers"

ulteq:
This reverts commit cba1035cec49874d3dbecd56f57847246510f4ae.

No idea how this works, but people seem to want it back.

### DIFF
--- a/DEPENDENCIES.md
+++ b/DEPENDENCIES.md
@@ -15,6 +15,7 @@ Rigs of Rods uses 3rd party libraries licensed under the following licenses:
 | curl                            | online services                                      | MIT/X derivative           |
 | pthreads                        | Threads                                              | LGPLv2.1                   |
 | UTFCpp                          | UTF-8 validation/conversion                          | Boost Software License 1.0 |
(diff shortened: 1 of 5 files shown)


AnotherFoxGuy:
CrashRpt doesn't work anymore with VS 2017, the last version it supported was 2013

only-a-ptr:
Might be a good alternative: https://chromium.googlesource.com/breakpad/breakpad/+/master/docs/getting_started_with_breakpad.md

AnotherFoxGuy:
Or maybe https://www.bugsplat.com ?

https://www.bugsplat.com/docs/platforms/cplusplus

only-a-ptr:
Bugsplat doesn't have non-commercial plan: https://www.bugsplat.com/pricing

ulteq:
Or we simply leave everything as is? I don't really see the point in having something like `CrashRpt` to be honest.

only-a-ptr:
There's never enough stack traces when reporting a bug, but we manage very well without.

I'd leave everything as-is for now and look into Breakpad later.

PR #2312, opened 2019-03-05
AngelScript - a lot to clean up

only-a-ptr:
## Status

About half the script events is dead.

## To discuss

I don't like the callback mechanism:
 * events must be bitmasks (to enable filtering) which limits us to 64 events (longer bitmasks are possible but insane)
 * There are 2 distinct kinds of events: 1. the collision box entry 2. all the others. See notes below.
 * Events invoke the handler immediately in various circumstances -> bad for maintenance, prone to mysterious bugs

## Notes

```
// **** crash course **** //
You must not define function twice (same name and parameters) in a single script (also counting those it #indclude-s) 
If you do, loading script will fail with Error = A function with the same name and parameters already exists


// ***** built-in includes ****** //

// Defines a basic 'defaultEventCallback()' function, if you don't want to define your own.
// NOTE: loaded automatically if terrain has no custom script
#include "base.as" 

// Only #include-s base.as and defines dummy "main()"
// NOTE Loaded automatically if terrain has no custom script
#include "default.as"                      


// ***** built-in function names ****** //

// Required (in each individual script), otherwise loading the script fails.
// Invoked immediately after script is loaded
void main()
{
	
}

// Optional
// a fallback handler for collision-box events, if user didn't specify one.
// NOTE: those can be only specified when adding object from script using "game.spawnObject()"
// Each new definition (in case of multiple script files) overrides the previous
// param #0 "trigger_type" is always 0
// param #1 "inst" - defined in ODEF using "event [eventname] [filter]" line
// param #2 "box" - defined in AngelScript using "game.spawnObject()" parameter "instancename"
// param #3 "nodeid" - index of the softbody node which triggered the event, or 0xFFFFFFFF if not available
void void defaultEventCallback(int trigger_type, string inst, string box, int nodeid)
{
    
}



// The default RoR event callback
// see docs on enum "scriptEvents"
void eventCallback(int eventnum, int value)
{
    
}
```

### DIFF
--- a/doc/angelscript/global_functions.h
+++ b/doc/angelscript/global_functions.h
@@ -43,7 +43,6 @@ void print(const string message);
  	SE_TRUCK_ENGINE_FIRE               = 0x00000020, //!< triggered when the planes engines start to get on fire, the argument refers to the truck number
  	SE_TRUCK_TOUCHED_WATER             = 0x00000040, //!< triggered when any part of the truck touches water, the argument refers to the truck number
  	SE_TRUCK_LIGHT_TOGGLE              = 0x00000400, //!< triggered when the main light is toggled, the argument refers to the truck number
- 	SE_TRUCK_SKELETON_TOGGLE           = 0x00000800, //!< triggered when the user enters or exits skeleton mode, the argument refers to the truck number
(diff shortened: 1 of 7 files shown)


only-a-ptr:
Closing as outdated, to be continued eventually...

Issue #2243, opened 2019-02-12
Animators for transfer cases


ulteq:
Something comparable to the `difflock` animator we already have for differentials.

Reference: https://docs.rigsofrods.org/vehicle-creation/fileformat-truck/#animators

only-a-ptr:
`difflock` animates like this: https://github.com/RigsOfRods/rigs-of-rods/blob/3455f533241df966008d6c6524c2af15c9e94cbd/source/main/physics/Beam.cpp#L1964-L1970

How should `transfercase` animate? Pass-thru the current gear ratio?

ulteq:
> How should transfer case animate?

I don't know. But it won't be easy. Especially when we also want to consider gear ratios.

only-a-ptr:
> I don't know. But it won't be easy

I don't see why, there aren't so many variables to visualize, just mode and gear ratio IMO.
```
source: tcase_mode -> 1.0 (2wd Lo) / 2.0 (2wd Hi) / 4.0 (4wd)
source: tcase_gear ->VALUE of current gear (0.0 if not applicable)
```

Mr-The-Ghost:
What if it was displayed with the transfer case mode then gear ratio. 

2wd lo -gear ratio
2wd hi-gear ratio

Something of that nature.

PR #2328, opened 2019-03-13
Fixed re-loading [flexbody/mesh]wheels

only-a-ptr:
Introduced by https://github.com/RigsOfRods/rigs-of-rods/issues/2243
Kudos to @CuriousMike56 for reporting and @tritonas00 for investigating.

### DIFF
--- a/source/main/physics/flex/FlexFactory.cpp
+++ b/source/main/physics/flex/FlexFactory.cpp
@@ -118,9 +118,11 @@ FlexMeshWheel* FlexFactory::CreateFlexMeshWheel(
     std::string const & rim_mesh_name,
     std::string const & tire_material_name)
 {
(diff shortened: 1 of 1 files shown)


CuriousMike56:
Also happens with the Viper skinzip:
```
Ogre::FileNotFoundException::FileNotFoundException: Cannot locate resource DodgeViperWheel_Edge.png in resource group Mod-1...
Ogre::FileNotFoundException::FileNotFoundException: Cannot locate resource DodgeViperBody_Serpent.png in resource group Mod-1...
```
image

tritonas00:
>I figure what's wrong with the skinzips: they're loaded into different resource group than the mod, that's why the textures aren't found.

only-a-ptr:
I'll open another PR for the skinzips, they needed more work.
Viper has some missing/misplaced textures, CuriousMike is working on it.

PR #2336, opened 2019-03-28
Updated icon

CuriousMike56:
Updated the executable icon with @DODGEcharger's new one, now with embedded sizes:
image


only-a-ptr:
The icon is semi-transparent, and I think it needs more opacity or at least a more visible outline.
This is how it looks on my desktop:
image

CuriousMike56:
How about this?
image

only-a-ptr:
That's more like it!

PR #2359, opened 2019-05-10
Updated icon with blue outline + added input maps

CuriousMike56:
- Icon now has a blue outline for better visibility and to be consistent with the other icons:
image
- Added input maps for the G29 and Xbox controllers (G29 input map by @DODGEcharger)
- Fixed G27 input map name


only-a-ptr:
@CuriousMike56 Consistency with which icons? Both https://www.rigsofrods.org/ and https://docs.rigsofrods.org/ have red outlines. Also, I don't like how the blue outline fades into the transparent gray-blue background, it's hard to make out on light-gray background.

CuriousMike56:
The icon shown on the README and the [in-game icon](https://github.com/RigsOfRods/rigs-of-rods/blob/master/resources/mygui/rorlogo.png) have blue outlines. The forum and docs icons are red because of the blue background.

PR #2375, opened 2019-06-10
🔧 Load debug plugins in debug build

AnotherFoxGuy:


only-a-ptr:
@AnotherFoxGuy :+1: Thanks!

PR #2382, opened 2019-06-29
Moved direction arrow back to it's overlay

tritonas00:
- Restores original direction arrow position
- Reverts https://github.com/RigsOfRods/rigs-of-rods/commit/6afaa854b99ab80b24dbc16e52e6bbd14a54ed62

Back then ulteq struggled to bring back the direction arrow in it's [overlay](https://github.com/RigsOfRods/rigs-of-rods/blob/master/resources/overlays/directionarrow.overlay) so he came up with an alternative way https://github.com/RigsOfRods/rigs-of-rods/commit/6afaa854b99ab80b24dbc16e52e6bbd14a54ed62 (attached directly to vehicle) which was a bit annoying sometimes so he also added a switch to turn it off.

The problem was wrong `zorder` value in the overlay

### DIFF
--- a/resources/materials/ror.material
+++ b/resources/materials/ror.material
@@ -2418,7 +2418,9 @@ material tracks/directionArrowNormal
     {
         pass
         {
-			lighting off
+            lighting off
+            depth_check on
+            cull_hardware none 
             scene_blend alpha_blend
             texture_unit
             {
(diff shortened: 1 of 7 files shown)


only-a-ptr:
:+1:

PR #2394, opened 2019-07-22
👷 Upgrade travis to bionic 

AnotherFoxGuy:


tritonas00:
It would be better if you keep xenial, so the precompiled zip can run on a wider selection of distrros.

With this it will not run on anything earlier than bionic, probably.

Issue #2334, opened 2019-03-22
Incorrect actor spawning on some terrains


CuriousMike56:
Barges (20KLpont) on [BDM3-20Kleagues](https://forum.rigsofrods.org/resources/bdm3-20kleagues.58/):
![screenshot_2019-03-21_21-27-27_1](https://user-images.githubusercontent.com/46073351/54795060-07ccba00-4c21-11e9-970b-7caf289481a9.png)
[Penguinville with trolley line](http://forum.rigsofrods.org/resources/penguinville-with-trolley-line.87/)'s slidenode track:
![screenshot_2019-03-21_21-30-48_1](https://user-images.githubusercontent.com/46073351/54795126-52e6cd00-4c21-11e9-8e2d-718f3f514b18.png)

tritonas00:
Its caused by this commit https://github.com/RigsOfRods/rigs-of-rods/commit/c2dfa46e785a4f8803d3be207a85609be6026021

CuriousMike56:
Would there be any side effects if we just reverted that commit?

tritonas00:
that was @ulteq's hardest work, im sure there is a way to fix it

tritonas00:
Did some research. This https://github.com/RigsOfRods/rigs-of-rods/blob/9837679be3f248b1cd3237cf155ec21fc6338be0/source/main/physics/Beam.cpp#L1166-L1170 is needed to fix the trolley line and this https://github.com/RigsOfRods/rigs-of-rods/blob/9837679be3f248b1cd3237cf155ec21fc6338be0/source/main/physics/Beam.cpp#L1224-L1233 to fix the racemax.

So i think we need back those `// horizontal/vertical/total displacement` code and convert/place them here https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/physics/BeamFactory.cpp#L158

They removed by this commit https://github.com/RigsOfRods/rigs-of-rods/commit/c2dfa46e785a4f8803d3be207a85609be6026021

only-a-ptr:
@tritonas00 Thanks for the research, I guess Ulteq thought those transformations are not used by anyone anymore.

I sense we might need unit tests for these cases, to both check for regressions like this and to document what exactly is the expected behavior, including all the quirks.

Anyway, I'll check the Barges now.

only-a-ptr:
**BUMP!**

[...]

PR #2397, opened 2019-07-25
Created test harness (using GoogleTest)

only-a-ptr:
Hello RoR fans.

I'm proud to announce I've developed an environment for automated testing for RoR. This means it's possible, without much hassle, to create short programs which load a portion of RoR, feed it with some predefined data, run RoR and check the data are modified correctly. This provides 2 advantages:
1. Developer will quickly find out that he accidentally broke something (if it's an intentional change, s/he must modify the test).
2. Developer can read the test to find out what exactly are RoR features supposed to do - because figuring it out directly from the codebase can be quite hard.

Why now? Because we've just had a textbook example of a regression bug which could have been avoided by tests: #2334. It doesn't look like a big deal, but it's actually pretty sneaky:
1. It didn't affect trucks which have node#0 on position 0,0,0 - it's not required, but apparently some trucks are made that way (for example ScaniaWrecker on NHelens). This is too tricky for a human tester to figure out.
2. While the bug was in upstream, one of the affected maps (NHelens with misplaced bridge) actually got updated to curb the problem! I didn't know it and I actually thought there is another bug. So much for finding regressions using content.

And so, to preserve RoR's backward compatibility (and my sanity), I decided the time has come to accept some good programming manners and start testing. I've never written tests before, so I had to learn as I go - that's why all the test stuff comes in a single commit, the development history was simply too messy. Also, the spawn-positions test is quite convoluted because setting up RoR + Ogre without any resources or rendering was a challenge, and calculating the node positions for verifications is simply as complicated as the spawning itself. As soon as I add another test, I'll give it a proper 'fixture' shape.


tritonas00:
`-DROR_USE_GOOGLETEST=FALSE` builds fine, things spawn correctly now

`-DROR_USE_GOOGLETEST=TRUE` needs package gtest, installed but got

```[  9%] Building CXX object test/suites/UtilsTestSuite/CMakeFiles/UtilsTestSuite.dir/__/__/__/source/main/utils/SHA1.cpp.o
In file included from /home/babis/Downloads/ror-dependencies/rigs-of-rods/source/main/utils/Utils.h:30,
                 from /home/babis/Downloads/ror-dependencies/rigs-of-rods/source/main/utils/Utils.cpp:22:
/home/babis/Downloads/ror-dependencies/rigs-of-rods/source/main/RoRPrerequisites.h:24:10: fatal error: OgrePrerequisites.h: No such file or directory
   24 | #include <OgrePrerequisites.h>
      |          ^~~~~~~~~~~~~~~~~~~~~
compilation terminated.```

tritonas00:
i think its time to have https://github.com/RigsOfRods/rigs-of-rods/pull/2397/commits/6b9fe563d0788de7eda2849da23965cabdb66db2 on master please

tritonas00:
@only-a-ptr there is no conan package OIS https://github.com/RigsOfRods/rigs-of-rods/blob/master/conanfile.txt, ois lib is included in OGREdeps conan package

only-a-ptr:
@tritonas00 Thanks. That's all I can do for today 💤

only-a-ptr:
A linking error, Travis? I give up for today :zzz:

CuriousMike56:
Works fine on Windows 👍

only-a-ptr:
Closing as out of date, should be revisited later - automated testing is a good thing to have!

Issue #2344, opened 2019-04-06
Spawning vehicle with existing custom particle crashes


CuriousMike56:
- Download [DAF-95XF-4x2-semi-air-suspension.zip](https://github.com/RigsOfRods/rigs-of-rods/files/3050951/DAF-95XF-4x2-semi-air-suspension.zip) and [DAF-95XF-6x4-semi-air-suspension.zip](https://github.com/RigsOfRods/rigs-of-rods/files/3050952/DAF-95XF-6x4-semi-air-suspension.zip)

- Spawn both trucks, doesn't matter in what order

- Game will crash with this error:
![image](https://user-images.githubusercontent.com/46073351/55675305-a21e3600-588e-11e9-98ba-f4af04c34a1d.png)

only-a-ptr:
Until now I didn't have a mod with custom particles, thanks!

tritonas00:
Spawn [PittsSpecial.zip](https://github.com/RigsOfRods/rigs-of-rods/files/3144727/PittsSpecial.zip) twice. Press `g` on both. Particles work only on the one spawned first

```
22:20:29: ItemIdentityException: An object of type 'ParticleSystem' with name 'cparticle-0-PittsSpecial' already exists. in SceneManager::createMovableObject at /home/babis/Downloads/ror-dependencies/Source/ogre/OgreMain/src/OgreSceneManager.cpp (line 3894)
```

only-a-ptr:
Current upstream will just show ingame notification instead of crashing, but the error remains.

tritonas00:
https://github.com/RigsOfRods/rigs-of-rods/issues/2344#issuecomment-489357708 heavily affects multiplayer, only the first spawned actor can use particles (another example 2008 KME predator - use `g` to throw water)

tritonas00:
is now fixed on master

PR #2398, opened 2019-07-27
Fixed custom particle name conflicts between actors

only-a-ptr:
Turns out I already resolved the same issue for 'exhausts', but I didn't for 'particles'. Now I unified the setup for both, always using correct (actor-specific) OGRE resource group.

I'm not sure how to test, though. The [DAF-semi-air](https://github.com/RigsOfRods/rigs-of-rods/issues/2344#issue-430077406) trucks don't use 'particles' keyword, only 'exhausts'. The [PittsSpecial](https://github.com/RigsOfRods/rigs-of-rods/issues/2344#issuecomment-489357708) does use 'particles', but those don't seem to show up when toggled.

### DIFF
--- a/source/main/physics/RigSpawner.cpp
+++ b/source/main/physics/RigSpawner.cpp
@@ -1252,44 +1252,39 @@ void ActorSpawner::ProcessFixedNode(RigDef::Node::Ref node_ref)
 
 void ActorSpawner::ProcessExhaust(RigDef::Exhaust & def)
 {
-    // parse exhausts
     if (m_actor->m_disable_smoke)
     {
         return;
     }
 
-    node_t & ref_node = GetNodeOrThrow(def.reference_node);//id1;
-    node_t & dir_node = GetNodeOrThrow(def.direction_node);//id2;
-
     exhaust_t exhaust;
-    exhaust.emitterNode = ref_node.pos;
-    exhaust.directionNode = dir_node.pos;
+    exhaust.emitterNode   = this->GetNodeIndexOrThrow(def.reference_node);
+    exhaust.directionNode = this->GetNodeIndexOrThrow(def.direction_node);
 
-    Ogre::NameValuePairList params;
-    params["resourceGroup"] = m_custom_resource_group;
-    if (def.particle_name.empty() || def.particle_name == "default")
+    std::string template_name = def.particle_name;
+    if (template_name.empty() || template_name == "default")
     {
-        params["templateName"] = "tracks/Smoke"; // defined in `particles/smoke.particle`
-    }
-    else
-    {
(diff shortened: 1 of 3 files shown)


tritonas00:
>but those don't seem to show up when toggled.

They don't show up with this PR, master works

only-a-ptr:
PittsSpecial works fine now
image

only-a-ptr:
BUMP! The DAFs problem is trickier; it may need to be patched at OGRE level.

@paroj It looks like OGRE disregards resource groups when registering 'particle system templates': https://github.com/OGRECave/ogre/blob/8b4f28a7e61f75dc085d721d8ad56d8fb1dff679/OgreMain/src/OgreParticleSystemManager.cpp#L120. It does make sense, but for us, it causes name clashes in user-made content. What's your opinion?

tritonas00:
@only-a-ptr are you sure there is nothing wrong with https://github.com/RigsOfRods/rigs-of-rods/commit/3c15ab833769d290bc644ea96afb42de3697a92b? Because just once commit before, both DAFs spawn correctly

only-a-ptr:
@tritonas00 I can't test right now, but I'm pretty sure the problem already existed before 3c15ab8, but it didn't cause spawn failure - it only caused the second spawned DAF to use first DAF's 'Smoke.particle'.    I guess I'll just make it work that way for time being.

paroj:
> Apparently OGRE disregards resource groups when registering 'particle system templates': https://github.com/OGRECave/ogre/blob/8b4f28a7e61f75dc085d721d8ad56d8fb1dff679/OgreMain/src/OgreParticleSystemManager.cpp#L120. It does make sense, but for us, it causes name clashes in user-made content. What's your opinion?

urgh.. thats quite ugly. Particle templates should behave as resources, but we cannot break existing code during the 1.12 cycle.
It will take me 1-2 weeks until I can take a closer look at this though..

only-a-ptr:
@tritonas00 I added a workaround - particles with conflicting names are simply skipped, so the second DAF will use first DAF's particles - which hardly matters since it's the same file.

tritonas00:
```
Parsing script 3d_dashblend.overlay


Technical details below: 

ItemIdentityException: Could not find material tracks/dashblend in OverlayElement::setMaterialName at /home/babis/Downloads/ror-dependencies/Source/ogre/Components/Overlay/src/OgreOverlayElement.cpp (line 341
```

when loading terrain

only-a-ptr:
I fixed the GCC complaint (my fault), but the material problem is weird. Try rebuilding resources.

tritonas00:
actually it works fine now, tried both pitts and dafs :+1:

PR #2403, opened 2019-08-03
Fixed air dashboard showing previous vehicle's state

### LINKED ISSUE #1560: Transmission indicators freeze in some vehicles when going back to menu
~Load simple terrain and spawn this http://archives.rigsofrods.net/repo/files/repofiles-3rd-batch/df0ec67470bf500641d7f05f1a47cf0a7631b9ab_Mercedes-Benz-Actros-8x8-wrecker.zip~

~Switch to first person view (dashboard view) and drive: transmission indicators work fine~

~Press Back to menu, reload simple terrain and respawn it again: transmission indicators are stuck~

FIXED

only-a-ptr:
Fixes #1560

UPDATE: Original solution scrapped, see https://github.com/RigsOfRods/rigs-of-rods/pull/2403#issuecomment-518029318

### DIFF
--- a/doc/angelscript/BeamClass.h
+++ b/doc/angelscript/BeamClass.h
@@ -60,11 +60,6 @@ public:
 	void setReplayMode(bool rm);
 	
 	/**
-	 * Resets the auto-pilot.
-	 */
-	void resetAutopilot();
-	
-	/**
 	 * Toggles the custom particles.
 	 */
 	void toggleCustomParticles();
--- a/source/main/gameplay/AircraftSimulation.cpp
+++ b/source/main/gameplay/AircraftSimulation.cpp
@@ -40,7 +40,7 @@ void AircraftSimulation::UpdateInputEvents(Actor* vehicle, float seconds_since_l
     //autopilot
     if (vehicle->ar_autopilot && vehicle->ar_autopilot->wantsdisconnect)
     {
-        vehicle->disconnectAutopilot();
+        vehicle->ar_autopilot->disconnect();
     }
     //AIRPLANE KEYS
     float commandrate = 4.0;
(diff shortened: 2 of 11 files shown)


tritonas00:
Great but the reset values should be like the first spawn: GPWS ON and the four levers visible like this

kk

Instead its
kk

tritonas00:
For some strange reason even with https://github.com/RigsOfRods/rigs-of-rods/pull/2403/commits/35712b27ae69c0649097688334253a9f4faf9489 sliders still disappear :confused:

only-a-ptr:
Updated and tested. I had to open C++ reference manual to find the problem :smile:

tritonas00:
:+1: 

Any idea why
> spawn airplaine-> turn off GPWS -> backspace: GPWS becomes on as it should but it takes two presses to turn it off

?

only-a-ptr:
Doh! That's because I'm just resetting the button texture to ON but apparently the vehicle remembers the state.

tritonas00:
> All the issues are valid again.

Not all, all four throttle sliders are now resetted into the correct position :+1:

only-a-ptr:
Finished. Whole aero dashboard is now updated from SimBuffer on every render, and thus impervious to the lingering outdated values.

tritonas00:
Works like a charm!

tritonas00:
one minor issue GPWS gets off when you reset (backspace or i)

only-a-ptr:
Updated. Turned out the ad-hoc updating of the overlays was a real cancer 🤢 

The backspace (COMMON_REPAIR_TRUCK) works as intended - it keeps the autopilot as-is.

The reset-to-initial-pos (COMMON_RESET_TRUCK) was buggy - it updated the dashboard GUI directly, but not the actual state. Fixed.

tritonas00:
works now!

other issue: this section is wrongly updated comparing master and this PR
kk

just spawn the antonov12 or the skyvan start the engines and raise throttle levers

tritonas00:
now its epic :heart:

PR #2405, opened 2019-08-05
Fixed switching actors on multiplayer

tritonas00:
**Issue**

Spawn the same vehicle twice or more in multiplayer, go to `Vehicles` drop down menu and try to switch between them by clicking on the entries. Only the first spawned vehicle works, others not.

**Problem**

Entries must have a unique name like in single player

**Solution**

Added hidden numbers and user uniqueid on MP actor list. Now each entry name is unique and you can switch between all actors.


only-a-ptr:
Hi, good spot!
And yes, there's actually a more elegant solution: https://github.com/ocornut/imgui/blob/62143dff64c8cfb960248b84cf6c4566ffc5a743/imgui.cpp#L742 - I'd use the leading numbers from your patch as the hidden ID part.

tritonas00:
Thank you, indeed!

Issue #2278, opened 2019-02-18
Regression bug? Zoom after play game


KinG-InFeT:
Hi all,
i try game from itch on my fedora 29 system and after play game the zoom is wrong
![image](https://user-images.githubusercontent.com/193969/52961914-49591380-339c-11e9-9c63-8c6ae7b476e2.png)

It seems that at the first start you are wrong to set the size of the starting screen, but in the other pages remains that resolution.
![image](https://user-images.githubusercontent.com/193969/52961938-57a72f80-339c-11e9-93da-4337f925e0c2.png)

My laptop has 2 screens (main laptop) and a monitor in HDMI connected. can this be the problem?

i'm here for any test

CuriousMike56:
This issue was fixed when MyGUI was reverted to an earlier commit. Perhaps RC4 was built with a newer MyGUI by mistake? @AnotherFoxGuy

KinG-InFeT:
other screen after run game (in game)
![image](https://user-images.githubusercontent.com/193969/54020706-b0453d80-418e-11e9-89d2-76b293757b69.png)

tritonas00:
@AnotherFoxGuy did you tag mygui to `07da0620e0d835346452b02b7b1ba7c583429736`?

KinG-InFeT:
i confirm this issue with latest version (Edge version from itch)

tritonas00:
@KinG-InFeT it should be fixed on the latest dev build
![kk](https://user-images.githubusercontent.com/2660424/56383263-21fab780-6222-11e9-88bb-0084773b96d3.png)

KinG-InFeT:
hi @tritonas00 i try with latest dev release and confirm this bug:
![image](https://user-images.githubusercontent.com/193969/58811958-1b34d200-8621-11e9-8805-6d4b07584921.png)
![image](https://user-images.githubusercontent.com/193969/58811972-2425a380-8621-11e9-9907-a1b1f45b7db1.png)
![image](https://user-images.githubusercontent.com/193969/58812014-39023700-8621-11e9-9078-177e84c42305.png)

tritonas00:
did you resize the window manually?

KinG-InFeT:
ror starts in a small window (like in screen) then I try to enlarge the window but in reality the content remains "small" I will try to make a video to try to explain better.

CuriousMike56:
You can't resize the window manually. You must set the window size by going to Settings -> Render System -> Video Mode.

[...]

PR #2413, opened 2019-08-12
🔧 Set window border to fixed

AnotherFoxGuy:
This fixes the issue with players resizing the game window and breaking the GUI


tritonas00:
Is it possible to set minimum resolution to 800 x 600 also to prevent this https://github.com/RigsOfRods/rigs-of-rods/issues/2278#issuecomment-498992772 ?

tritonas00:
hacky but works (because you can select resolutions below 800x600 and the result will be 800x600)

another option would be to set the initial resolution (when you start RoR for first time) to 800x600

Issue #2419, opened 2019-08-19
Client should send a hashed token to race API


Zentro:
This should be simple.
https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/scripting/GameScript.cpp#L791

It should be sending a hashed user token, but it's not.

only-a-ptr:
This is what currently happens:
1. User inputs the naked token to RoR.cfg - loaded as `mp_player_token`
2. User may change `mp_player_token` via MultiplayerSelector GUI
3. When connecting, RoR sends naked token as part of `MSG2_USER_INFO`
4. Script authenticates messages using naked token

I don't know how it originally worked, but since we don't encrypt the communication (do we?), transmitting the token as hashed or naked makes little difference IMO.

tritonas00:
> should RoR.cfg contain the naked or hashed token? 

naked, it is user input

ror does the hashing and sends it

PR #2420, opened 2019-08-19
Bugfix: send hashed user token to server API

only-a-ptr:
See https://github.com/RigsOfRods/rigs-of-rods/issues/2419


tritonas00:
works fine

PR #2421, opened 2019-08-21
Ressurected 'hookgroup' in Truckfile

only-a-ptr:
I just noticed we have code for 'ropable-only hooks', but no way to setup them. I looked in 0.3x sources and found out 'hookgroup' which got left out from the new parser.

I re-created the section as best as I could, but I don't know any vehicles to test, hence the 'draft PR'.

### DIFF
--- a/source/main/physics/RigSpawner.cpp
+++ b/source/main/physics/RigSpawner.cpp
@@ -3029,6 +3029,30 @@ void ActorSpawner::ProcessHook(RigDef::Hook & def)
     }
 }
 
+void ActorSpawner::ProcessHookGroup(RigDef::HookGroup & def)
+{
+    hook_t hook;
+    hook.hk_locked       = UNLOCKED;
+    hook.hk_group        = def.id;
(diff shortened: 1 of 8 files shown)


tritonas00:
i think this uses hookgroups

https://forum.rigsofrods.org/attachments/bbd-ltm-11200-9-2-zip.6155/

CuriousMike56:
Hookgroups are a part of the [hooks](https://docs.rigsofrods.org/vehicle-creation/fileformat-truck/#hooks) section. What would be the use case for a separate section?

only-a-ptr:
> I've never seen a vehicle that uses a separate hookgroups section.

That may be the reason why it was never ported. I only noticed it because it's the only way to limit `hooks` to catch pre-defined nodes instead of all nodes, which seems like an important feature to me. 

But now that I read the docs, it looks like code doesn't match... I need to take another look

only-a-ptr:
Abandoned in favor of #2423

PR #2422, opened 2019-08-23
✨ Switched from semver to rolling releases

### LINKED ISSUE #2415: Switch from semver to rolling releases
Currently, we use the semver release format (0.38, 0.4)
I think it would be better to switch to date based, so we can release a new version when we think that it is stable enough
The reason I think it would be better is that the last "official" stable release was back in 2016 and that we now use the release candidates a stable releases 

Example of the new release format:

Old | New
---|---
0.37.126 | 2010-12
0.38.64 | 2011-10
0.4.5.0 | 2015-08
0.4.7.0 | 2016-11
0.4.8 RC5 | 2019-07

Thoughts?

AnotherFoxGuy:
Closes #2415

**Changes**
* CMake will now automatically generate a version number based on the date of the build
* Moved the user dir on windows to `My Games/Rigs of Rods`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.5.1)
+cmake_minimum_required(VERSION 3.8.0)
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_SOURCE_DIR}/cmake ${CMAKE_SOURCE_DIR}/cmake/3rdparty)
 
 include(CMakeDependentOption)
@@ -12,7 +12,7 @@ set(PROJECT_NAME_SHORT "RoR")
 set(PROJECT_NAME_UNDERSCORE "Rigs_of_Rods")
 set(PROJECT_NAME_LONG "Rigs of Rods")
 
-project(${PROJECT_NAME_UNDERSCORE} VERSION 0.4.8.0)
+project(${PROJECT_NAME_UNDERSCORE})
 
 ################################################################################
 # Available build options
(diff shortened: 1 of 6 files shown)


only-a-ptr:
Pardon the delay, I'll test tonight.

However, question remains: Merge this right now (and abandon the 0.4.8 release) or first release 0.4.8 because it's been announced for so long? I sort of prefer "right now" because the denouncement of 0.4.8 would provide publicity to the new versioning scheme.

CuriousMike56:
Is it possible to add a message telling users about the user folder change on startup if it detects `Documents\Rigs of Rods 0.4`? Some don't use the installer (e.g. installing through Itch.io) and would be confused as to why their mods no longer appear.

only-a-ptr:
> @CuriousMike56 add a message telling users about the user folder change on startup

Sure, I'm thinking a custom IMGUI panel in main menu. I'll create a demo tonight.

tritonas00:
maybe you should get rid of the -dev suffix also?

AnotherFoxGuy:
> Maybe you should get rid of the `-dev` suffix also? 

The dev suffix is there to indicate it is a development build
> Where is that come from?

https://github.com/RigsOfRods/rigs-of-rods/blob/master/cmake/GenerateVersionInfo.cmake#L91

AnotherFoxGuy:
> @only-a-ptr I found a windows bug

This should be fixed in https://github.com/RigsOfRods/rigs-of-rods/pull/2422/commits/40806ace3d99612a2a98e583c1c1cc4e67eef3a3

> add a message telling users about the user folder change on startup

Added in https://github.com/RigsOfRods/rigs-of-rods/pull/2422/commits/ff2285b1edaa0410cbc0b1f7979799524d715758

AnotherFoxGuy:
Ive made the message a litebit nicer and removed `std::stream`

PR #2423, opened 2019-08-23
Removed dead hook flag 'lock nodes'

only-a-ptr:
This flag makes hooks lock only to ropables instead of all nodes (default). It's been non-configurable for years and nobody noticed. Also, you can already define hook-able nodes using `lockgroups` and `lockgroup_default_nolock`

This is a diabolical PR to https://github.com/RigsOfRods/rigs-of-rods/pull/2421

### DIFF
--- a/source/main/physics/Beam.cpp
+++ b/source/main/physics/Beam.cpp
@@ -3670,74 +3670,39 @@ void Actor::ToggleHooks(int group, hook_states mode, int node_number)
                 if (this == actor && !it->hk_selflock)
                     continue; // don't lock to self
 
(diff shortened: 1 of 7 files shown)


only-a-ptr:
@CuriousMike56 I inverted a condition 😊 https://github.com/RigsOfRods/rigs-of-rods/pull/2423/commits/7ce852333aa86d4a329071803e34bb79f146cdcc#diff-53cb603e59ebba81abd7871dabda891bL3004-R3004

CuriousMike56:
Everything seems to be fine now 👍

PR #2425, opened 2019-08-26
Load 'inertia_models.cfg' just once.

only-a-ptr:
I noticed we're loading and parsing it three times when spawning any actor - for no reason (it's not configurable per actor, there's only one global config file).

UPDATE: With the last commit there's also minor performance improvement:
image


tritonas00:
[SST-Shocks3.truck.zip](https://github.com/RigsOfRods/rigs-of-rods/files/3542034/SST-Shocks3.truck.zip)

uses `set_inertia_defaults` and segfaults with this PR

only-a-ptr:
@tritonas00 Fixed, turns out I misunderstood how it works without inertia assigned.

tritonas00:
no crashes now :+1:

PR #2438, opened 2019-09-08
Fixed unzipping mods causing materials not to load

### LINKED ISSUE #523: Unzipping terrain causes materials not to load

AnotherFoxGuy:
Fixes #523

### DIFF
--- a/source/main/resources/CacheSystem.cpp
+++ b/source/main/resources/CacheSystem.cpp
@@ -1139,7 +1139,8 @@ void CacheSystem::LoadResource(CacheEntry& t)
     }
 
     static int rg_counter = 0;
-    String group = "Mod-" + std::to_string(rg_counter++);
+    String group = std::to_string(rg_counter++) + "-" + t.fname;
+
     try
     {
         if (t.fext == "terrn2")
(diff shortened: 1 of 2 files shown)


source/main/resources/CacheSystem.cpp
```diff
@@ -1106,7 +1106,9 @@ bool CacheSystem::CheckResourceLoaded(Ogre::String & filename, Ogre::String& gro
     if (ResourceGroupManager::getSingleton().resourceExistsInAnyGroup(filename))
     {
         group = ResourceGroupManager::getSingleton().findGroupContainingResource(filename);
-        return true;
+        // Files are not allowed to be loaded from the Content resource group, except odef files
+        if(group != "Content" || filename.find("odef") != std::string::npos)
```
only-a-ptr: Umm, neat hack.

I can't recall why we have both CONTENT_RG "Content" + TEMP_RG "Temp" while both are only used during cache regen. I think CONTENT_RG should be destroyed afterwards, in which case we wouldn't need this check.

source/main/resources/CacheSystem.cpp
```diff
@@ -1138,8 +1140,7 @@ void CacheSystem::LoadResource(CacheEntry& t)
         return;
     }
 
-    static int rg_counter = 0;
-    String group = "Mod-" + std::to_string(rg_counter++);
+    String group = t.fname;
```
only-a-ptr: Well, there's non-zero chance this will cause conflicts, I can imagine thins like `woodenbox.load`

tritonas00:
seems to work fine!

PR #2434, opened 2019-09-06
Added physics pause/unpause notification

tritonas00:
Requested from discord

PR #2439, opened 2019-09-09
:bug: Fixed possible divison by zero

only-a-ptr:
performing physics update with 0 steps makes no sense and causes division by zero in `Actor::ForceFeedbackStep()` (and likely other complications)

Found here: https://github.com/RigsOfRods/rigs-of-rods/pull/2434#issuecomment-529402927


tritonas00:
cant find anything wrong :+1:

PR #2441, opened 2019-09-10
Aero dashboard: hide unused throttles & starters

only-a-ptr:
image

What do you think?


4x4convoy:
i like it, the whole plane gui needs and overhaul but that works for now lol

tritonas00:
works fine!

PR #2446, opened 2019-09-19
Reworked notifications panel

tritonas00:
Before:
bef
After:
after

Needs a bit testing


CuriousMike56:
Was just about to post that:
image
image

tritonas00:
Is it now better? If no i guess we will keep the original size

CuriousMike56:
Close: 
image

tritonas00:
will come with better solution

PR #2455, opened 2019-10-10
Improved default settings

CuriousMike56:
- Set input grabbing to 'All' (None/Dynamic causes two mouse cursors to appear, we get reports about this almost daily)
- Enabled PSSM
- Set flares mode to 'All vehicles, all lights'


only-a-ptr:
Thanks

PR #2447, opened 2019-09-20
[WIP] UI update

tritonas00:
- Moved notification panel on top center, added text shadows
- Prevents notifications overlapping with minimap and client list
- Moved best lap time beside the timer
- Updated chat, client list and notifications background to match imgui theme
(thanks to @CuriousMike56 )

after

PR #2456, opened 2019-10-11
Declutter widgets

tritonas00:
- Smaller notifications widget on Top Center
- Minimap now overlaps the client list
- Moved best lap time beside the timer

Should be good for now to declutter [this mess](https://user-images.githubusercontent.com/2660424/66669468-429b4b80-ec60-11e9-9c12-2dc53a62409b.png)

### DIFF
--- a/resources/mygui/Utils.layout
+++ b/resources/mygui/Utils.layout
@@ -46,7 +46,7 @@
               <Property key="MaxTextLength" value="99999"/>
             </Widget>
         </Widget>
-        <Widget type="ImageBox" skin="ImageBox" position="535 5 485 100" align="Right Top" name="notification">
+        <Widget type="ImageBox" skin="ImageBox" position="535 5 485 80" align="Top Center" name="notification">
             <Property key="ImageTexture" value="NotificationBg.png"/>
             <Property key="Visible" value="false"/>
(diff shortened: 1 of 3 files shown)


CuriousMike56:
What happened to the notification box from #2447? The orange will eventually be phased out as it's no longer used in our branding.
Also probably should find a way to dismiss the notification if the top menu is shown, otherwise this happens:
image

tritonas00:
> The orange will eventually be phased out 

Yeah, but then we need to squash it from all widgets not only the notifications one

> Also probably should find a way to dismiss the notification if the top menu is shown, otherwise this happens

I can't see any possible way to things not overlap other things with what we have now. This PR declutters the top right corner. 

@only-a-ptr plans to replace mygui with imgui+ogre overlays, lets see how that will turn out, then add a nice global theme

CuriousMike56:
Ok then. In that case I guess it's fine 👍

only-a-ptr:
Nice and minimal temporary improvement.

My plan is roughly:
* Get OGRE 1.12 port and ogre-imgui integrated.
* Remake all MyGui windows with dearImgui.
* Remake dashboard layout system using ogre overlays (but still support the XML format!)
* ditch MyGui for good.

PR #2459, opened 2019-10-23
Updated ImGui theme

tritonas00:
Updated ImGui theme to Blue/White/Dark colors to match the current site and RoR logo. Its a first attempt so feedback is appreciated.

kk

kkk


only-a-ptr:
It's decent, but somehow feels cold and reserved. I guess the blue works with forum's white/warm-lightgray backgrounds but doesn't with ror's dark transparent background.

I'd try:
- a bit darker but more contrasting blue
- thin light borders around widgets, like here: https://raw.githubusercontent.com/wiki/ocornut/imgui/web/v167/v167-misc.png - the bright orange widgets had enough contrast themselves, but the blue need help.
- semi-thick light border around windows, feels missing
- two-color backgrounds - lighter for menus or "tab" buttons - may require code change, but you can do it.

tritonas00:
Indeed is a bit cold, i used the same color and changed the alpha of it.
I will try with full alpha and 3 blue variants, darker too brighter, to make things more alive

tritonas00:
@only-a-ptr can you try now? I can only work colors, you do the code work :smiley:

would be nice if we had window shadows, to give some depth and feeling

tritonas00:
I will try a new one.

PR #2460, opened 2019-10-24
Finish replacing MyGUI with DearIMGUI

only-a-ptr:
This is needed for OGRE 1.12x port which has issues with MyGUI on DirectX9.

However, to make this transition worthwhile, it needs to use OGRE 1.12x and it's builtin DearIMGUI support. To break this circular dependency, I'll only create prototypes here and fine-tune them after the OGRE upgrade.

**THESE ARE PROTOTYPES - expect glitches and missing features until after OGRE 1.12x upgrade.**

Completed so far: Truck info box
screenshot_2019-10-24_11-22-28_1

Console prototype. UPDATE: now with filtering options!
screenshot_2019-10-25_12-09-54_1

Loading window - nothing fancy
image

FrictionSettings - very important for future development. TODO: add fx settings.
image

VehicleDescription - enhanced as requested: #2461
image

Netchat - unified with console (0x38x style) with custom filter settings.
screenshot_2019-11-12_14-44-14_1

FPS box + Multiplayer client list. You may notice glitch - clipped triangle count in FPS box - already fixed.
screenshot_2019-11-14_19-08-00_1

Selector UI - complete, except "tab" hotkey can only bring focus to search box, but not remove it.
image

TextureTool
image


tritonas00:
@only-a-ptr consider to unify the chatbox and console like it was before

it was more convenient and you can clean some code maybe

CuriousMike56:
Sorry about not commenting yesterday, but I believe the chat/console be unified again and leave the spawner log separate. The average RoR player does not care about the spawner log. 
Right now some server commands can only be executed from console and others chat.

tritonas00:
old chat code here https://github.com/RigsOfRods/rigs-of-rods/blob/master/resources/overlays/various.overlay#L1-L108

only-a-ptr:
@CuriousMike56 Many thanks for the screenshot, it's excellent reference.
@tritonas00 I'm thinking console window with any number of tabs, each with separately adjustable filtering like on my prototype screenshot above.

I'm all for unified console+chatbox, it makes sense both technically and humanly, I was going to suggest it myself later.

I'd go for all-inclusive on-screen log (like in Mike's screenshot, with optional window mode) with custom filtering and with spawner messages off by default (except fatal errors which will bite the player anyway).

tritonas00:
Nice and clean! 
kk

Love the filtering options! Still no input though

tritonas00:
@only-a-ptr in case you want to play more with the progress bar
https://github.com/ocornut/imgui/issues/2020
https://gist.github.com/ice1000/865c7666d13b945628254aa00bd9d62d
https://github.com/ocornut/imgui/issues/1901

CuriousMike56:
Small request: Can we get a "These settings are for advanced users only, you should only change these if you know what you're doing" warning under Settings -> Diagnostic? 
We've had quite a few people come to support about issues caused by changing these settings (e.g. cranking up the physics steps to 10k, then asking why the game runs terribly).

tritonas00:
I will post here issues i found so far, so you can review them after you are done.

- Open `About` panel and close via `X`: No main menu, escape works
- Open `About` panel and double click on the window title: An empty debug panel will appear
- Open `Friction Settings` and double click on the window title: An empty debug panel will appear
- Open `Texture Tool` and double click on the window title: An empty debug panel will appear
- `CTRL T` panel: Categories should be opened be default and not have an `X` button
- `Loader`: Categories don't work
- `Loader`: Scrolling doesn't work on Vehicles side panel
- `Loader`: Search should be focused be default (sow you can search instantly)
- `Chat`: Chat messages wont show up
- `Chat`: Please disable warnings and errors by default, to much spam
- Multiplayer user list: Admin/mod authentication level doesn't recognised, ranked user works
- Panels that appear in game should halt background focus (either input like console/chat either moving a window - should not move the actor also
https://github.com/RigsOfRods/rigs-of-rods/pull/2296
https://github.com/RigsOfRods/rigs-of-rods/pull/2299

tritonas00:
https://github.com/only-a-ptr/rigs-of-rods/blob/more-imgui/source/main/gui/GUIInputManager.cpp#L128-L131
```
    if (RoR::App::sim_state.GetActive() == RoR::SimState::RUNNING &&
            RoR::App::GetGuiManager()->IsVisible_TopMenubar() || RoR::App::GetGuiManager()->IsVisible_MainSelector() || RoR::App::GetGuiManager()->IsVisible_VehicleDescription() || RoR::App::GetGuiManager()->IsVisible_TextureToolWindow()) // dirty hack to block imgui handled input events
    {
        return true;
    }
```
https://github.com/only-a-ptr/rigs-of-rods/blob/more-imgui/source/main/gui/GUIInputManager.cpp#L263
```
    if (RoR::App::sim_state.GetActive() == RoR::SimState::RUNNING && RoR::App::GetGuiManager()->IsVisible_MainSelector())
    {
            return true;
    }

    if (RoR::App::sim_state.GetActive() == RoR::SimState::RUNNING && RoR::App::GetGuiManager()->IsVisible_Console())
    {
        if (_arg.key != OIS::KC_GRAVE)
        {
            return true;
        }
    }

    if (RoR::App::sim_state.GetActive() == RoR::SimState::RUNNING && RoR::App::GetGuiManager()->IsVisible_ChatBox())
    {
        if (_arg.key != OIS::KC_Y)
        {
            return true;
        }
    }
```

fixes the background focus thing

tritonas00:
`source/main/gameplay/RoRFrameListener.cpp:37:24: fatal error: GuiManager.h: No such file or directory`

please correct to `GUIManager.h`

only-a-ptr:
I solved the GUI/game input overlapping:

**Keyboard:** Every window specifies when it wants to halt game input: most do it when cursor is above them, the SelectorUI does it when visible, non-interactive windows like truck stats don't do it at all.

**Mouse:** Input halted automatically when mouse is above any GUI panel, except those defined with "_NoInput" flag (invisible chatbox, truck stats)

tritonas00:
Great! But please load a map and open vehicle selector via CTRL G - try to select something or even close the window. When you open it from top menu works fine. Same with survey map.

only-a-ptr:
Oh! I suspected a problem like this - when the GUI starts capturing keyboard, the gameplay never receives "key released" and the input event is triggered over and over. Anyway, I think I found an elegant solution.

tritonas00:
Ok now it works, but can't close the survey map (tab)

only-a-ptr:
That's a different problem - surveymap employs the "capture keyboard if mouse is inside" mode. I'll change it to capture only mouse when inside and pass keyboard thru.

tritonas00:
@only-a-ptr there is something going with the small survey map

Load this terrain for example https://forum.rigsofrods.org/resources/romania.405/ and open the small survey map -> fps drops to 22. Big map doesn't have the issue.

only-a-ptr:
The FPS drop on small surveymap is most definitely not new, the small map always worked by rendering the terrain in real time, if you zoom close enough you can see actual vehicles. The big map just renders the map once on startup. I left it this way because the big map shows rendering issues and the optimal case would be another remake: 
1. RoR gets 'full terrain screenshot' feature 
2. modders use it to manually author nice map backgrounds
3. terrn2 formats gets 'survey map background texture' feature

tritonas00:
please check those two 
kk

and it should be enough for now

there are other issues, but not show stoppers

CuriousMike56:
Things I've noticed so far:
- Window sizes in general are too small (1080p display)
image
image
image
- OK/Cancel buttons should be on the right side and could be a tad larger
- Warnings/errors should be disabled by default - too much spam
- Notifications become hidden with console open - can only be shown if "Logfile echo" is enabled, these should be shown all the time
- Going back to menu and updating cache after loading a terrain/vehicle appears with a scrollbar:
image

only-a-ptr:
@tritonas00 Both admin+mod show fine for me, but I hacked rorserver to test it - maybe there's an issue with the rorserver?
image

tritonas00:
@only-a-ptr can you try on official servers?

only-a-ptr:
@tritonas00 OK, but later

CuriousMike56:
Chat messages appear now, but own username is missing:
image
Names also don't appear correctly in the console:
image
The first 'test' message was from me, and the MOTD name is just garbage text
Also, the easiest way to check the official servers is to just look at the bot, it should have admin:
image

tritonas00:
@only-a-ptr https://github.com/only-a-ptr/rigs-of-rods/blob/more-imgui/source/main/gui/panels/GUI_ConsoleWindow.cpp#L33-L35

please make it
```
ImGuiWindowFlags win_flags = ImGuiWindowFlags_NoCollapse | ImGuiWindowFlags_MenuBar;
ImGui::SetNextWindowPosCenter(ImGuiSetCond_Appearing);
```
so we can resize and move the console

tritonas00:
@only-a-ptr is it possible to execute `ImGui::SetKeyboardFocusHere();` after https://github.com/only-a-ptr/rigs-of-rods/blob/more-imgui/source/main/gui/panels/GUI_MainSelector.cpp#L119 but only once? (not on every frame) So when you open the selector, search box will be activated and ready for input

only-a-ptr:
> @tritonas00 So when you open the selector, search box will be activated and ready for input

Currently, DearIMGUI can set keyboard focus programmatically, but it can't remove it. So the search box can be focused with Tab key, but there's no hotkey to remove the focus. For that reason, I'd leave it as-is for the time being.

tritonas00:
Ok, sounds good.

Is it possible to make the console auto-scroll to the last line and the selector auto-scroll when you use arrows to select items?

only-a-ptr:
`SetWindowFocus()` - Just tried it to be sure, no, it doesn't remove keyboad focus.

Mousewhell is most likely our fault, the input handling is flaky, I'll tackle it after OGRE 1.12x, ideally while migrating to SDL

tritonas00:
@only-a-ptr this does what we want

```
    if (ImGui::IsRootWindowOrAnyChildFocused() && !ImGui::IsAnyItemActive() && !ImGui::IsMouseClicked(0))
       ImGui::SetKeyboardFocusHere(0);
````

https://github.com/ocornut/imgui/issues/455#issuecomment-167440172

only issue, is when in game you can close the selector only via the cancel button

only-a-ptr:
@tritonas00 If we do that, the left/right arrow hotkeys (previous/next category)  will stop functioning (the input field steals them for the cursor movement) - there's no way to remove the keyboard focus.

I actually thought it would also steal up/down arrows (previous/next entry) but that works, probably because 'HistoryCallback' is not registered with the input box.

So the question is - do we want to focus the text field by default and lose left/right arrows for categories?

tritonas00:
No we want the arrows also. I would say leave it as is, tab key is fine.

The big problem is with `y` for the chat, that should have focus by default.

tritonas00:
`this->SetVisible(false);` right after https://github.com/only-a-ptr/rigs-of-rods/blob/more-imgui/source/main/gui/panels/GUI_GameChatBox.cpp#L97 resolves the chat thingy - once you hit enter, message is send and widget closes to continue the game.

updated https://github.com/RigsOfRods/rigs-of-rods/pull/2460#issuecomment-566040058 with two remaining issues.

only-a-ptr:
image

tritonas00:
hmm, so why this is not showing correctly for example?
https://github.com/CuriousMike56/RoRServerBot/blob/master/RoR_client.py#L1663

https://github.com/CuriousMike56/RoRServerBot/blob/master/RoR_client.py#L10

only-a-ptr:
I can't test online now, I'll check later

tritonas00:
Works now thanks!
aff465c27e808e4b01ead3891e86f0da

CuriousMike56:
Selector also needs color support:
image
Example mod: https://forum.rigsofrods.org/resources/grave-digger-35.648/

tritonas00:
Idea for bottom left messages (notifications/chat etc..) improvement. The font is white and is difficult to see with bright map backgrounds. Instead of messing with using other font/add shadows/outline i came out with this simple, three line code solution.

https://streamable.com/wz6hi

It creates an unclickable imgui button for each message. Greatly improves visibility and its fun to see, especially the chat.

only-a-ptr:
@CuriousMike56: I don't want colorized mod names or server names or any other user-controlled strings. It's disruptive and liable to conflict with built-in coloring (actor list in multiplayer for example).

I didn't even know it was possible before - I'll just strip the tags in the parser.

PR #2466, opened 2019-12-04
 Added odef option to disable shadow casting

CuriousMike56:
Currently, all objects cast shadows except for grass. When you try to enable shadows on a terrain that features a skybox, a shadow will cast across the entire terarin, thus preventing shadows from functioning.  

This fixes that by adding a new `nocast` option to the odef file format which disables casting of shadows for the object. Allows terrains to utilize a skybox or other large objects while retaining shadow support:
screenshot_2019-12-03_13-59-49_1
screenshot_2019-12-03_14-14-13_1
screenshot_2019-12-03_19-17-49_1
Example usage:
```
AHSkyBox.mesh
1.0, 1.0, 1.0

nocast

end
```
There's probably a better way, but this works. (my coding knowledge is extremely limited lol)

### DIFF
--- a/source/main/terrain/TerrainObjectManager.cpp
+++ b/source/main/terrain/TerrainObjectManager.cpp
@@ -1268,6 +1268,12 @@ void TerrainObjectManager::LoadTerrainObject(const Ogre::String& name, const Ogr
             continue;
         }
 
+        if (!strncmp("nocast", ptline, 6))
+        {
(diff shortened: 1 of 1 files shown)


4x4convoy:
This is a great addition that will fix/make alot of maps better!

tritonas00:
Nice!

PR #2467, opened 2019-12-05
Added settings option to enable classic (NiceMetal) shader

### LINKED ISSUE #2444: 0.4.8.x Managed Materials Issues
Hello, today i am bringing this problem to github so it can hopefully be resolved. The issue is that the managed materials were changed after version 0.4.0.7 to support pssm shadows on vehicles and it kinda pretty much ruined the look of the things that use managed materials (so ALOT of mods lol). I propose to revert back to the managed materials used on 0.38.x versions (aka the nicemetal shaders) but with pssm support. Here are some examples of the differences.

example #1
example #2
example #3

CuriousMike56:
Since I don't see the NiceMetal shader getting shadow support anytime soon (https://github.com/RigsOfRods/rigs-of-rods/issues/2444#issuecomment-532035335) and it really needs to be included in upstream, I figured it would be best to just add a settings option to toggle them for now.

image

Closes #2444

### DIFF
--- a/resources/managed_materials/managed_mats_vehicles_nicemetal.material
+++ b/resources/managed_materials/managed_mats_vehicles_nicemetal.material
@@ -1,0 +1,187 @@
+import * from "shadows.material"
+
+material managed/flexmesh_standard/specularonly_nicemetal
+{
+    technique BaseTechnique
+    {
+        pass BaseRender
+        {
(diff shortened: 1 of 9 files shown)


tritonas00:
Works fine, nice one!

CuriousMike56:
@only-a-ptr

PR #2477, opened 2020-01-20
Virtual keyboard

only-a-ptr:
**Status: compiles, not tested to run yet**

I suggested this on Discord chat:
> regarding the 'ever confusing controls', would it help if I added this panel (https://cryham.tuxfamily.org/portfolio/crystal-keys/) to RoR with colors and descriptions like in https://docs.rigsofrods.org/gameplay/controls-config/ and made it accessible from the ingame top menubar (menu 'Settings')?

I got positive feedback, so here I go. Changes done so far:
* Incorporated ckeys to RoR codebase, sources placed to /source/main/gui/ckeys with orig. layout
* Removed ckeys AppMain.h/cpp, setup + update code put to GUI_Virtualkeyboard.h/cpp
* Direct calls to SFML (used to paint the virtual keyboard body) replaced with DearIMGUI's `DrawList` API. Everything's drawn using DearIMGUI now.
* Removed JSON support, the format is very poorly designed because autor used SAX-style JSON library and subsequently dumbed down the format for easier handling.
* Removed XML config file support, the defaults are good enough.
* Removed 'KeyHook' input handling, using existing  DearIMGUI inputs supplied by RoR


tritonas00:
just a simple imgui panel showing keyboard layout images from https://docs.rigsofrods.org/gameplay/controls-config/ with explanation arrows isn't enough?

only-a-ptr:
@tritonas00 That would probably be the 20% work which makes for 80% of the result, I'll create another PR.

PR #2481, opened 2020-01-28
Created special 'physics paused' state box

only-a-ptr:
@tritonas00 @CuriousMike56 My take at #2434, what do you think? The 'J' is read dynamically from 'input.map'.

image


CuriousMike56:
Looks good!
However, you can freeze physics per-actor with the 'end' key, so a status message should appear then also. Something like "Vehicle physics paused. Press 'end' to resume"

only-a-ptr:
image

Thanks for the hint, I didn't even know about that feature. I changed the global pause notification to "All physics paused..."

tritonas00:
Very nice!

CuriousMike56:
@only-a-ptr Just realized the message doesn't hide with the rest of the UI, meaning it appears in screenshots - Definitely want to fix that in another PR 😊

PR #2483, opened 2020-01-30
Hide 'physics paused' box with GUI

CuriousMike56:
See https://github.com/RigsOfRods/rigs-of-rods/pull/2481#issuecomment-579423118


only-a-ptr:
Thanks!

PR #2484, opened 2020-02-06
Fixed toggle physics and replay mode keys conflicting

CuriousMike56:
This fixes the toggle physics and replay mode keys conflicting. Also added a notice about adding other device inputs to input.map.
However, with this fix the physics paused notification shows EXPL:
image
The EXPL 'key' needs to be suppressed first before this can be merged.


tritonas00:
`EXPL` shows without this also

CuriousMike56:
I'll just mark this as ready - the EXPL issue can be fixed later.

PR #2488, opened 2020-02-23
ImGui improvements

tritonas00:
- Aligned SimActorStats with the fps counter window
- Added warning under Settings/Diagnostic https://github.com/RigsOfRods/rigs-of-rods/pull/2460#issuecomment-559516520
- Auto calculate selector window size
- Auto calculate console window size
- Moved best lap time below timer again since now notifications are on bottom left
- Corrected text under GameAbout window (Build time)
- Increased chat messages timeout to 10 sec
- Moved selector ok/cancel buttons on the right side
- Hide errors from chat
- Hide main menu when connected
- Selector: focus search by default
- Αdded slight transparency on all windows
- Console: auto scroll to bottom
- Chat/notifications: always scroll to bottom to see new lines
- Hide chat/notifications also  when gui is hidden


CuriousMike56:
image
image
Much better. Could you do the same to the console window? Not as large, just a tad smaller.

tritonas00:
Done

source/main/gui/panels/GUI_MainSelector.h
```diff
@@ -38,12 +39,16 @@ namespace GUI {
 class MainSelector 
 {
 public:
-    const float LEFT_PANE_WIDTH = 200.f;
-    const float WINDOW_WIDTH = 600.f;
+    Ogre::Viewport* viewport = RoR::App::GetOgreSubsystem()->GetRenderWindow()->getViewport(0);
+
+    const float LEFT_PANE_WIDTH = 250.f;
+    const float WINDOW_WIDTH = viewport->getActualWidth() / 1.4;
+    const float WINDOW_HEIGHT = viewport->getActualHeight() / 1.2;
```
only-a-ptr: TIP: read screen size from `ImGui::GetIO().DisplaySize` (example: https://github.com/RigsOfRods/rigs-of-rods/blob/4f8658fcd653bda8f0ebea05760fc9adf05a13c7/source/main/gui/panels/GUI_MultiplayerClientList.cpp#L53) - it's the same numbers, updated every frame.

source/main/gui/panels/GUI_SimActorStats.cpp
```diff
@@ -36,7 +36,7 @@ void RoR::GUI::SimActorStats::Draw(RoR::GfxActor* actorx)
     // TODO: Localize all! See https://github.com/RigsOfRods/rigs-of-rods/issues/1269
     ImGuiWindowFlags flags = ImGuiWindowFlags_NoInputs | ImGuiWindowFlags_NoCollapse |
         ImGuiWindowFlags_NoMove | ImGuiWindowFlags_NoResize | ImGuiWindowFlags_NoTitleBar;
-    ImGui::SetNextWindowPos(ImVec2(25.f, 150.f));
+    ImGui::SetNextWindowPos(ImVec2(10.f, 160.f));
```
only-a-ptr: TIP: use common 'theme' setting 'screen edge padding' (example: https://github.com/RigsOfRods/rigs-of-rods/blob/4f8658fcd653bda8f0ebea05760fc9adf05a13c7/source/main/gui/panels/GUI_MultiplayerClientList.cpp#L53)

tritonas00:
@only-a-ptr is it possible to move a scrollbar on the left side of a window?

only-a-ptr:
@tritonas00 No it isn't, this code would have to change: https://github.com/ocornut/imgui/blob/24bd33ace808ed77fd1840b19f2b8b4789a3ae62/imgui_widgets.cpp#L934-L935. We currently bundle older imgui with different code, you can try hacking it: https://github.com/RigsOfRods/rigs-of-rods/blob/4f8658fcd653bda8f0ebea05760fc9adf05a13c7/source/main/gui/imgui/imgui.cpp#L4472-L4474

tritonas00:
Something overwrites the console instead of appending, want to fix it then its ready to merge.

tritonas00:
@only-a-ptr any idea why this is happening? Spawning the tatra rewrites the log instead of appending on it. Demo videos:

https://streamable.com/edt3d
same without Actors filter
https://streamable.com/y8ky3

only-a-ptr:
@tritonas00 It's most definitely this: https://github.com/RigsOfRods/rigs-of-rods/blob/4f8658fcd653bda8f0ebea05760fc9adf05a13c7/source/main/system/Console.cpp#L108-L114 - `MESSAGES_CAP` is 1000 and increasing it will most definitely eat extra FPS when console is displayed - it's already not so good. The whole system is a straightforward prototype: the console filters the entire message history every frame. It should probably do this instead: keep a persistent 'display list'. When filter changes, re-scan entire history to get new display-list. When new message arrives, filter it individually and eventually add to display list.

tritonas00:
Yep, it was that, increased to 2000 "fixed" it. So i'm done here, maybe you want to do this console change here and merge, or merge now and do it later.

only-a-ptr:
Thanks for doing this - it's all so much nicer now 👍

PR #2495, opened 2020-03-10
Αdded 'commands' filter

tritonas00:
Added `cvw_filter_type_cmd`, enabled in console and disabled in chat/notifications so when you enter a command in console it will not printed as notification.


only-a-ptr:
👍

PR #2496, opened 2020-03-10
Re added 'Default' skin in selector

tritonas00:
This will select the default skin if no one is selected, but i don't know how to show it as an entry also.
It should always be available on top of the skinzips list.

(Example: 767 airplane)

### DIFF
--- a/source/main/gui/panels/GUI_MainSelector.cpp
+++ b/source/main/gui/panels/GUI_MainSelector.cpp
@@ -67,11 +67,6 @@ void MainSelector::Show(LoaderType type, std::string const& filter_guid)
     {
         m_selected_entry = m_last_selected_entry[m_loader_type];
     }
-
-    if (type == LT_Skin)
-    {
-        m_selected_entry = m_selected_entry["Default"];
-    }
 }
 
 void MainSelector::Draw()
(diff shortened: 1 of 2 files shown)


source/main/gui/panels/GUI_MainSelector.cpp
```diff
@@ -67,6 +67,11 @@ void MainSelector::Show(LoaderType type, std::string const& filter_guid)
     {
         m_selected_entry = m_last_selected_entry[m_loader_type];
     }
+
+    if (type == LT_Skin)
+    {
+         m_selected_entry = m_selected_entry["Default"];
```
only-a-ptr: 😱 I had**NO** idea this could even compile, but apparently it does: http://madebyevan.com/obscure-cpp-features/

tritonas00:
Its just to select no skinzip hehe. The Default option isn't there in the new selector, and the first custom skinzip is automatically selected. Feel free to close this and do it properly. All we need is a 'Default' top entry which means no skinzip is selected.

CuriousMike56:
When you select a vehicle with extra skins available, the old selector would give the option to select the default texture (no custom skin): 
image
This isn't shown on the new selector, you have to select a custom one instead.

tritonas00:
with this PR it will load no custom skin if none is selected but yes a "Default Skin" entry would be the best way.

tritonas00:
Done, thanks to only-a-ptr :star:

kk

PR #2498, opened 2020-03-14
[WIP] Added notification/chat icons and other improvements

tritonas00:
- Added notices/chat/warning icons
Logic already used in `GUI_MultiplayerClientList`, but here uses filters to fetch and draw icons.
Icons draw only for default enabled chat filters (notices/chat/warnings) and disabled in console, so we don't hurt performance.
kkkk

- Added a nice initial message when you load a terrain for first time
kk

- Removed chat filtering options: It was available only in multiplayer and enabling extra filters causes too much spam unlike in console where they make total sense. Currently default enabled filters are all we need.

kk

- Reduced max chat/notification lines to 15

- Added an extra space between chat message input box and messages draw area to prevent overlapping when messages go to full bottom


only-a-ptr:
👍 Awesome

PR #2504, opened 2020-03-17
Survey map fixes

tritonas00:
- Fixed teleport in big map when cropped

- Fixed `mMapCenterOffset` was not reset when a new terrain is loaded

- Fixed `Overview map icons` and `Declutter overview map` settings options
First hides all icons, second hides only terrain icons

- Fixed big map when terrain has no heightmap


CuriousMike56:
Would like to keep the cropping. It's very helpful for meshed terrains.

tritonas00:
fixed teleport without disabling cropping

PR #2505, opened 2020-03-19
Hotfix: Truck stats not being updated

only-a-ptr:
Introduced in e23ea80ad43e7650c0aae5de8a5ed25be18be043

Reported by @tritonas00 👍


CuriousMike56:
Tritonas requested me to add these here:
image
image
- G-forces are missing
- Extra decimal places aren't necessary
- Kw/HP should be swapped

PR #2522, opened 2020-03-30
[WIP] Implemented chat history

tritonas00:
By pressing `Y` now you can see and scroll through chat messages. It shows only chat messages for a cleaner output, different icon also.

Normal chat/notifications
kk
Pressing `Y` 
kkk

For this to work as it should [this](https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/system/Console.cpp#L112) must not happen. But if is not, the console window will cause fps drop when visible with many lines (mostly spawner errors/warnings)

@only-a-ptr must do this https://github.com/RigsOfRods/rigs-of-rods/pull/2488#issuecomment-594464050 before this can be merged


tritonas00:
will make a new PR

Issue #2527, opened 2020-04-09
Option to disable the self-cancelling feature of turn signals


MarkROR:
When driving any vehicle that requires constant steering inputs just to go straight, these minor inputs are enough to trigger the self-cancelling of turn signals/indicators, before taking the actual turn.

It would be nice to have a setting to disable this feature, or to improve it so it doesn't trigger unless the steering wheel is turned enough and then returned to center (I assume this is already in place, so maybe just increase the degrees of turn required, while keeping in mind wide turns are a thing)

**To see this in action**: any map, any car (St Helens, Honda Accord) - drive in a straight line without any steering input, turn either of the indicators on, perform the slightest steering input in that direction. The indicator will be cancelled when you return the wheel to center.

**What should happen**: the indicator should remain on until the steering angle reaches a certain (TBD) treshold, and when the steering is straightened again, only then it should cancel the indicator.

tritonas00:
probably https://github.com/RigsOfRods/rigs-of-rods/blob/4f8658fcd653bda8f0ebea05760fc9adf05a13c7/source/main/physics/Beam.cpp#L3168

CuriousMike56:
https://github.com/RigsOfRods/rigs-of-rods/blob/4f8658fcd653bda8f0ebea05760fc9adf05a13c7/source/main/physics/Beam.cpp#L3165-L3194

tritonas00:
You can now adjust blinker lock range in Settings->Controls tab

PR #2528, opened 2020-04-10
Made blinker lock range configurable

tritonas00:
fixes https://github.com/RigsOfRods/rigs-of-rods/issues/2527

kk

`RoR.cfg` option: `Blinker Lock Range`


tritonas00:
we can add a slider in settings, but `0.1` was to low anyway

tritonas00:
@CuriousMike56 @AnotherFoxGuy in which tab this should be? Is `Controls` right?

CuriousMike56:
@tritonas00 Controls seems fine.

Issue #2535, opened 2020-04-13
Corrupted mini image crashes


CuriousMike56:
If a mod has a corrupted -mini image, RoR will crash instead of failing silently. 
![image](https://user-images.githubusercontent.com/46073351/79170163-8fe9a880-7dbc-11ea-9c7e-9e73203de7c7.png)

How to reproduce:
- Download and install  [CorruptedMiniTest.zip](https://github.com/RigsOfRods/rigs-of-rods/files/4472345/CorruptedMiniTest.zip). This zip contains a DDS image renamed as a JPG.
- When you attempt to load the file in-game, RoR will crash with the above error.

There are different versions of this error, but they all have the same or similar cause:
![image](https://user-images.githubusercontent.com/46073351/79170361-26b66500-7dbd-11ea-8d1b-2e650198e179.png)
![image](https://user-images.githubusercontent.com/46073351/79170135-78aabb00-7dbc-11ea-8f69-5b145a36ac2d.png)

In 2020.01, loading the above test file works, just no image will be displayed as FreeImage failed sliently:
![image](https://user-images.githubusercontent.com/46073351/79170404-46e62400-7dbd-11ea-9a4b-06fbcbbb6260.png)

PR #2536, opened 2020-04-14
Selector: ignore invalid mini images

tritonas00:
fixes https://github.com/RigsOfRods/rigs-of-rods/issues/2535

### DIFF
--- a/source/main/gui/panels/GUI_MainSelector.cpp
+++ b/source/main/gui/panels/GUI_MainSelector.cpp
@@ -201,23 +201,30 @@ void MainSelector::Draw()
         if (!sd_entry.sde_entry->filecachename.empty())
         {
             ImVec2 cursor_pos = ImGui::GetCursorPos();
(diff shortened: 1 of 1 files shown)


source/main/gui/panels/GUI_MainSelector.cpp
```diff
@@ -201,23 +201,30 @@ void MainSelector::Draw()
...
+                    ImGui::SetCursorPos(cursor_pos);
                 }
-                // Draw the image
-                ImGui::SetCursorPos((cursor_pos + ImGui::GetWindowSize()) - size);
-                ImGui::Image(reinterpret_cast<ImTextureID>(preview_tex->getHandle()), size);
-                ImGui::SetCursorPos(cursor_pos);
+            }
+            catch(...)
+            {
+                Ogre::TexturePtr preview_tex; // Invalid image
```
only-a-ptr, line 227: This line doesn't make sense - replace it with `// OGRE exception already logged`

Issue #529, opened 2016-01-06
disable stopping blinker when turning 


DirtGamer301:
When you blink to the left and slightly steer to the left for a short time, when finished steering the blinker will go out. This might be a realistic feature for modern cars, but I'm currently working on a 70s small car which I don't think had the ability to do this. 

Forum link (possible discussions fullowing): http://www.rigsofrods.com/threads/120838-disable-stopping-blinker-when-turning

PR #2538, opened 2020-04-15
Added option to disable blinker auto reset

tritonas00:
fixes https://github.com/RigsOfRods/rigs-of-rods/issues/529

kk
kkk


tritonas00:
Reading this https://archives.rigsofrods.net/old-forum/forums/106-General-Chat/120838/page-0001.html, it seems the option should be in the truck file

>Is there a way to turn this off in the truck file

So this PR is no go

PR #2544, opened 2020-04-21
Increase max gear limit in RoR

disloyalpick:
just a small change to be tested . line 40 and line 155 have been edited to attempt to increase the maximum gear limit ror supports


CuriousMike56:
Those are comment lines...

PR #2547, opened 2020-04-22
Increase maximum gear limit(unlimited in future)

disloyalpick:
increase maximum gear limit in RoR


AnimosusTheWolf:
For what purpose?

only-a-ptr:
Actually, since the code already handles unlimited number of gears, it doesn't make sense to have any magic-number limitation in the parser.

PR #2552, opened 2020-05-01
Added Game Controls panel

tritonas00:
A simple game controls panel
kk


CuriousMike56:
I feel like adding a "Help" section to the top menu would be better than under Tools.

tritonas00:
I think its fine there, Petr will touch at some point and then it will definitely become a tool lol

PR #2555, opened 2020-05-05
⬆️ Upgrade OGRE to 1.12

AnotherFoxGuy:
Update on this:

I've updated OGRE to 1.12.6 
Everything seems to be working
Crashes on DX9 because the DX11 imgui shaders fail to compile

@only-a-ptr Could you try and update imgui to the one included with OGRE on this branch?


AnotherFoxGuy:
Will be upgrading to OGRE 13 instead

PR #2557, opened 2020-05-05
Fixed EV_COMMON_FOV_LESS not working

tritonas00:
from discord:

>hm. it seems the lower FOV key (numpad 7) doesn't work now
ctrl+numpad7 to increase FOV works

Seems it was a typo


only-a-ptr:
Doh! That wasn't actually a typo, that was a derp :blush:
Thanks for the fix!

PR #2570, opened 2020-06-20
disable cotire to uncover include errors

paroj:


paroj:
here is a list of missing includes: https://github.com/paroj/rigs-of-rods/blob/6ac0d06b3e29051e33256425ddea47c6fa3383c5/snapcraft.yaml#L133

tritonas00:
@paroj i think we can close this now https://github.com/RigsOfRods/rigs-of-rods/pull/2572

PR #2575, opened 2020-07-07
Fixed slow cache regen

tritonas00:
Brings back ulteq's https://github.com/RigsOfRods/rigs-of-rods/pull/2127

### DIFF
--- a/source/main/gui/panels/GUI_LoadingWindow.cpp
+++ b/source/main/gui/panels/GUI_LoadingWindow.cpp
@@ -39,9 +39,10 @@ void RoR::GUI::LoadingWindow::setProgress(int percent, std::string const& text, 
         ++m_text_num_lines; // New line
     }
 
-    if (render_frame)
+    if (render_frame && m_timer.getMilliseconds() > 10)
     {
         Ogre::Root::getSingleton().renderOneFrame();
+        m_timer.reset();
     }
 }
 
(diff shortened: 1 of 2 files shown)


tritonas00:
i think its the `OGRE_THREAD_SUPPORT` option

- OGRE_THREAD_SUPPORT=1 is full thread safety for resource creation and loading. Allows GPU resources to be created in a background thread but requires quite a lot of locking, and the rendersystem is a bit slower because it has to be put in threadsafe mode.
- OGRE_THREAD_SUPPORT=2 is thread safety for loading data from disk only. Resources are 'prepared' in the background, which means that disk loading is done, and potentially some other tasks which do not require access to the GPU. For example in trunk, the terrain calculates lightmaps, LOD deltas and other data but does it on CPU data only. All GPU resources are synchronised in the main render thread only. This requires considerably less locking and the rendersystem can operate in non-threadsafe mode which is faster. We generally recommend OGRE_THREAD_SUPPORT=2 now, and it's the default in trunk if Boost, POCO or TBB are detected in your environment.

https://forums.ogre3d.org/viewtopic.php?p=353461#p353461

only-a-ptr:
@tritonas00 My thoughts exactly: https://github.com/RigsOfRods/rigs-of-rods/issues/2576 😄

Issue #1359, opened 2017-07-31
Empty submesh section crashes using DirectX


Michael10055:
If you have a submesh section that is empty and you're running DirectX,  RoR will crash. The error does not happen on OpenGL.
![image](https://user-images.githubusercontent.com/11002490/28795317-74881df2-7607-11e7-8807-a928f6f23152.png)

Steps to reproduce:

- Set RoR to use DirectX if you don't use it already
- Open any truck file and add these lines to the end:
```
submesh
texcoords
cab
```
- Save the file and spawn it ingame. The error shown above should happen.

only-a-ptr:
What the... 😆 

OK, this is quite interesting, I didn't notice any potential for a crash like this. Nonetheless, it should be an easy fix.

Michael10055:
AFAIK this crash has been around for many years too..not sure why it wasn't fixed.

tritonas00:
@paroj it seems we can't reproduce this anymore. Is there any ogre 1.11 change that may fixed it?

Or its gpu related?

paroj:
> @paroj it seems we can't reproduce this anymore. Is there any ogre 1.11 change that may fixed it?

no

> Or its gpu related?

it might be driver dependent

PR #2578, opened 2020-07-23
Check if submesh section is empty

tritonas00:
Hopefully fixes https://github.com/RigsOfRods/rigs-of-rods/issues/1359

### DIFF
--- a/source/main/physics/RigSpawner.cpp
+++ b/source/main/physics/RigSpawner.cpp
@@ -1309,6 +1309,16 @@ std::string ActorSpawner::GetSubmeshGroundmodelName()
 
 void ActorSpawner::ProcessSubmesh(RigDef::Submesh & def)
 {
+    std::list<std::shared_ptr<RigDef::File::Module>>::iterator module_itor = m_selected_modules.begin();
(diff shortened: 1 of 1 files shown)


only-a-ptr:
Completely off 😆  - I suggest you add some `RoR::Log()` calls there and check the actual control flow.
* The macro `PROCESS_SECTION_IN_ALL_MODULES` already does this same check implicitly by calling `ProcessSubmesh()` for each entry in array `RigDef::File::Module::submeshes` - if the array is empty, the function is never invoked.
* An empty `submesh` in truckfile will insert an entry to `RigDef::File::Module::submeshes` anyway, just with `texcoords` and `cab_triangles` being empty vectors - see RigDef_File.h

The problem will be somewhere in `RigSpawner::FinalizeGfxSetup()` - there are checks for submeshes and cabs arrays not being empty, but something doesn't work as intended. The actual OGRE manipulation which causes the crash is positively within `new FlexObj(...snip...)`

Anyway thanks for looking into it.

tritonas00:
If i could reproduce the issue, i could find exactly where the exception occurs and try/catch that code

PR #2580, opened 2020-07-27
⏪ Revert "migrate Conan packages to Cloudsmith" 

AnotherFoxGuy:
Cloudsmith doesn't download the correct packages for the Conan OS/compiler settings


apoclyps:
Hi @AnotherFoxGuy, 

I'm sorry to hear about the issues you encountered whilst using Conan on Cloudsmith. Conan is currently in beta at Cloudsmith and we are always trying to make improvements. If you have any feedback, I'd be more than happy to receive it either via Github, Intercom, or support@cloudsmith.com. In the meantime, I'll open an investigation ticket to look into why this was an issue and what we can do to resolve it.

PR #2585, opened 2020-08-25
Updated MyGUI to 3.4.0

AnotherFoxGuy:


tritonas00:
Does it fix the black screen issues with ogre 1.12 and Dx9 also?

AnotherFoxGuy:
Since we use imgui for the main menu, that is no longer issue

tritonas00:
yes but what about the HUD?

AnotherFoxGuy:
MyGUI still has the issue that it doesn't render on 1.12.8

@tritonas00 Could you test this PR on Linux and see if everything still works?

tritonas00:
Sure

tritonas00:
I get this https://pastebin.com/30Ta1m4G (i dont think its related with this PR since its an openal issue it seems)

PR #2603, opened 2020-09-29
Added turbojet reverse

tritonas00:
Demo video: https://streamable.com/pgiwaz


only-a-ptr:
I didn't even know this was a thing: https://en.wikipedia.org/wiki/Thrust_reversal

Also, it seems this is more of a bugfix, we already had a key binding and state variable for this.

PR #2607, opened 2020-10-17
Fixes

### LINKED ISSUE #2430: Default aircraft lights don't work anymore on 0.4.8
On the 0.4.7 version, every aircraft had default red and green navigation lights as well as flashing white strobe lights, but since I updated to 0.4.8, I figured out that these don't work anymore on any aircraft. You can try it on the default an12 : only the red flashing lights (these were added by the author I think) will turn on. Tell me how I can help more if needed

only-a-ptr:
Fixes #2430

Problem: Beacon props generated by `ActorSpawner::ProcessWings()` weren't passed on to GfxActor.
Reason: Props were passed on GfxActor creation which  happened  before   `ActorSpawner::ProcessWings()`
Solution: Added `GfxActor::RegisterProps()` which is invoked after `ActorSpawner::ProcessWings()`

### DIFF
--- a/source/main/gameplay/RecoveryMode.cpp
+++ b/source/main/gameplay/RecoveryMode.cpp
@@ -36,6 +36,13 @@ void RecoveryMode::UpdateInputEvents(float dt)
         return;
     }
 
+    if (!App::GetGameContext()->GetPlayerActor())
+    {
+        m_advanced_vehicle_repair = false;
+        m_advanced_vehicle_repair_timer = 0.0f;
+        return;
+    }
+
     if (!App::GetInputEngine()->getEventBoolValue(EV_COMMON_REPAIR_TRUCK))
     {
         m_advanced_vehicle_repair_timer = 0.0f;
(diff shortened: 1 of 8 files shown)


tritonas00:
Seems to works fine!

Although they stop when you get out of the actor, don't know if that's normal.

tritonas00 (approved):

po185021:
> @tritonas00 Although they stop when you get out of the actor, don't know if that's normal.

Uhh... I bet all other prop animations stop as well when you get out of the vehicle. That's most likely bug with my 'async scene' code, I'm pretty sure I don't update props for non-occupied actors.

only-a-ptr:
> @tritonas00 Although they stop when you get out of the actor, don't know if that's normal.

That's actually a fantastic observation which provoked 2 additional fixes 👍

tritonas00:
Works great except one issue, the blinker lights in the HUD (mygui) are not working now.

only-a-ptr:
Fixed both the blinker HUD lamps and the logged `mp_*` cvars.

tritonas00 (approved):

PR #2610, opened 2020-10-19
RFC: adapt "Xbox 360 (Wireless) Controller Input Map" for Linux

paroj:
on Linux the controller seems to have a different axes mapping than on the system where the existing map was created on.

- Is there a system in RoR for handling such OS dependent differences?
- Are there plans to move to SDL2?
  - [it provides a canonical layout for all gamepads](https://wiki.libsdl.org/CategoryGameController)

### DIFF
--- a/resources/skeleton/config/Xbox_360_Wireless_Receiver.map
+++ b/resources/skeleton/config/Xbox_360_Wireless_Receiver.map
@@ -16,25 +16,26 @@
 ;Right stick button: Button 9
 
 ;Axes:
-;Left Stick (Left/Right): Axis 1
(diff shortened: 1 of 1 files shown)


only-a-ptr:
Thanks for the patch!

> Is there a system in RoR for handling such OS dependent differences?
Not yet but I just drafted one: #2611

> Are there plans to move to SDL2? 
Yes but not in a close future. First I'd like to take a shot at migrating to OGRE 2.x which doesn't have a developed SDL2 subsystem, so I'll decide based on the resulting needs and possibilities.

only-a-ptr:
Superseded by #2611

PR #2612, opened 2020-10-21
Fixed updating survey map texture when zoomed

tritonas00:


only-a-ptr:
What's the effect on FPS?

tritonas00:
none when map is normal, not zoomed

when zoomed, there is some impact, but its necessary because the texture should move along with the actor

PR #2614, opened 2020-10-25
Restored text coloring in SelectorUI

only-a-ptr:
color wrap

### DIFF
--- a/source/main/gui/GUIManager.h
+++ b/source/main/gui/GUIManager.h
@@ -52,6 +52,7 @@ public:
 
         ImVec4 semitransparent_window_bg = ImVec4(0.1f, 0.1f, 0.1f, 0.8f);
         ImVec4 semitrans_text_bg_color   = ImVec4(0.1f, 0.1f, 0.1f, 0.6f);
+        ImVec4 color_mark_max_darkness   = ImVec4(0.2, 0.2, 0.2, 0.0); //!< If all RGB components are darker than this, text is auto-lightened.
 
         ImVec2 screen_edge_padding       = ImVec2(10.f, 10.f);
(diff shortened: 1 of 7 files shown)


tritonas00:
Nice except one issue, if you scroll, you can see the title moving up
kk

tritonas00:
Here is a vehicle with dark color name https://forum.rigsofrods.org/resources/hoodrat-2020.671/

only-a-ptr:
> @tritonas00  Nice except one issue, if you scroll, you can see the title moving up

Nice spot. It turns out to be an advanced exercise with imgui internals. So far I only managed to get it clipped when outside of the sroll area, not when partially visible.

tritonas00:
Just skip right side coloring if it is too much hassle.

only-a-ptr:
Text clipping fixed.
I solved the black text by adding 'darkness limit' - if below, the color gets inverted.

image

tritonas00:
Nice! Two things noticed:

- Please increase `size.x` in https://github.com/only-a-ptr/rigs-of-rods/blob/wee-colors/source/main/gui/panels/GUI_MainSelector.cpp#L180 otherwise mouse pointer needs to hover text only, to make a selection
`if (ImGui::Selectable("##dummy", &is_selected, 0, ImVec2(size.x + 100.f, size.y))) // Use invisible label + size parameter.` did the trick

- Spawn a vehicle (like desperado) and press `w` a couple of times:
kk

only-a-ptr:
@tritonas00 both corrected

tritonas00:
Looks good!

tritonas00 (approved):

PR #2616, opened 2020-10-26
Unicode support in UI

tritonas00:
Load every glyph in the font.
kk


only-a-ptr:
image
Works great! I wasn't able to enter all czech diacritics, but I'm sure that's an input issue rather than display issue.

PR #2629, opened 2020-12-04
Restore net colors

### LINKED ISSUE #2619: MP's player list does not match the local player's state (in vehicle or not)
### **Follow-up of November 2nd's discussion in #developpment channel**
_Hope I'm going to be clearer._

The color-shift is a new feature as It is told. The UI title bar and player name change color depending on the player's state.

> Notice the title bar and player name color

Text color changes when in vehicle

Basically, the color shift feature when I get in or out of vehicles doesn't update the player's net label's color.

> Look at UnnamedPlayer's UI name vs netlabel name

Head MP name dont change

Also, it doesn't affect every color. As of my own testing, I determined that these colors are affected :
- Dark Yellow
- Bright Yellow
- Orange
- Bright Red
- Yellow
- Light Blue
- Light Orange

image

The fact that only some colors are affected is a bit annoying to keep track of players.

The color doesn't update the netlabel player name depending on the local player's state. (keeps the player not in a vehicle state)
image

Zentro:
Fixes #2619


only-a-ptr:
I do trust you it works, but technically it doesn't make sense, see https://github.com/RigsOfRods/rigs-of-rods/issues/2619#issuecomment-739316646

PR #2630, opened 2020-12-10
Final touches before release (well, almost)

only-a-ptr:
TODO:
* ...

DONE:
*  keep nicknames in chat visible after player disconnects - now shown as "Myname [offline]: Hello!"
*  bugfix - values from commandline (MP server, terrain, vehicle) propagate to RoR.cfg
* bugfix - preselected vehicle spawns with cab not visible
*  ui - remove the PhysicsTimeStep slider from TopMenubar
* bugfix - some vehicles (i.e. Desperado) lose prop visuals on respawn.
*  ui - remove the PhysicsTimeStep slider from GameSettings, too

GEEKY:
image

### DIFF
--- a/source/main/Application.cpp
+++ b/source/main/Application.cpp
@@ -225,7 +225,6 @@ CVar* gfx_fixed_cam_tracking;
 CVar* gfx_fps_limit;
 CVar* gfx_speedo_digital;
 CVar* gfx_speedo_imperial;
-CVar* gfx_flexbody_lods;
 CVar* gfx_flexbody_cache;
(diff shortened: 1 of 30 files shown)


tritonas00:
one issue found, race arrow now doesn't work it always shows up:

kk

also if you start a race and go back to menu, the overlay isn't hidden:

kk

only-a-ptr:
> one issue found, race arrow now doesn't work it always shows up:

Unbelieveable. Makes absolutely no sense. 😫  I've backed out those commits and created #2637

PR #2636, opened 2020-12-19
Fixed airplane artifacts with static cam

tritonas00:
Fixes:

kk

which started since https://github.com/RigsOfRods/rigs-of-rods/pull/2047


only-a-ptr:
The artifacts actually happen because of this phenomenon: https://en.wikipedia.org/wiki/Z-fighting

The static camera uses extremely narrow FOVs, so it's actually very far away from the airplane. The rest is explained on Wikipedia.

From what I understood, this patch changes camera position calculation so that it's always a bit closer to the airplane... fair enough.

PR #2637, opened 2021-01-04
Nonsense issues with Renderdash

only-a-ptr:
Separated out from #2630

Original problem: Windows/Direct3D9 renderer: when player enters any vehicle, DearIMGUI colors temporarily change (looks like a hue shift). Leaving the vehicle restores them. Memory stays intact.

Orig. solution: make renderdash RTT manually updated, see commits below.

Consequence: some GUI indicators break for no apparent reason (both Windows/Linux). Code flow is unchanged. See https://github.com/RigsOfRods/rigs-of-rods/pull/2630#issuecomment-753672831

### DIFF
--- a/source/main/gfx/GfxActor.cpp
+++ b/source/main/gfx/GfxActor.cpp
@@ -2384,9 +2384,7 @@ void RoR::GfxActor::UpdateRenderdashRTT()
 {
     if (m_renderdash != nullptr)
     {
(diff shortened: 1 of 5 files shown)


only-a-ptr:
Conclusions:
* under D3D9 renderer (tested with itch build), the renderdash RTT must be manually updated, otherwise the color corruption happens.
* The race direction arrow must be disabled while rendering the renderdash, otherwise node transforms get broken and the arrow ends up pointing somewhere upwards.
* In fact, all overlays except the renderdash-specific must be disabled while rendering renderdash, or they show up in the rendered texture.

Issue #2480, opened 2020-01-23
Change the speedometer from the vehicles speed, to the wheel speed


4x4convoy:
Whenever the new GUI update happens I am suggesting to edit the speedometer on screen to display the vehicles actual wheel speed instead of the the vehicles speed. As of right now it acts like a gps would display speed irl, which is how fast the entire vehicle is moving along. Say while in game you put your vehicle up against the wall and start pressing the gas, the wheels will turn but the speedometer will still read 0 since the vehicle isn't moving. If you want to see how fast the wheels are turning you have to go into the "T" menu and look at the "Wheel Speed" number. This is kind of annoying when you're stuck in mud for example with the wheels spinning but the GUI is just reading 0 mph.

only-a-ptr:
Agreed, the speedometer should be based on wheel speed rather than vehicle (actually cinecam#0) motion vector.

PR #2640, opened 2021-01-07
Changed speedometer to wheel speed

tritonas00:
Closes https://github.com/RigsOfRods/rigs-of-rods/issues/2480

### DIFF
--- a/source/main/physics/Actor.cpp
+++ b/source/main/physics/Actor.cpp
@@ -3936,9 +3936,9 @@ void Actor::updateDashBoards(float dt)
     {
         velocity = cam_dir.dotProduct(ar_nodes[0].Velocity);
     }
-    float speed_kph = velocity * 3.6f;
+    float speed_kph = ar_wheel_speed * 3.6f;
     ar_dashboard->setFloat(DD_ENGINE_SPEEDO_KPH, speed_kph);
-    float speed_mph = velocity * 2.23693629f;
+    float speed_mph = ar_wheel_speed * 2.23693629f;
     ar_dashboard->setFloat(DD_ENGINE_SPEEDO_MPH, speed_mph);
 
     // roll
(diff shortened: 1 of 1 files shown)


only-a-ptr:
The original calculation is suspicious:
* we start with node0's velocity. Since our physics run at 2000 cycles per sec, the value is in "meters per 2000th of a second".
* to get vehicle's velocity (algined with cinecam's  forward axis), we should do a scalar projection: https://en.wikipedia.org/wiki/Dot_product#Scalar_projection_and_first_properties - this looks OK, `GetCameraDir()` gives us an unit vector and we do a dot product.
* at this point, we should multiply by 2000 (to get velocity in meters/sec) and then divide by 1000 (to get kilometers/sec). In other words, we should multiply by 2. Instead we multiply by 3.6. What the heck?

I think the reason is historical: until around 2014 the physics speed wasn't constant but subtly fluctuated via some perverted scheme. Neither me nor Ulteq could understand it and in general physics should _always_ run at constant rate, so we fixated it at 2000hz which was the closest. I think this magic simply comes from the previous dark era when you couldn't make a proper calculation.

only-a-ptr:
Right now our physics rate is governed by cvar `diag_physics_dt (float)` which contains the fraction of time one physics step takes (1/2000 = 0.005). This is impractical for calculating the speed.

I suggest:
* killing that cvar and introducing `sim_physics_freq (int)` containing directly the value 2000.
* redefining `PHYSICS_DT` as `1.f / App::sim_physics_freq->GetFloat()`

only-a-ptr:
If the value is jumpy, apply some smoothing to it:
```
speedo = cur_wheelspeed * 0.7 + prev_speedo * 0.3;
```
To diagnose, you can use DearIMGUI's awesome auto-Debug-window feat:
```
// outside ImGui::Begin() / ImGui::End()
ImGui::Text("num: %f", mynumber); // Opens a default "Debug" window
```

only-a-ptr (approved):
Works great!

PR #2645, opened 2021-01-11
🌐 Added _LC macro to all strings that can be localized

AnotherFoxGuy:


source/main/gui/panels/GUI_MultiplayerSelector.cpp
```diff
@@ -378,7 +379,7 @@ void MultiplayerSelector::UpdateServerlist(MpServerInfoVec* data)
     m_draw_table = true;
     if (m_serverlist_data.empty())
     {
-        m_serverlist_msg = "There are no available servers :/";
+        m_serverlist_msg = _L("There are no available servers :/");
```
only-a-ptr, line 382: If the string wasn't localized before, add a context to it right away

source/main/gui/panels/GUI_NodeBeamUtils.cpp
```diff
@@ -151,20 +153,20 @@ void NodeBeamUtils::Draw()
...
+        ImGui::Text(_L("Movement: %f (%f)"),   actor->ar_nb_reference[5] / actor->ar_num_nodes, actor->ar_nb_reference[4]);
+        ImGui::Text(_L("Stress: %.2f (%.2f)"), actor->ar_nb_reference[1] / actor->ar_num_beams, actor->ar_nb_reference[0]);
+        ImGui::Text(_L("Yitter:   %f (%f)"),   actor->ar_nb_reference[3] / actor->ar_num_beams, actor->ar_nb_reference[2]);
         ImGui::NextColumn();
-        ImGui::Text("Movement: %f (%f)",   actor->ar_nb_optimum[5] / actor->ar_num_nodes, actor->ar_nb_optimum[4]);
-        ImGui::Text("Stress: %.2f (%.2f)", actor->ar_nb_optimum[1] / actor->ar_num_beams, actor->ar_nb_optimum[0]);
-        ImGui::Text("Yitter:   %f (%f)",   actor->ar_nb_optimum[3] / actor->ar_num_beams, actor->ar_nb_optimum[2]);
+        ImGui::Text(_L("Movement: %f (%f)"),   actor->ar_nb_optimum[5] / actor->ar_num_nodes, actor->ar_nb_optimum[4]);
+        ImGui::Text(_L("Stress: %.2f (%.2f)"), actor->ar_nb_optimum[1] / actor->ar_num_beams, actor->ar_nb_optimum[0]);
+        ImGui::Text(_L("Yitter:   %f (%f)"),   actor->ar_nb_optimum[3] / actor->ar_num_beams, actor->ar_nb_optimum[2]);
```
only-a-ptr, line 169: again... the language string shouldn't contain `printf()` format specifiers.

In this instance, I suggest doing 
```
ImGui::Text(_L("%s:   %f (%f)"), _L("ValueType"),   val1, val2);
```

tools/l10n/extract_pot.cmake
```diff
@@ -5,6 +5,8 @@ file(GLOB_RECURSE cpp_files "${base_dir}/source/*.cpp" RELATIVE "${base_dir}" LI
 
 set(ror_pot "${base_dir}/languages/ror.pot")
 
+file(WRITE ${ror_pot} "")
```
only-a-ptr, line 8: this isn't "adding _L macro to all strings", it looks like some bugfix.
I'd like it to be in a separate commit

only-a-ptr (changes requested):
This is a monstercommit; most changes are OK but some ruin it.
* translated strings must not be used as `printf()` format strings. This includes `ImGui::Text()`.
* where already translated, I suggest not adding `_LC` context right now to avoid dumping old but working translations... unless Weblate can handle it.
* where not translated yet, I suggest adding `_LC` context right away.
* fixes must go to separate commits

only-a-ptr:
I have an idea... let's use https://fmt.dev/latest/index.html
```
fmt::print("I'd rather be {1} than {0}.", "right", "happy");
```
@AnotherFoxGuy Would you add fmtlib to the build, replace all sprintf()s and then add the _L macros?

source/main/gui/panels/GUI_FrictionSettings.cpp
```diff
@@ -43,70 +43,70 @@ void FrictionSettings::Draw()
...
-    this->DrawTooltip(_L("Drag anisotropy"), _L("This parameter is for making it easier(cheating) to get out from mud. To get stuck in real mud isn't fun at all, so this makes the mud push up. Ranges in this parameter are from 0 to 1 . If you set it at 1 then you'll get real mud. For values from 0 to 1, the behavior goes from real mud to easy mud depending on this parameter and the value of Adhesion velocity. For velocity 0 real mud it is. For velocity >= adhesion velocity easy mud it is."));
+    if (ImGui::SliderFloat(_LC("FrictionSettings", "Drag anisotropy:"), &gm.drag_anisotropy, 0.f, 1.f)) { dirty = true; }
+    this->DrawTooltip(_LC("FrictionSettings", "Drag anisotropy"), _LC("FrictionSettings", "This parameter is for making it easier(cheating) to get out from mud. To get stuck in real mud isn't fun at all, so this makes the mud push up. Ranges in this parameter are from 0 to 1 . If you set it at 1 then you'll get real mud. For values from 0 to 1, the behavior goes from real mud to easy mud depending on this parameter and the value of Adhesion velocity. For velocity 0 real mud it is. For velocity >= adhesion velocity easy mud it is."));
```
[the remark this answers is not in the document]
AnotherFoxGuy: The translations will be kept, but they will be marked as `Strings needing action`
The only thing translators will have to do it check if the translation is still correct and remove the `Needs editing` flag

source/main/gui/panels/GUI_GameMainMenu.cpp
```diff
@@ -183,3 +181,7 @@ void GameMainMenu::DrawVersionBox()
     ImGui::PopStyleColor(1);
 }
 
+bool GameMainMenu::HighlightButton(const std::string& txt,ImVec2 btn_size, int index) const{
+    std::string button_txt = (m_kb_focus_index == index) ? fmt::format("--> {}  <--", txt) : txt;
```
only-a-ptr, line 185: the caption is de-centered (1space on left, 2 on right). Otherwise, :heart: this function

source/main/gui/panels/GUI_GamePauseMenu.cpp
```diff
@@ -110,3 +110,7 @@ void GamePauseMenu::Draw() // TODO: Copypaste of 'GameMainMenu' -- cleanup and u
     m_kb_enter_index = -1;
 }
 
+bool GamePauseMenu::HighlightButton(const std::string& txt,ImVec2 btn_size, int index) const{
+    std::string button_txt = (m_kb_focus_index == index) ? fmt::format("--> {}  <--", txt) : txt;
```
only-a-ptr, line 114: ditto

only-a-ptr (approved):
Good work! Just address the few comments please.

PR #2648, opened 2021-01-12
👷 Upload redist folder as artifact

AnotherFoxGuy:


tritonas00:
Nice! Two issues with the linux zip

- RunRoR and bin/RoR are not executable (missing permissions)
- Our extra libs are missing https://github.com/RigsOfRods/rigs-of-rods/blob/master/tools/CI/upload-CI-build.sh#L7

AnotherFoxGuy:
> RunRoR and bin/RoR are not executable (missing permissions)

That is a limitation of the upload-artifact action
https://github.com/actions/upload-artifact#permission-loss

> Our extra libs are missing

This should be fixed now

tritonas00:
permissions now fixed :+1: 

some libs are not correct though
`libzzip-0.so.13.0.62`
`libHalf.so.12.0.0` 
`libpng16.so.16.34.0`
`libraw.so.16.0.0`

are executable, not libs, so RoR cant work them

```
./bin/RoR: error while loading shared libraries: libzzip-0.so.13: cannot open shared object file: No such file or directory
```

tritonas00:
First launch got
```
./bin/RoR: error while loading shared libraries: libzzip-0.so.13: cannot open shared object file: No such file or directory
```
renaming `libzzip.so` symlink to `libzzip-0.so.13` fixed it RoR starts but complains
```
ItemIdentityException: Can not find codec for 'png' image format.
```
iirc we should have `libopenjp2-7` to fix this issue on ubuntu so `apt-get install libopenjp2-7` and cp that lib also should fix the issue

tritonas00:
```
string(REGEX MATCHALL "/lib/x86_64-linux-gnu/([A-Za-z0-9.]+.so[0-9.]*)" reg_match "${linked_libs}")
```
should be 
```
string(REGEX MATCHALL "/usr/lib/x86_64-linux-gnu/([A-Za-z0-9.]+.so[0-9.]*)" reg_match "${linked_libs}")
```
i think

tritonas00:
same as https://github.com/RigsOfRods/rigs-of-rods/pull/2648#issuecomment-761576903

tritonas00:
Sorry i'm an idiot, i downloaded yesterday's zip.

Now RoR segfaults and that's normal, you fetch core libs now (like libX11) and it will never going to work.

Solution is to cp our certain libs like before, correct libzzip-0.so.13 name (which is fixed in the newer zip) and add the libopenjp2-7 package and cp that lib also

tritonas00:
```
./bin/RoR: error while loading shared libraries: libzzip-0.so.13: cannot open shared object file: No such file or directory
```
now the lib is missing entirely

tritonas00:
good work :+1: 

kk

PR #2656, opened 2021-01-20
Don't process incoming network data when paused

tritonas00:
Fixes crash when we are in paused menu and new clients come in/out. Data processing resumes when unpaused.

### DIFF
--- a/source/main/main.cpp
+++ b/source/main/main.cpp
@@ -660,7 +660,7 @@ int main(int argc, char *argv[])
 
 #ifdef USE_SOCKETW
             // Process incoming network traffic
-            if (App::mp_state->GetEnum<MpState>() == MpState::CONNECTED)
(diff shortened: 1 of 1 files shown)


only-a-ptr:
Since when is pausing possible in MP?

tritonas00:
eeh, a long time i think

only-a-ptr:
It's still weird. And even weirder with this change - the player will "pause" other players 😮

tritonas00:
i paused player one (presses escape), moved and spawned a car in player two
scene in player one updated as soon as exited pause menu again

but i will test more

tritonas00:
> the player will "pause" other players

tested again. the player pause himself to other players, as expected, other players continue normally

tritonas00:
https://github.com/RigsOfRods/rigs-of-rods/pull/2658

PR #2563, opened 2020-05-14
Making sourcecode look more mature

only-a-ptr:
A long overdue internal tidy-up of duplicities, misplaced/fragmented logic and other irregularities in the codebase.

Story:
Recently I broke FOV setting (again!) and soon after savegames (terrain change). Both had the same reason: our 3-state cvars are too complicated (my fault, I designed them). I decided to "flatten" them to 1 value once and for all. That's a change across most of the codebase, so while I was at it, I also decided to tidy up all the ugly things that I was passing by for a long time.

Changes:
* Done: Use only 1 main loop + message queue to change game state
* Done: Transform cvars to contain only 1 value instead of current 3 (active/pending/stored) - only used in few of them, complicated.
* Done: Unify global object management - remove singletons, use only `App::GetFoo()`
* Done: Removed `gEnv` global object
* Done: Upgrade DearIMGUI to v1.75 - ported from AngelHUD branch.
* Done: Eliminate `class SimController` (RoRFrameListener.h|cpp) - it's a very old mashup of feats (direction arrow, race times, actor helpers) which need a better place to live.
* Done: rename files to match content (Beam.h/cpp, BeamFactory.h/cpp, SkinManager.h/cpp)
* Done: Split header BeamData.h (old name) to PhysicsData.h (also merge datatypes/*.h files) and GfxData.h - [merge from AngelHUD branch](https://github.com/RigsOfRods/rigs-of-rods/pull/2490/files#diff-8a159d830251a5ba8c667a278eb15131)
* Done: move terrain editor from gameplay/RoRFrameListener to terrain/TerrainEditor

PR #2657, opened 2021-01-21
Fixed re-spawning last truck enables blinker

tritonas00:
Spawn a truck and press `ctrl .` to respawn it -> right blinker turns on (assigned to `.`)

It was fixed before (https://github.com/RigsOfRods/rigs-of-rods/pull/2200) but it seems it re introduced with https://github.com/RigsOfRods/rigs-of-rods/pull/2563


only-a-ptr:
This isn't right, and Ulteq's original fix wasn't either.

If there's a conflict in default controls, default controls needs to change. If the user assigned conflicting controls themselves, they wouldn't report it as a bug either.

In this case, I think simply changing the blinker binding to `EXPL+PERIOD` woulf fix the issue.

tritonas00:
> I think simply changing the blinker binding to EXPL+PERIOD woulf fix the issue.

Indeed, works fine!

tritonas00:
will be included in https://github.com/RigsOfRods/rigs-of-rods/pull/2658

PR #2658, opened 2021-01-23
Minor fixes and tweaks

only-a-ptr:


only-a-ptr:
Anything left to fix here?

tritonas00:
Don't think so, everything looks solid so far

PR #2669, opened 2021-02-07
Removed extra username text from join/leave messages

tritonas00:
Requested from discord

Before:
kk
After:
kk


tritonas00:
wtf `--force` is broken?

will create a new one

Issue #2665, opened 2021-02-03
Friction setting menu become corrupted.


WillyB1337:
To reproduce :

1. Load a map (Any map)
2. Go back to menu
3. Either press on **Resume Game** or load another map.

![image](https://user-images.githubusercontent.com/16364065/106771139-a8a5a200-660c-11eb-9318-a24ac4e0b069.png)
![image](https://user-images.githubusercontent.com/16364065/106771333-d7bc1380-660c-11eb-9c17-c9b87c60eed7.png)

My theory is that it's loading other parts of memory that it's supposed to not do.

I'm surprised nobody found and reported that issue yet...

PR #2671, opened 2021-02-07
Fixed corrupted ground type entries on resume

tritonas00:
Make sure the vector is empty before adding things in it

Fixes https://github.com/RigsOfRods/rigs-of-rods/issues/2665


only-a-ptr (approved):
Perfect. Thanks!

PR #2673, opened 2021-02-08
Removed extra username text from join/leave messages

tritonas00:
Requested from discord

Before:
kk
After:
kk

Also cleaned up extra leading spaces in `putNetMessage` messages since we already have the required space in https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/gui/panels/GUI_ConsoleView.cpp#L188

Now all `putNetMessage` messages are consistent


only-a-ptr (approved):
Looks good.

PR #2680, opened 2021-02-12
Fixed crash with SkipMainMenu

tritonas00:
Reported from discord

Fixes segfault when SkipMainMenu option is enabled

PS: The only way to make it set correct window position and size was `ImGuiCond_Appearing`, @only-a-ptr  is there a better way?

PS2: I think i got it

### DIFF
--- a/source/main/gui/panels/GUI_MainSelector.cpp
+++ b/source/main/gui/panels/GUI_MainSelector.cpp
@@ -72,8 +72,8 @@ void MainSelector::Draw()
     GUIManager::GuiTheme const& theme = App::GetGuiManager()->GetTheme();
 
     ImGuiWindowFlags win_flags = ImGuiWindowFlags_NoCollapse;
-    ImGui::SetNextWindowPosCenter(ImGuiCond_FirstUseEver);
-    ImGui::SetNextWindowSize(ImVec2((ImGui::GetIO().DisplaySize.x / 1.4), (ImGui::GetIO().DisplaySize.y / 1.2)), ImGuiCond_FirstUseEver);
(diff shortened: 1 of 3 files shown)


only-a-ptr (changes requested):
Tricky situation.

This fix pretty much works except it ignores forced regen/update from RoR.cfg or command line. Those are done with `MSG_APP_MODCACHE_*_REQUESTED`. You could do it directly but that would duplicate with those messages, and I want it unified. So basically you can't show the selector before the message loop starts processing.

Conclusion: instad of doing `App::GetGuiManager()->GetMainSelector()->Show()`, we need to start doing `MSG_GUI_SHOW_SELECTOR_REQUESTED`. That way everything will be in sequence again.

I don't get the imgui code: `ImGuiCond_FirstUseEver` is the same as `ImGuiCond_Appearing` except done only once.

tritonas00:
> I don't get the imgui code: ImGuiCond_FirstUseEver is the same as ImGuiCond_Appearing except done only once.

when SkipMainMenu is enabled, using ImGuiCond_FirstUseEver the selector opens on the upper left corner and with minimal size, while with ImGuiCond_Appearing opens correctly

only-a-ptr:
> when SkipMainMenu is enabled, using ImGuiCond_FirstUseEver the selector opens on the upper left corner and with minimal size, while with ImGuiCond_Appearing opens correctly

That's not what's supposed to happen. Looks like our integration doesn't work quite right if you use the flags before first frame is rendered.

only-a-ptr:
Closing in favor of #2684

PR #2684, opened 2021-02-13
Added  MSG_GUI_[OPEN/CLOSE]` for main menu and selector.

only-a-ptr:
Fixes issue with "Skip main menu" setting not working.
Remake of #2680


tritonas00:
it opens like

kk

only-a-ptr:
> it opens like

I tried all scenarios. Must be linux-specific. Try adding this before the message loop:
```
// Hack - force rendering image to properly init DearIMGUI
App::GetGuiManager()->GetLoadingWindow()->SetProgress(100, "Hack", /*renderFrame=*/true);
App::GetGuiManager()->SetVisible_LoadingWindow(false);
```

tritonas00:
> Try adding this before the message loop:

worked fine! :+1:

tritonas00 (approved):
Main Menu and Selector seem to show/hide when they should.

PR #2685, opened 2021-02-14
Dont load savegame with different terrain in MP

tritonas00:
Reported from discord.

Loading a savegame with different terrain than the running one while on multiplayer causes unpredictable behavior (server will kick you for too many actors or if it loads other players can't see you)

With this PR it loads only the savegame which matches the running terrain or else:
kk


only-a-ptr (approved):
Very good fix, thanks!

PR #2686, opened 2021-02-16
Terrain editor fixes

tritonas00:
Reported from discord

- Fixed ImGui mouse index (middle button -> 2)
- Fixed closing editor (`UpdateSimInputEvents` was running only in `RUNNING` state)


only-a-ptr (approved):
Looks good

PR #2687, opened 2021-02-17
Enlarged selector buttons

tritonas00:
For easier catching

kk


only-a-ptr:
Good idea. In fact, I'd go even further and put a wide/tall [OK] button on the bottom right and shorter/narrower [Cancel] on the left side. I'll leave it up to you.

PR #2688, opened 2021-02-17
Disconnect if terrain not found

tritonas00:
Reported from discord

Disconnect If we are trying to connect to a server and terrain is not found


only-a-ptr (approved):
Correct fix, but for more elegance please move the disconnect-snippet to main.cpp, to the else branch of `if (LoadTerrain())`, so that all network manipulations are consistently done from there.

Thanks.

PR #2692, opened 2021-02-19
Fixed incorrect reverse light sound updating over network

tritonas00:
Reported from discord:

> there’s this bug with the reverse beep being triggered every frame when the remote vehicle engine is not running

### DIFF
--- a/source/main/physics/Actor.cpp
+++ b/source/main/physics/Actor.cpp
@@ -618,7 +618,7 @@ void Actor::CalcNetwork()
     else
         SOUND_STOP(ar_instance_id, SS_TRIG_HORN);
 
-    if (m_net_reverse_light)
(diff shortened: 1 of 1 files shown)


only-a-ptr:
this way, the sound will never play. Try this example program at http://cpp.sh/:
```
    bool b0 = false;
    bool b1 = true;
    const int mask = 8; // NETMASK_REVERSE
    
    if (s.b0 & mask) std::cout << "b0 " << std::endl;
    if (s.b1 & mask) std::cout << "b1 " << std::endl;
    std::cout << "finished" << std::endl;
```

tritonas00:
tried it, i get the same sound when the engine is not running

tritonas00:
So in single player player the reverse light sound plays only if engine is running, made it that way on MP also.

only-a-ptr (approved):
Looks good.

PR #2694, opened 2021-02-21
Fixed main menu not showing after connect failure

tritonas00:
Main menu not showing when https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/network/Network.cpp#L427 happens


only-a-ptr (approved):
Looks good

PR #2697, opened 2021-02-27
Fixed black username color on black background

tritonas00:
Reported from discord.

Happened when online players are more than our colors (25) https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/network/Network.cpp#L98

Before
bef
After
after

`ZERO` worked when client list panel/chat was an overlay and had white background


only-a-ptr (approved):
Perfect

PR #2710, opened 2021-03-15
Unified Main/Pause menu

tritonas00:
Based on comment in https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/gui/panels/GUI_GamePauseMenu.cpp#L42

Changes:

- Removed `GUI_GamePauseMenu(cpp/h)`
- Merged code to `GUI_GameMainMenu.cpp` and run by checking `AppState`
- Pressing Esc on multiplayer, only the menu is shown now, simulation continues
- Keep "Single player" button in Single player, changes map
- Keep "Multiplayer" button in Multiplayer, changes server
- Added "Cache updated" notice:
kk


source/main/GameContext.cpp
```diff
@@ -736,12 +736,20 @@ void GameContext::UpdateGlobalInputEvents()
...
+                {
+                    this->PushMessage(Message(MSG_SIM_PAUSE_REQUESTED));
+                }
+                this->PushMessage(Message(MSG_GUI_OPEN_MENU_REQUESTED));
             }
             else if (App::sim_state->GetEnum<SimState>() == SimState::PAUSED)
             {
                 this->PushMessage(Message(MSG_SIM_UNPAUSE_REQUESTED));
             }
+            if (App::GetGuiManager()->IsVisible_GameMainMenu())
```
only-a-ptr, line 749: this check is redundant - you should put the GUI_CLOSE_MENU inside the `if PAUSED` condition above. 

you may also wrap the MSG_UNPAUSE in `!CONNECTED` check - that way, it will also be nicely consistent with the `if RUNNING` condition above.

source/main/gui/panels/GUI_GameMainMenu.cpp
```diff
@@ -84,58 +108,97 @@ void GameMainMenu::DrawMenuPanel()
...
+                this->SetVisible(false);
+                if (App::diag_preset_terrain->GetStr().empty())
+                {
+                    RoR::Message m(MSG_GUI_OPEN_SELECTOR_REQUESTED);
+                    m.payload = reinterpret_cast<void*>(new LoaderType(LT_Terrain));
+                    App::GetGameContext()->PushMessage(m);
+                }
+                else
+                {
+                    App::GetGameContext()->PushMessage(Message(MSG_SIM_LOAD_TERRN_REQUESTED, App::diag_preset_terrain->GetStr()));
```
only-a-ptr, line 135: this will do nothing because the current terrain is already the preset terrain. I suggest to daisplay the selector anyway.

source/main/gui/panels/GUI_GameMainMenu.cpp
```diff
@@ -84,58 +108,97 @@ void GameMainMenu::DrawMenuPanel()
...
             if ( HighlightButton(_LC("MainMenu", "Resume game"), btn_size, button_index++))
             {
-                App::GetGameContext()->PushMessage(Message(MSG_SIM_LOAD_SAVEGAME_REQUESTED, "autosave.sav"));
+                if (App::app_state->GetEnum<AppState>() == AppState::MAIN_MENU)
+                {
+                    App::GetGameContext()->PushMessage(Message(MSG_SIM_LOAD_SAVEGAME_REQUESTED, "autosave.sav"));
+                }
+                else
+                {
+                    App::GetGameContext()->PushMessage(Message(MSG_SIM_UNPAUSE_REQUESTED));
```
only-a-ptr, line 150: this line shouldn't be inside `if FileExists()` check :)

source/main/gui/panels/GUI_GameMainMenu.cpp
```diff
@@ -84,58 +108,97 @@ void GameMainMenu::DrawMenuPanel()
...
+        if (App::app_state->GetEnum<AppState>() != AppState::MAIN_MENU && App::mp_state->GetEnum<MpState>() != MpState::CONNECTED)
         {
-            App::GetGuiManager()->SetVisible_MultiplayerSelector(true);
-            this->SetVisible(false);
+            if ( HighlightButton(_LC("MainMenu", "Change map"), btn_size, button_index++))
+            {
+                if (App::diag_preset_terrain->GetStr().empty())
+                {
+                    this->SetVisible(false);
+                    App::GetGameContext()->PushMessage(Message(MSG_SIM_UNLOAD_TERRN_REQUESTED));
```
only-a-ptr, line 163: this is correct but not what I had in mind. I'd like the map selector to be displayed on top of the simulation and only send UNLOAD message if user presses OK. also, we may reuse the "single player" button, see the other comment.

source/main/gui/panels/GUI_GameMainMenu.cpp
```diff
@@ -84,58 +108,97 @@ void GameMainMenu::DrawMenuPanel()
...
         {
-            this->SetVisible(false);
-            if (App::diag_preset_terrain->GetStr().empty())
-            {
-                RoR::Message m(MSG_GUI_OPEN_SELECTOR_REQUESTED);
-                m.payload = reinterpret_cast<void*>(new LoaderType(LT_Terrain));
-                App::GetGameContext()->PushMessage(m);
-            }
-            else
+            if (HighlightButton(_LC("MainMenu", "Single player"), btn_size, button_index++))
```
only-a-ptr, line 124: what about keeping this button visible even in simulation? It will open selector for new map, and if cancelled, user stays in current simulation. You'll have to adjust the Selector/OkButton code a bit (to also send UNLOAD and DISCONNECT), but I think it's the way to go.

source/main/gui/panels/GUI_GameMainMenu.cpp
```diff
@@ -84,58 +108,97 @@ void GameMainMenu::DrawMenuPanel()
...
+            }
         }
 
-        if (HighlightButton(_LC("MainMenu", "Settings"), btn_size, button_index++))
+        if (App::app_state->GetEnum<AppState>() == AppState::MAIN_MENU)
         {
-            App::GetGuiManager()->SetVisible_GameSettings(true);
-            this->SetVisible(false);
-        }
+            if (HighlightButton(_LC("MainMenu", "Multiplayer"), btn_size, button_index++))
```
only-a-ptr, line 173: how about keeping this visible during simulation? it would open the MP selector, and if cancelled, the player will remain in current session. The only change needed is to adjust "MpSelector/Join" button code to unload current terrain and disconnect current MP session if needed.

only-a-ptr (commented):
Very good work so far, but I think you can take it even further. See the comments and tell me what you think.

source/main/GameContext.cpp
tritonas00: That way there times that Esc does nothing, the only way i managed getting it to work every time is independent from the those checks and after them:

https://github.com/RigsOfRods/rigs-of-rods/blob/6015013855f977950d2e53ed04e8c766f90bf841/source/main/GameContext.cpp#L749-L756,

tritonas00:
Almost there, need to investigate why opening the selector (terrain) in simulation reverses the pause/menu

only-a-ptr:
About the 'cache updated' notification: https://github.com/RigsOfRods/rigs-of-rods/pull/2718#issuecomment-809964087

tritonas00:
https://github.com/RigsOfRods/rigs-of-rods/pull/2723

PR #2717, opened 2021-03-25
Update cache when content changes

tritonas00:
Reported from discord

Till now RoR updated the cache when content added/removed.

With this PR it also checks the `filetime` for every entry and if is different from the last time it was modified it sets `validity` to `CACHE_NEEDS_UPDATE`.

Means any change inside content (zip. skinzip ...) will now trigger cache update.

Don't know if this is the right way to do it but it works.


only-a-ptr:
yes, it is the way, AND it works 👍

PR #2718, opened 2021-03-26
Added cache updated notice

tritonas00:
Requested from discord

Updating cache can be ultra fast, especially with little content, so the progress window will not even appear in that case (opens/closes ultra fast). This PR adds a notice every time cache is updated so the user knows that its done.

kk

### DIFF
--- a/source/main/resources/CacheSystem.cpp
+++ b/source/main/resources/CacheSystem.cpp
@@ -941,6 +941,7 @@ void CacheSystem::ParseZipArchives(String group)
     }
 
     RoR::App::GetGuiManager()->SetVisible_LoadingWindow(false);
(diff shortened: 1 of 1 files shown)


tritonas00:
https://github.com/RigsOfRods/rigs-of-rods/pull/2710

only-a-ptr:
Reopening because I noticed the two 'thumbsup' reactions above.

I don't really like the #2710 prototype - it feels confusing to have a notification in the same box (same style) as static version info:
image

tritonas00:
> If we want to put the notification above the version box, I'd make a separate box for it and put an icon on the front to make it stand out. And don't forget to hide it next time the user visits the menu.

all done https://github.com/RigsOfRods/rigs-of-rods/pull/2710

PR #2720, opened 2021-03-29
Reduced FPS hit from multiplayer client list window.

only-a-ptr:
Reduces performance hit from the MP client list window. User infos are now cached in the window and only updated when network state changes.


tritonas00 (approved):
Works great!

PR #2721, opened 2021-03-29
Fixed static reflection map being black.

only-a-ptr:
Even if disabled in config, the realtime envmap is fully rendered once after loading terrain, to provide static reflections.

Kudos to NegativeIce for reporting!


tritonas00:
Works fine. Merging so Neg can test also.

tritonas00 (approved):

tritonas00:
I was wrong, now even with reflections disabled in config, i can still see reflections:

kk

tritonas00:
> With "Reflections=No", the reflection map was black.

Was actually:

With "Reflections=Yes , ReflectionUpdateRate=0", the reflection map was black.

PR #2723, opened 2021-03-30
Menu cleanups and improvements

tritonas00:
- Removed `GUI_GamePauseMenu(cpp/h)` and merged code in `GUI_GameMainMenu.cpp`
- Pressing Esc in multiplayer, only the menu is shown now, simulation continues
- Added "Cache updated" notice:
kk


only-a-ptr:
Works great, just one detail: the "cache updated" notification only shows if cache is actually updated. If it's alredy up-to-date, nothing shows at all. I suggest adding "cache already up-to-date" variant of the notification for this case.

tritonas00:
> notification only shows if cache is actually updated. If it's already up-to-date and user presses "update cache" button, nothing shows at all

this is not the case on linux, it is always shown when i hit the update cache button
i can reproduce the issue on windows, i moved the notice to LoadModCache, also the same

only-a-ptr:
@tritonas00 I don't understand why linux behaves differently, but we definitely have a bug: CACHE_STATE_UNKNOWN and CACHE_NEEDS_UPDATE have the same value. Without this bug, the notification would always display after pressing the button.

tritonas00:
Epic

kk

PR #2728, opened 2021-04-06
🚀 Simplified cmake install logic

AnotherFoxGuy:
Changes:
- Simplified cmake install logic, it is no longer required to use `BUILD_REDIST_FOLDER` to generate a build for itch.io.
- Removed the `zip_and_copy_resources` target, cmake is now smart enough not to recopy the files if they haven't changed.
- Cleaned up unused cmake macros
- Organized the generated targets in folders


tritonas00:
It ended up with an empty `bin/resources` dir

AnotherFoxGuy:
@tritonas00 Does cmake generate `build/source/main/CMakeFiles/fast_copy_fonts.dir/build.make`?

tritonas00:
oops, i think i did something wrong, it works

would be nice if the copying was done at the end of compilation though, its a bit messy now

kk

tritonas00 (approved):

tritonas00:
ror-linux.zip issues:

```
./RunRoR 
./RoR: error while loading shared libraries: libOIS.so: cannot open shared object file: No such file or directory
```

tritonas00 (changes requested):
ror-linux.zip issues:

```
./RunRoR 
./RoR: error while loading shared libraries: libOIS.so: cannot open shared object file: No such file or directory
```

tritonas00 (approved):
Works perfectly now.

AnotherFoxGuy:
> would be nice if the copying was done at the end of compilation though, its a bit messy now

That is because the targets are added as a regular build target, so make will run it as a normal compile rule

tritonas00:
OK, just dont forget to update the docs, since `zip_and_copy_resources` will be removed

CMakeLists.txt
```diff
@@ -59,9 +59,9 @@ if (WIN32)
     include_directories(${DirectX_INCLUDE_DIR})
 
     set(CMAKE_EXE_LINKER_FLAGS_DEBUG "${CMAKE_EXE_LINKER_FLAGS_DEBUG}          /DEBUG /LARGEADDRESSAWARE ")
-    set(CMAKE_EXE_LINKER_FLAGS_RELEASE "${CMAKE_EXE_LINKER_FLAGS_RELEASE}        /DEBUG /LARGEADDRESSAWARE /OPT:REF /LTCG:STATUS /RELEASE")
```
only-a-ptr, line 70: We want to keep `/LTCG` - it's "whole program optimization". I haven't measured but theoretically it could make big difference. It's OK to remove the `:STATUS` part tho, that's just a build log setting.

only-a-ptr:
Actually, I only tested now - my previous note wasn't a conclusion. Nevermind.

Project "RoR" doesn't have a dependency on "zip_folder_resources/zip_folder_content" - is it intentional? After running those manually everything works fine.

image
😍

AnotherFoxGuy:
> Project "RoR" doesn't have a dependency on "zip_folder_resources/zip_folder_content" - is it intentional? After running those manually everything works fine.

Yes, this is intentional. If you build the complete solution it should also build the `zip_folder_resources/zip_folder_content` targets

PR #2734, opened 2021-04-11
Removed diag_physics_dt

tritonas00:
The UI code was removed by Petr because big values could lead to bad performance (https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/gui/panels/GUI_TopMenubar.cpp#L413-L420)

Still `diag_physics_dt` could be read from RoR.cfg (`PhysicsTimeStep`), so if a big value was set from a user with previous RoR version it could still lead to bad performance.

This completely removes `diag_physics_dt`, sets `PHYSICS_DT` at default 0.5 ms and eliminates user errors.


only-a-ptr (approved):
It's perfect

PR #2741, opened 2021-04-25
Don't check if chat window is mouse hovered

tritonas00:
Reported from forum:

> I can't steer, move around, change the view, or do any main command if my cursor is placed anywhere near the bottom part of my screen.

https://forum.rigsofrods.org/threads/2021-04-discussion-thread.3128/#post-14973


WillyB1337:
I can confirm this happened to me too.
Was kinda annoying but tolerable. 
Glad someone reported it 😄

Goetterescu:
Thanks for that

only-a-ptr:
Works fine 👍

PR #2747, opened 2021-05-04
Settings UI now uses actual tabs

only-a-ptr:
image

"Update cache" button moved under "General".
Separated "Simulation settings" from "General" to it's own tab "Gameplay".

### DIFF
--- a/source/main/gui/panels/GUI_GameSettings.cpp
+++ b/source/main/gui/panels/GUI_GameSettings.cpp
@@ -69,349 +69,15 @@ void GameSettings::Draw()
     ImGui::SetCursorPosY(ImGui::GetCursorPosY() + 4.f);
     ImGui::Separator();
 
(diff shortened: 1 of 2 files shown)


tritonas00:
Nice. Is it possible for the tabs to not scroll with the rest of the body?

only-a-ptr:
> Nice. Is it possible for the tabs to not scroll with the rest of the body?

Yes, I'll change it.

> ... Render System settings to checkboxes instead of dropdowns?

It works by looping over OGRE's settings list and possible values. I'd keep it as is for simplicity.

only-a-ptr:
image

WillyB1337 (approved):
Nice change!

Zentro (approved):

tritonas00 (approved):

WillyB1337:
I just noticed you can scroll the tabs.
05-mai-21-21h01,25

PR #2749, opened 2021-05-05
Controls UI now editable, and opens from main/pause menus.

only-a-ptr:
image


WillyB1337 (approved):

tritonas00:
Its epic.

Issues:
- Map tab is empty
- Needs to check if buffer is empty because if you enter with empty input box game crashes
- No Esc handler (pressing Esc closes the game instead of the window)
- The actual edit doesn't written to input.map

Suggestion:
```
ImGuiWindowFlags input_flags = ImGuiInputTextFlags_EnterReturnsTrue | ImGuiInputTextFlags_CharsUppercase | ImGuiInputTextFlags_CharsNoBlank;
```
in https://github.com/RigsOfRods/rigs-of-rods/blob/bcca3796448c78f85e7b3acaf96890d1702da60c/source/main/gui/panels/GUI_GameControls.cpp#L69

auto converts everything to uppercase and doesn't allow spaces

When you merge, i will add extra code to handle special keys (like if you press control it will write CTRL etc...)

WillyB1337:
The game is crashing when going into the Controls window in an airplane and boat.

only-a-ptr:
image

I added this panel to reload/save individual .map files. It's a prototype, the "save" button actually writes ".map.dbg" file.

tritonas00:
Thats great! So if you going to support other map files other than keyboard, then maybe should remove the `ImGuiInputTextFlags_CharsUppercase` option, because i see for example `Logitech_Formula_Force_EX_USB.map` contains lower case also: `CAMERA_ROTATE_DOWN             JoystickPov          0 0 South`

only-a-ptr:
Final version:
obrazek
The output:
```
AIRPLANE_AIRBRAKES_FULL            JoystickSlider      0 X 2 XXX
AIRPLANE_AIRBRAKES_LESS            JoystickPov         0 4 East
AIRPLANE_AIRBRAKES_MORE            JoystickAxis        0 6 FOO
AIRPLANE_AIRBRAKES_NONE            JoystickButton      0 8
```
The config formats are described in detail here: https://docs.rigsofrods.org/gameplay/controls-config/#config-file-inputmap

tritonas00:
:heart::heart::heart:

One issue found, example common tab

kk
kk

Those are hardcoded i assume? If yes no problem to show them, just align them

only-a-ptr:
@tritonas00 Thanks for reporting, looks like a crash waiting to happen, will investigate.

Also notes by **XD_DominicXD#0746** on Discord:
 - [ ] add add/delete buttons - it's trivial.
 - [ ] add short tooltips next to the input field (a "?" which displays tooltip when you mouse over). For joystick POV, for example, it will say 
 ```
 Format: <X/Y> <Axis number 0+> <North/East/South/West> <options>
 See https://docs.rigsofrods.org/gameplay/controls-config/#config-file-inputmap
 ```
 - [ ] take a shot (a hour max) at teaching RoR how to open a browser link - would be handy.
-  [ ] take a shot (2hr max) at creating interactive "press a key combo" panel, just for keyboard this time - there was one already in the past (RoRConfig launcher).

only-a-ptr:
Added ability to add/edit/delete multiple bindings.
obrazek
Be careful with the <all active values> view - it lets you remove bindings from any .map file but the bindings you add don't save to any .map file.

only-a-ptr:
I felt the controls panel got too complicated, so I separated it to simple/expert modes. The simple mode works as in every other game - displays only keyboard bindings from "input.map" (or builtins) and offers an interactive "please press a new key" dialog. After user changes something, buttons [save/reset] appear.
obrazek
obrazek

only-a-ptr:
I added handling of the EXPL tag, as discussed on Discord.
obrazek

tritonas00 (approved):

Zentro (approved):

PR #2490, opened 2020-03-04
Project AngelHUD

only-a-ptr:
This is a replacement for our MyGUI-based [custom HUD system](https://docs.rigsofrods.org/vehicle-creation/making-custom-hud/). Goal: Allow modders to create and test custom vehicle HUD in-game using AngelScript + DearIMGUI.

**THIS IS A SNEAK PREVIEW, USE AT YOUR OWN RISK**

image

image

User scenario:
1. enter simulation and create a project using top menubar. Projects are created in "MyDocuments\MyGames\RigsOfRods\projects" as subdirectories, each containing project JSON and truckfiles (dubbed snapshots). Unlike regular mods, all resources within a project are always fully reloaded from disk, allowing modder to add/remove/change data on the go.
image
2. add angelscript file to the truckfile using button "+ Add example script" in projects top menu. Details: https://github.com/RigsOfRods/rigs-of-rods/pull/2490#issuecomment-599122949
image

3. within the angelscript, define function `int setup(string arg)` to be invoked once and function `int loop()` to be invoked on each rendered frame.
4. within the angelscript function, draw the HUD using DearIMGUI, reading data from simbuffer (see example code). The simbuffer contains all data described in https://docs.rigsofrods.org/vehicle-creation/making-custom-hud/ and a lot more, allowing modder to create truly diagnostic GUI - DearIMGUI is packed with very capable widgets and a flexible drawing API. Graphical HUD can be done simply by calling ImGui::Image() with texture name.

Notes:
* These scripts are executed separately from the classic RoR scripting subsystem - the "GameObject" interface is not available. The reason is multithreading - a script drawing HUD (or doing other visuals) should run on a rendering thread, while the classic scripts must run when both sim+rendering(main) threads are halted.
* This is a spinoff from #2486 - may include outdated commits from that branch.

PR #2750, opened 2021-05-07
[Alpha!] Addon API

only-a-ptr:
Addons are mods (like terrains and trucks) which bundle and execute a script. All installed addons are listed in Console window, menu Addons. User must press "Load" button to load an addon. The "Autoload" checkbox is dummy (WIP). The script will remain running until unloaded (TODO).

This is a lightweight edition of project AngelHUD (#2490) - it contains the same DearIMGUI features (WIP) and allows the user to load (TODO: reload) them while program is running.

obrazek

obrazek

## README
```
Addons are AngelScript-powered feature packages for Rigs of Rods.

Addons are mods (ZIPs or directories under 
  'userprofile/Documents/My Games/Rigs of Rods/mods')
  with an INI-syntax config file '*.addon'.

All addons found during cache update are assembled under
  menu "Addons" in the console window (hotkey ~)
  with actions "Load" and "AutoLoad". 

An <.addon> file lists script file names (.as) 
  and entry points (script function names).
  
Addon scripts are initialized once using builtin 'setup()'
  function, the other builtin 'loop()' function is then
  invoked on every rendered frame (in both menu and sim).
```


only-a-ptr:
Superseded by much more lightweight https://github.com/RigsOfRods/rigs-of-rods/pull/2840

PR #2752, opened 2021-05-08
Fixed Settings window always scrolling

tritonas00:
Fixes https://github.com/RigsOfRods/rigs-of-rods/pull/2747#issuecomment-833150030

Code copied from multiplayer selector


only-a-ptr:
Works great 👍

PR #2755, opened 2021-05-19
"Reload current terrain" action.

only-a-ptr:
Works like "reload current vehicle" - reloads all resources in the Zip/directory.
obrazek


WillyB1337 (approved):

tritonas00:
Doesn't play well when connected

kk

tritonas00 (approved):

Issue #2622, opened 2020-11-11
MP player name with special character appear blank on the player head.


WillyB1337:
![image](https://user-images.githubusercontent.com/16364065/98841592-55a9bd80-2416-11eb-9a97-8b8a3f87bd51.png)

only-a-ptr:
Problem identified: RoR correctly keeps the nickname in utf8 'Français' but tries to render it as narrow-text, so the utf8 'ç' (0xC3 0xA7) ([source](https://www.compart.com/en/unicode/U+00E7)) become 2 narrow text characters 'Ã' and '§' ([source](https://en.wikipedia.org/wiki/ISO/IEC_8859-1#Code_page_layout)) but they end up as empty space because the're left out from the used fontdef (see "highcontrast_black" in file "resources/OgreCore/higcontrast.fontdef")

Issue #2497, opened 2020-03-10
Ability to hide other peoples spawned objects


4x4convoy:
Just to clarify this is just a suggestion.

After watching the multiplayer servers for awhile now it seems to be a running theme of players complaining that there are other players in their way and causing lag. Would it be possible to make it so you can click on the persons name and turn off the rendering for their spawned objects so they will not be shown at all? This would cut down on players telling other players to leave the server or move away since they're all jumbled together.

Issue #479, opened 2015-12-12
Multiplayer nicknames improvements 


Speciesx:
- Nicknames should not be reflected on cars or at all. (Performance impact?)
- Add the ability to hide own nickname.

![ror_2015_12_12_12_40_53_525](https://cloud.githubusercontent.com/assets/1196375/11761908/ee514c54-a0d3-11e5-8357-8f17a6234667.jpg)

TechGiga:
I agree there shouldn't be a reflection, but if there's no nametag, nobody will know who crashed into them or so on.

Hotrod55:
"Add the ability to hide <b>own</b> nickname."

Meaning other people will still see your nickname, but there would be a option to hide your nickname ;)

TechGiga:
Oh, ok lol.

Hotrod55:
I've changed the font + color of the player labels to be basically identical to 0.36.x's labels, I think it looks a lot better than the current one IMO.

The chat log from Gitter with @only-a-ptr :

**Michael10055 13:42**
>Changed the multiplayer player font color to White, looks a lot better IMO.
![1](http://i.imgur.com/AEpc2Uj.jpg)

**Petr Ohlídal 14:23**
>The multiplayer labels would look better, both on screen and in sourcecode, if they were handled as >on-screen elements rather than floating text. This way, you either use ridiculously detailed font >texture, or it looks bad almost all the time. In code, it looks bad all the time 
>On your screenshot it sure looks better, but what about bright sunny scenarios? We need to figure >some subtitle-style white text with thick black outline

**Michael10055 16:24**
>What about this? It's basically identical to 0.36.x's labels:
![2](http://i.imgur.com/NCMU4tb.jpg)

![3](http://i.imgur.com/LKifsrn.png)

>(0.36.x)

**Petr Ohlídal 18:01**
>Very nice text. How did you do that?
>However, as I said, there's the "floating text" problem - it gets too big for close objects and too tiny >for distant objects (or is there a lower/upper limit? IDK, really).

**Michael10055 18:07**
```
-- Beam.cpp line 5548-5553 ---

            netMT->setFontName("highcontrast_black");
            netMT->setTextAlignment(MovableText::H_CENTER, MovableText::V_ABOVE);
            //netMT->setAdditionalHeight(2);
            netMT->showOnTop(false);
            netMT->setCharacterHeight(2);
            netMT->setColor(ColourValue::White);
```
```
[...]

PR #2760, opened 2021-05-27
Multiplayer improvements

tritonas00:
You can now remove your own actors and hide/unhide remote actors (useful if someone lags you) from top menu bar in multiplayer.

kkk

Also net labels remade using ImGui.

kk

Closes https://github.com/RigsOfRods/rigs-of-rods/issues/2497
Fixes https://github.com/RigsOfRods/rigs-of-rods/issues/2622
Fixes https://github.com/RigsOfRods/rigs-of-rods/issues/479

Thanks Petr for the help!

### DIFF
--- a/source/main/physics/air/TurboProp.cpp
+++ b/source/main/physics/air/TurboProp.cpp
@@ -208,6 +208,11 @@ void Turboprop::updateVisuals(RoR::GfxActor* gfx_actor)
 #endif
 }
 
+void Turboprop::setVisible(bool visible)
+{
+    smokePS->setVisible(visible);
+}
+
 void Turboprop::updateForces(float dt, int doUpdate)
 {
     if (doUpdate)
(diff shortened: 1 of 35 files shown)


only-a-ptr:
Oops.
glitch
I'm coding a fix

only-a-ptr:
As usual, this was harder than it should be. Please cherrypick the last 2: https://github.com/only-a-ptr/rigs-of-rods/commits/mp-actor-list

only-a-ptr:
The only thing I found left is MP labels. While testing I noticed Unicode is completely broken - characters don't display at all. To kill 2 bugs with one blow, I made a prototype of drawing the labels with DearIMGUI.

@tritonas00  you'll have to pick from here...
* to fix size, use the one calculated but multiply with e.g. 40. the original value is 0 - 0.6, but imgui needs it in pixels.
* to give the text some highlight, use `drawlist->AddQuad()` to put darker rectangle around the text.
* Finally, text centering needs to be fixed, currently drawn all to the right from the correct spot.

:zzz:

tritonas00:
> The only thing I found left is MP labels

and turbojet/props 

Nice one with imgui labels, will definitely check it out

only-a-ptr:
obrazek
I centered and visually improved the labels.

But I also noticed a bug (in master) - when you join while existing player is already in vehicle, the rorbot mesh will not pair correctly.

source/main/physics/air/TurboProp.cpp
```diff
@@ -208,6 +208,11 @@ void Turboprop::updateVisuals(RoR::GfxActor* gfx_actor)
 #endif
 }
 
+void Turboprop::setVisible(bool visible)
```
tritonas00, line 211: Needs deleting i guess

only-a-ptr: My mistake. Updated.

PR #2764, opened 2021-06-21
"Overdash" HUD system remake

only-a-ptr:
Right now there are 2 HUD systems in place:
* Hardcoded Ogre overlays - the original approach, used only for airplanes at the moment. Not customizable.
* User-made MyGUI layouts - allows modders to create custom HUD by creating .layout files and referencing them from .truck files. Used for builtin truck/boat UIs.

**The goals:**
* Ditch MyGUI because it's buggy and blocks us from upgrading Ogre to 1.12 and beyond.
* Remake the custom-dash system to use customized Ogre overlays. For details, see new README: https://github.com/only-a-ptr/rigs-of-rods/blob/overdash/source/main/gui/overlay/DashReadMe.txt

**Extra measures:**
Unfortunately, Ogre is presently unable to unload/reload overlay scripts. Until that is fixed, the new overlay-based dashboards will not be available for truck modding, only for creating builtin HUDs. To track the progress, see this issue ticket: https://github.com/OGRECave/ogre/issues/1335.


only-a-ptr:
I'm closing this in favor of keeping the current MyGUI-based system:
* It works, even with OGRE13.
* It has a functional render-to-texture feature.
* We have GUI editor and docs for it (https://docs.rigsofrods.org/vehicle-creation/making-custom-hud/).
* We even have a few existing mods for it.

PR #2773, opened 2021-08-03
Draw notices with same prefix in same line

tritonas00:
For cleaner output and less spam.

https://user-images.githubusercontent.com/2660424/127962064-49ee639b-b312-4188-970e-54e1a943502c.mp4


only-a-ptr (approved):
Spiffy. Improves the UI experience a lot.

The code works by comparing the first word (before space) of the message (in target language) so there's some risk of clashes. In the future we may need to add some message codes. But for now it's just fine.

Issue #2771, opened 2021-07-26
RigDef::Parser::GetArgFloat broken


luc65r:
`std::strtod` is locale dependent, so `RigDefParser::GetArgFloat` is broken in locales where the decimal-point character is `,`.
Fixed by setting `LC_ALL=C`.

Why this is a problem:
```
[RoR|Actor|Warning] Daf Semi truck (line 5 'globals'): Argument [1] (type: float) has invalid trailing characters (".0")
[RoR|Actor|Warning] Daf Semi truck (line 5 'globals'): Argument [2] (type: float) has invalid trailing characters (".0")
[RoR|Actor|Warning] Daf Semi truck (line 13 'engine'): Argument [1] (type: float) has invalid trailing characters (".0")
[RoR|Actor|Warning] Daf Semi truck (line 13 'engine'): Argument [2] (type: float) has invalid trailing characters (".0")
[RoR|Actor|Warning] Daf Semi truck (line 13 'engine'): Argument [3] (type: float) has invalid trailing characters (".0")
[RoR|Actor|Warning] Daf Semi truck (line 13 'engine'): Argument [4] (type: float) has invalid trailing characters (".00")
[RoR|Actor|Warning] Daf Semi truck (line 13 'engine'): Argument [5] (type: float) has invalid trailing characters (".85")
[RoR|Actor|Warning] Daf Semi truck (line 13 'engine'): Argument [6] (type: float) has invalid trailing characters (".86")
[RoR|Actor|Warning] Daf Semi truck (line 13 'engine'): Argument [7] (type: float) has invalid trailing characters (".52")
[RoR|Actor|Warning] Daf Semi truck (line 13 'engine'): Argument [8] (type: float) has invalid trailing characters (".56")
[RoR|Actor|Warning] Daf Semi truck (line 13 'engine'): Argument [9] (type: float) has invalid trailing characters (".48")
[RoR|Actor|Warning] Daf Semi truck (line 13 'engine'): Argument [10] (type: float) has invalid trailing characters (".58")
[RoR|Actor|Warning] Daf Semi truck (line 13 'engine'): Argument [11] (type: float) has invalid trailing characters (".83")
[RoR|Actor|Warning] Daf Semi truck (line 13 'engine'): Argument [12] (type: float) has invalid trailing characters (".02")
[RoR|Actor|Warning] Daf Semi truck (line 13 'engine'): Argument [13] (type: float) has invalid trailing characters (".53")
[...]

PR #2775, opened 2021-08-11
Hotfix: Number parsing in truck file was sensitive to locale (regional settings) 

only-a-ptr:
see #2771


tritonas00:
All good here :+1:

tritonas00 (approved):

PR #2780, opened 2021-08-22
Added AngelScript shortcuts in console

tritonas00:
For convenient use
kk


only-a-ptr (commented):
This is an excellent concept which is about to become more and more important in the future.

I made a couple improvements and I'm happy with it. Now it needs to be reviewed by original author. :D

PR #2784, opened 2021-08-28
Hotfix: only load prop mesh from own resource group.

only-a-ptr:
**Problem**: when 2 mods (each with it's own bundle) contain mesh with the same file name, they will clash with each other. The first mod loaded will override the mesh for all others until the game is restarted. This applies to props (vehicle) and odefs (terrain).

**Cause**: class MeshObject was searching for the mesh name globally instead of using the mod's resource group. This is classic behavior, some mods might rely on it.

**Fix**: use the mod's resource group, as already done for most resources.

Mods I used for testing:
[agora-meshmix1.zip](https://github.com/RigsOfRods/rigs-of-rods/files/7069858/agora-meshmix1.zip)
[agora-meshmix2.zip](https://github.com/RigsOfRods/rigs-of-rods/files/7069859/agora-meshmix2.zip)

### DIFF
--- a/source/main/utils/MeshObject.cpp
+++ b/source/main/utils/MeshObject.cpp
@@ -69,9 +69,7 @@ void MeshObject::createEntity(Ogre::String meshName, Ogre::String entityRG, Ogre
 
     try
     {
-        // Classic behavior: look in all resource groups.
-        //   The mesh may live in terrain's resource bundle (aka ZIP) or ror's bundled resource ZIPs.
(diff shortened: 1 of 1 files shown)


tritonas00:
It fixes train valley trains also:

Master
screenshot_2021-08-28_12-20-20_1
This
screenshot_2021-08-28_12-21-01_1

tritonas00 (approved):

PR #2790, opened 2021-09-07
Remade race text overlay with ImGui

tritonas00:
Also lap times color varies depending the time difference, like before.

kk
kkk
kkkk


only-a-ptr:
I integrated it with top menubar. When you hover mouse over the top it shows like this:
obrazek

only-a-ptr (approved):
Nice work

tritonas00:
TODO: bring back lap times

only-a-ptr (approved):
very good

Issue #2745, opened 2021-04-28
Races not working correctly in 2021.02/.04


only-a-ptr:
From Discord/game-support:

> Xploder98
 — 
Today at 5:56 PM
Hello everybody. I think I might have finally figured out why races on some terrains haven't been working in the latest two releases!
Most of the starting checkpoints for the terrains with non-working races have `event checkpoint` listed in their ODEFs. The terrains with working races have checkpoints that explicitly list `event racestart truck` or just `event racestart` in their ODEFs.
Maybe previous RoR versions implicitly converted `event checkpoint` to `event racestart` for the first listed checkpoint in a race?

Possibly related to #2703 - I rewrote ODEF parser and broke stuff.

Races which don't work:
* https://forum.rigsofrods.org/resources/ladoga.160/
* https://forum.rigsofrods.org/resources/hillside.541/
* https://forum.rigsofrods.org/resources/desert-rally-map.646/

PR #2796, opened 2021-09-16
Fixed races not loading when event type is not defined

tritonas00:
Old code was:
kk
https://github.com/RigsOfRods/rigs-of-rods/commit/2046c336068ffd0e024c6b5710fe2113c14cf559

Fixes https://github.com/RigsOfRods/rigs-of-rods/issues/2745

@only-a-ptr Is this correct or we need to be specific? Like
```
if strncmp(ev_type, "",  0)
```


only-a-ptr:
> @tritonas00 Is this correct or we need to be specific? like `if strncmp(ev_type, "",  0)`

Let's leave it like it is, I think it's the best solution.

The old parser was somewhat messy here - If the filter-string was bogus but not empty, the filter would be set to value of previous event filter (or EVENT_ALL for the very first event line). But I think this is an old bug and it really should behave the way you made it just now.

only-a-ptr (approved):
💯

Issue #2767, opened 2021-06-24
Tab map rendering black under "Bilinear" texture filtering.


WillMichals:
Tab map renders full black with Bilinear texture filtering.

Screenshots attached and RoR log.

![image](https://user-images.githubusercontent.com/51508848/123206580-90320900-d481-11eb-8299-1fb6646c3071.png)
![image](https://user-images.githubusercontent.com/51508848/123206643-a8098d00-d481-11eb-9121-c5982e93c526.png)
[RoR (2).log](https://github.com/RigsOfRods/rigs-of-rods/files/6706434/RoR.2.log)

tritonas00:
Can't reproduce it on linux. Other modes work? (none, trilinear, anisotropic - you need to restart the game to take effect)

WillMichals:
> Other modes work? (none, trilinear, anisotropic - you need to restart the game to take effect)

Every other mode fixes the issue. Bilinear is the only mode that produces this.

tritonas00:
we should just disable filtering in survey map, i will look into this

tritonas00:
Cant reproduce it on windows either (vbox)

tritonas00:
https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/gui/panels/GUI_SurveyMap.cpp#L153

For some reason it renders like
```
ImGui::Image(reinterpret_cast<ImTextureID>(tex->getHandle()), ImVec2(0, 0));
```

PR #2797, opened 2021-09-21
Survey map fixes

tritonas00:
Fixes https://github.com/RigsOfRods/rigs-of-rods/issues/2767

For some reason when texture filtering is set to Bilinear, with some gpus, `ImGui::GetIO().DisplaySize.y` returns 0
Old code was: https://github.com/RigsOfRods/rigs-of-rods/blob/2020.01/source/main/terrain/map/SurveyMapManager.cpp#L185

Also fixes darker image in Big map with Anisotropic filtering

Before:
bef
After:
after

@paroj any idea why those nasty things happen? :rofl:

### DIFF
--- a/source/main/gui/panels/GUI_SurveyMap.cpp
+++ b/source/main/gui/panels/GUI_SurveyMap.cpp
@@ -70,10 +70,11 @@ void SurveyMap::Draw()
     }
 
     // Calculate window position
+    Ogre::RenderWindow* rw = RoR::App::GetAppContext()->GetRenderWindow();
     ImVec2 view_size(0, 0);
     if (mMapMode == SurveyMapMode::BIG)
     {
-        view_size.y = ImGui::GetIO().DisplaySize.y -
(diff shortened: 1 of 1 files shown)


only-a-ptr:
@paroj I think this problem may affect Ogre as well since our imgui integration is a port of imgui-overlay from Ogre 1.12

This is where we fill `DisplaySize`: https://github.com/RigsOfRods/rigs-of-rods/blob/c5ee81ab23dd864382d9501bdb38634c8699f9fc/source/main/gui/imgui/OgreImGuiOverlay.cpp#L139-L140

Ogre 13's code is equivalent: https://github.com/OGRECave/ogre/blob/493578624d6cf5513983b1df49fa945dd9cee26f/Components/Overlay/src/OgreImGuiOverlay.cpp#L146-L147

paroj:
but the `NewFrame()` signature and calling convention changed upstream

only-a-ptr (approved):

PR #2798, opened 2021-09-22
Disabled GL plugin on windows

tritonas00:
A lot of people with AMD have to comment GL plugin on windows, or else OGRE hangs when trying to load it.

This is probably an outdated driver issue, but for now Petr and i agreed to comment it by default on windows, since DirectX is far better choice there.


AnotherFoxGuy (approved):
This issue was fixed in a recent driver update, but this is also fixes the issue

Issue #193, opened 2015-05-19
Incorrect command button assignments in "T"-screen


gbakeman:
> Commands have extension and reduction buttons assigned which can be defined by the mod creator. Those buttons are showed by pressing "T" ingame. This screen is only capable to display direct pairs of "F"-Buttons, e.g. F1 + F2. If a user sets a command to F1 for shortening and F3 for extension, that screen still shows F1 + F2 as it can only display direct pairs.
> **How to reproduce:** Choose a vehicle with commands. Hit "T".
> **Content needed:** Any vehicle that uses commands

_This was moved from the wiki bug tracker. [Original post.](http://www.rigsofrods.com/threads/111967-Improvised-bugtracker-users-we-need-your-help-to-collect-all-bugs!?p=1307247&viewfull=1#post1307247)_

Hiradur:
Is this still valid for 0.4.5.0?

ulteq:
Yes, and even worse. The list is incomplete for many trucks.

AnimosusTheWolf:
![image](https://user-images.githubusercontent.com/10286567/79935963-6f9c9800-8413-11ea-9ddf-68aa8fb98f98.png)
![image](https://user-images.githubusercontent.com/10286567/79935968-73c8b580-8413-11ea-87b5-82f63e665fe4.png)

This is still an issue. It should be F5/F9 but it instead shows both pairs. Even though F6 and F10 do nothing.

PR #2799, opened 2021-09-24
Fixed incorrect command key in vehicle description UI

tritonas00:
Fixes https://github.com/RigsOfRods/rigs-of-rods/issues/193

Second command key always assumed as the first + 1 https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/gui/panels/GUI_VehicleDescription.cpp#L86, but actually we can have commands in other keys like F5/F12 for example.

This properly (i think :laughing:) fills empty command descriptions and pairs keys by description in vehicle description panel.

Example:
```
61, 113, 0.3, 0.5, 1.0, 4, 1, 2, f
62, 112, 0.3, 0.5, 1.0, 4, 1, 2, f Boom

136,116, 0.4,0.4,1.0, 10, 5,12, n
136,117, 0.4,0.4,1.0, 10, 5,12, n
136,118, 0.4,0.4,1.0, 10, 5,12, n
136,119, 0.4,0.4,1.0, 10, 5,12, n 

116,124,0.1,0.1,1.0, 2.6,   3, 11, n
117,125,0.1,0.1,1.0, 2.6,   3, 11, n Underlift

137, 64, 0.2,0.2,1.0, 3, 7,8, p
138, 70, 0.2,0.2,1.0, 3, 7,8, p Rear_Feet

140, 96, 0.2,0.2,1.0, 3, 9,10, p Left_Feet
```
kk

### DIFF
--- a/source/main/gui/panels/GUI_VehicleDescription.cpp
+++ b/source/main/gui/panels/GUI_VehicleDescription.cpp
@@ -83,8 +83,16 @@ void VehicleDescription::Draw()
 
             int eventID = RoR::InputEngine::resolveEventName(fmt::format("COMMANDS_{:02d}", i));
             Ogre::String keya = RoR::App::GetInputEngine()->getEventCommand(eventID);
-            eventID = RoR::InputEngine::resolveEventName(fmt::format("COMMANDS_{:02d}", i + 1));
-            Ogre::String keyb = RoR::App::GetInputEngine()->getEventCommand(eventID);
+            Ogre::String keyb;
+
+            for (int x = i+1; x < MAX_COMMANDS; x++)
+            {
+                if (actor->ar_command_key[x].description == actor->ar_command_key[i].description)
+                {
+                    eventID = RoR::InputEngine::resolveEventName(fmt::format("COMMANDS_{:02d}", x));
+                    keyb = RoR::App::GetInputEngine()->getEventCommand(eventID);
+                }
+            }
 
             // cut off expl
             if (keya.size() > 6 && keya.substr(0, 5) == "EXPL+")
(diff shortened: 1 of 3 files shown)


source/main/gui/panels/GUI_VehicleDescription.cpp
```diff
@@ -83,21 +83,36 @@ void VehicleDescription::Draw()
...
             int eventID = RoR::InputEngine::resolveEventName(fmt::format("COMMANDS_{:02d}", i));
             Ogre::String keya = RoR::App::GetInputEngine()->getEventCommand(eventID);
-            eventID = RoR::InputEngine::resolveEventName(fmt::format("COMMANDS_{:02d}", i + 1));
-            Ogre::String keyb = RoR::App::GetInputEngine()->getEventCommand(eventID);
+            Ogre::String keyb;
+
+            for (int x = 1; x < MAX_COMMANDS; x++)
+            {
+                if (!actor->ar_command_key[i].description.empty() && !actor->ar_command_key[i+x].description.empty() &&
+                     actor->ar_command_key[i+x].description == actor->ar_command_key[i].description)
```
only-a-ptr, line 91: `i+x` goes outside the array (both i and x range from 1 to MAX_COMMANDS - 1) - you want to use just `x`.

only-a-ptr (changes requested):
Thanks for finding and fixing this old usability issue.
Pairing the contract/extend keys by description works but only if description is not empty. I've seen a lot of trucks with empty descriptions. As a quick fix I suggest to fix empty descriptions in the parser, like this:
```
description = fmt::format(_L("~Unnamed command {}~"), commandCounter++);
```
Also, when drawing the GUI we can't tell which commandkey is for contraction and which for extension, so we don't know in what order to display them. As quick fix I suggest adding `bool is_extension` to `struct command_t` and filling it in spawner.

tritonas00:
Ok pairing by description is unreliable (will fail if descriptions are the same by the modder etc...)

We need to fetch contract/extend keys directly from the parser.

PR #2806, opened 2021-10-10
Gui open/close events

ohlidalp:
This centralizes how closing a GUI panel interacts with the game - instead of opening other GUI windows directly, a message is posted to the message queue and processed in main loop.

This also fixes problem in current master where in MP, closing Controls UI with ESC key also closes main menu.


only-a-ptr:
Converting back to draft because I found additional edge cases. When you load a savegame from main menu using hotkeys while Settings/About windows are visible, they will appear again when you return to menu. If Controls window is visible, it will remain visible in simulation.

PR #2815, opened 2021-10-28
Revived legacy icons

tritonas00:
Now if icon is given in `putMessage()` it will show up, if not, per area icons apply as before.

There are a lot, some of them:
screenshot_2021-10-28_20-46-14_1
screenshot_2021-10-28_20-46-29_1
kk
kk

Also cleaned up `putMessage()`, except https://github.com/RigsOfRods/rigs-of-rods/blob/d4ace1fb9787359c46697fc39980a12b7703b681/source/main/scripting/ScriptEngine.cpp#L269 to prevent breakage of https://github.com/tritonas00/RoRServerScripts


only-a-ptr (approved):
Excellent work.

PR #2816, opened 2021-11-05
Fixes player character stuttering at high FPS

tritonas00:
Fixes a very old bug with rorbot character stuttering at high fps


only-a-ptr (approved):
Fantastic.

PR #2828, opened 2021-12-02
Slightly modify gearbox names

TheDude53:
Just figured it'd be nice to give the gearbox names a bit more consistency. Let me know what you guys think. Apologies in advance if I come off as a grammar nazi.

PR #2839, opened 2021-12-26
New convention for localizing global enums, example: gearbox modes

only-a-ptr:
Extended rework of #2828

@TheDude53 Thanks for inspiring me to do this. If you want to help further, look into the "new convention" commit below and repeat the procedure for each enum in Application.h:
 1. add the `ToLocalizedString(Enum e)` function - thanks to function overloading in C++, the compiler will pick the correct one each time.
 2. Go through the codebase and replace all custom string conversions of that enum with the new function - except of course the helpers in AppConfig, those are for RoR.cfg only.
 3. Localize the combobox in Settings UI.
 4. Commit, one commit for each enum. 

As long as this PR exists, GitHub should allow you to push directly to this branch in my repo.


tritonas00 (approved):
All good.

PR #2840, opened 2022-01-05
Basic support for plugin scripts.

only-a-ptr:
This feature branch gives you the ability to run AngelScript scripts on launch and have them running continuously until the game shutdown, in both menu and simulation (both single and multiplayer).
The game interface is the same which terrain scripts use, with new extension: global object `input` which lets you retrieve configured controls and current input states.

There is an example script 'demo_script.as' in the "resources/scripts" directory (or 'resources/scripts.zip' package when installed). You can run it by 2 ways:
 * from command line: `ror -runscript demo_script.as` (you can load multiple scripts by using `-runscript` repeatedly)
 * from RoR.cfg: `app_custom_scripts=demo_script.as` (you can enter multiple filenames delimited by comma).

Showcase of the example script:
image
image
image

This is probably the least-nonsense feature I coded so far.

### DIFF
--- a/doc/angelscript/InputEngineClass.h
+++ b/doc/angelscript/InputEngineClass.h
@@ -1,0 +1,31 @@
+
+
+/**
+ * @brief Manages input devices, their configuration (input.map ...) and state.
+ * @note This object is created automatically as global variable `inputs`.
+ */
+class InputEngineClass
+{
+public:
+    /**
+     * @return full configuration string for given command, including the EXPL modifier.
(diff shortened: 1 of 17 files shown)


tritonas00 (approved):
Very nice! You may want to add the new -runscript option in -help also

only-a-ptr:
> @tritonas00 Very nice! You may want to add the new -runscript option in -help also

Good idea, done.

PR #2844, opened 2022-01-23
:video_game: Multiplayer: optional hash check of remote trucks.

only-a-ptr:
RoRnet version bumped to .44 - a STREAM_REGISTER message for a vehicle now contains SHA1 hash - always sent regardless of user setting.

The local check is controlled by cvar `mp_truck_hash_check` - if `true` and the hashes don't match, RoR behaves similarly as if the mod wasn't installed, reporting "Mod hash mismatch: FILENAME" to console and showing stream as failed in MP client list.

The check is enabled by default because I think that's what casual player would expect.

Note that the hashes aren't checked against the online repo, only against the locally installed mods.

### DIFF
--- a/source/main/Application.cpp
+++ b/source/main/Application.cpp
@@ -118,6 +118,7 @@ CVar* mp_server_password;
 CVar* mp_player_name;
 CVar* mp_player_token;
 CVar* mp_api_url;
+CVar* mp_truck_hash_check;
 
 // Diagnostic
 CVar* diag_auto_spawner_report;
--- a/source/main/Application.h
+++ b/source/main/Application.h
@@ -294,6 +294,7 @@ extern CVar* mp_server_password;
 extern CVar* mp_player_name;
 extern CVar* mp_player_token;
 extern CVar* mp_api_url;
+extern CVar* mp_truck_hash_check;
 
 // Diagnostic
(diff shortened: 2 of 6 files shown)


CuriousMike56:
I know this PR is old, but I don't think this is a good idea to have enabled by default.
People mod their vehicles for MP use all the time, especially for racing. This would effectively kill the ability to use your modded vehicle in MP, which would take the fun out of MP for a lot of people. Also if a newer version of a mod gets released with only a small change (e.g. adding a GUID) then everyone else would have to re-download that mod, even if the change doesn't affect the ability for others to see you. 
The only useful case for this would be competitive events (what Sim-Monsters does for example) to ensure that people don't cheat.

only-a-ptr:
@CuriousMike56 Makes sense. I'm re-marking this a draft before we figure this out. I agree that blocking mismatching mods is too strong a setting and should be disabled by default, nonetheless the player may want to know the mod he's seeing isn't what the remote player is using - maybe a ⚠️ icon in the MP client list, with a tooltip saying "Checksum failed - the remote mod is different than yours".

only-a-ptr:
Closing this because of general lack of interest - once this topic comes up again, I'll revisit it.

PR #2846, opened 2022-01-24
Node/beam diagnostic options roundup

ohlidalp:
https://forum.rigsofrods.org/threads/node-beam-diagnostic-options-roundup.3517/


ohlidalp:
Closing as stale - in the meantime a new T-panel was created so these are out of date.

Issue #2450, opened 2019-09-25
Content installation - user experience.


only-a-ptr:
Currently RoR mods are in plain .ZIP archives and installing them means to manually put them to `MyDocuments\MyGames\Rigs of Rods\mods` - ideally without unzipping.

I got inspired to improve the process; when you open any [Rainmeter Widget page on DeviantArt](https://www.deviantart.com/rainmeter) and click download, you'll see:
![image](https://user-images.githubusercontent.com/491088/65637902-4f276f00-dfe5-11e9-89f1-48cdb0641727.png)

Further:
![image](https://user-images.githubusercontent.com/491088/65638535-9eba6a80-dfe6-11e9-8b83-60e2a890f30e.png)

The .rmskin is an ordinary .ZIP (deflate-mode), just renamed.

I'd like to install RoR mods like this, too: decide on a file extension like ".ror-pak" or ".rrpak" and add necessary bindings for the desktop environments.

END public note to self.

disloyalpick:
The file extension could also be name like `.mod` or something like that?

only-a-ptr:
@disloyalpick Nope, taken: https://en.wikipedia.org/wiki/MOD_(file_format)

tritonas00:
i would like to see an in game "store"

an imgui panel that shows all content fron the official repo with thumbnails, with two buttons:
install -> download the selected mod zip into mods dir and update cache
remove ->  delete the zip and update cache

in my dreams lol

CuriousMike56:
@tritonas00 Exactly what I think. The XenForo API allows for this.

Issue #229, opened 2015-06-03
Feature Request: Repository Downloads


ghost:
A neat new feature for the next official release of RoR would be a built in repository downloader. So, the RoRConfig.exe checks for vehicles and terrains in the repository that you don't have and downloads them after it prompts you and you click Yes.

only-a-ptr:
This would be nice to have, indeed.

Unfortunately, this would require the repository web application, and only TDev has sources and access to do that. Further, this would cause much more frequent downloads which means more server bandwidth and possibly (I don't know if it matters) more money for TDev to pay for hosting.

A simple + feasible improvement would be to add [download content] button to the selector, which would open web repository in player's default browser.

vido89:
Why it needs to be on ror main site, and why we do not use more then one mirror/place for stuff like this

only-a-ptr:
@vido89 : It doesn't, really, we may just as well set up an alternate repository, if we find a hosting with enough space and bandwidth. To be honest, I don't really like the current reposotory - it has bloated UI with rather amateur design and the search box doesn't really search.

@vido89 :However, new repository would also need new content. I'm not sure authors would agree with us to move their stuff, and a lot of stuff is abandoned, so there's nobody to move it. 

This requires wider discussion in the community. But I like the idea

gigapouch:
It's OK if it won't work out, it was just a suggestion. ;)

Hiradur:
@vido89 
The repository is mirrored according to http://www.rigsofrods.com/CdnNodes/
Also see how much traffic it generates: http://www.rigsofrods.com/repository/stats

vido89:
@Hiradur yup some nice stats there, so there is need to mirror it on more places

Dogy144:
Just a suggestion, but maybe integrate with steam? (Steam Workshop?)

gigapouch:
No, not Steam! I'm not quite 13 yet! (I will be in September, though.)

gigapouch:
Not Steam, because then the RoR fans that are under 13 (like me, I'm 12 1/2) won't get to download any mods!

DirtGamer301:
By what I know, it's not that easy to release a game via Steam and it does cost money.

Also, I'm always happy to have something which I don't need to be logged in all time to play and which I don't need an account to get (I have a Steam Account, but it's nicer to have RoR as its own).
[...]

PR #2850, opened 2022-02-08
Repository Manager

tritonas00:
An in game repository manager, thanks to the new upcoming API provided by @Zentro and the help of @only-a-ptr. Three view modes are available, categories, searching, sorting and the basic functionality of installing/updating and removing mods, in the `mods` folder.

kk1
kk2
kk3
kk4

### Limitations

Currently the panel will detect/(re)install/update/remove mods **only** in the `mods` directory. We may refine in the future.

Closes https://github.com/RigsOfRods/rigs-of-rods/issues/2450
Closes https://github.com/RigsOfRods/rigs-of-rods/issues/229


cryham:
Wow looks great.
I wanted to ask something. Sorry if off topic, but:
the only thing that I miss in here (and on forum browser) is a license info like e.g. GPL, CC-BY-SA, etc.

I guess it is safe to assume that except for just 3 things (that are GPLed in https://github.com/RigsOfRods/content repo), all else has no license what-so-ever right? Meaning anybody can only download it and use in RoR right?
And likely nothing will happen change here, because that would require contacting authors (likely some left years ago) and ask them to specify and release their content with a license, like CC or so. Also BTW some cars have textures with logos etc, which is not legal to use anyway.

tritonas00:
> Wow looks great. I wanted to ask something. Sorry if off topic, but: the only thing that I miss in here (and on forum/resources browser) is a license info like e.g. GPL, CC-BY-SA, etc.
> 
> I guess it is safe to assume that except for just 3 things (that are GPLed in https://github.com/RigsOfRods/content repo), all else has no license what-so-ever right? Meaning anybody can only download it and use in RoR right? And likely nothing will change here, because that would require contacting authors (likely some left years ago) and asking them to specify and release their content with a license, like CC or so, and them agreeing. Also BTW some cars have textures with logos etc, which is not legal to use anyway or probably textures made from non-CC stuff etc which makes CC licensing impossible too.

We will add this info in the panel

only-a-ptr (approved):
I fixed a few details I noticed and I think it's perfect now.

tritonas00:
> improved download speed/rem. time calc

welp now its off for me, both speed and remaining time
kk

my max download speed in 2MB/sec, `speed/10` shows it correctly, i guess it depends on the connection

maybe we should just remove them and leave only downloaded MB of full filesize MB

only-a-ptr:
> welp now its off for me, both speed and remaining time

Thanks for the CURL link, I took another shot and I'm pretty confident it works right now. In both cases I coded & tested using jumpy wi-fi on train :)

> maybe we should just remove speed and leave only downloaded MB of full filesize MB + a better remaining time and call it a day

Agreed. If you're still not happy with it, just erase my commit or do whatever edit you like, then merge this.

tritonas00:
Yep, still problems with. Lets go KISS

kk

bippy89:
wheres the download

only-a-ptr:
> @bippy89 wheres the download

https://www.rigsofrods.org/download

PR #2852, opened 2022-02-12
Bigger shortcut buttons in controls panel

tritonas00:
Makes it more consistent and easier to catch

kk


only-a-ptr (approved):
I like it.
Also, the header/body separator line is a nice touch.

PR #2855, opened 2022-02-22
Hide mouse when loading terrain

tritonas00:
Since it can't be moved at that stage.


only-a-ptr (approved):
Makes complete sense, thanks.

Issue #330, opened 2015-07-24
In cabin view sky is black in the mirrors


Speciesx:
This happens only with Caelum sky. Load a vehicle and change to cabin view
Tested with Tatra-T813 8x8 Dakar and Jantor BigMachines 1.0
cabin view active:
![ror_2015_07_24_14_23_57_258](https://cloud.githubusercontent.com/assets/1196375/8874366/8dc8605c-3210-11e5-959f-fb8554480135.png)
![ror_2015_07_24_14_25_05_326](https://cloud.githubusercontent.com/assets/1196375/8874367/8dc9f12e-3210-11e5-8a1d-1387eb367eb4.png)
In other view sky is visible.
![ror_2015_07_24_14_24_56_174](https://cloud.githubusercontent.com/assets/1196375/8874368/8dca1294-3210-11e5-8af4-d4f3b0174dd0.png)

Max98:
This looks like some badly made optimizations. Because on the other mirror, you have the sky (tatra)
We will look into this.

Hotrod55:
Adding to this issue, on some vehicles it'll show a really low-quality "zoomed in" image on the other mirror.
![screenshot_23](https://cloud.githubusercontent.com/assets/11002490/9336255/a04f926e-45a6-11e5-9e93-94c355ce6391.png)

only-a-ptr:
Removing from v0.4.8 because we have SkyX now and Caelum might be removed soon. Also, this would better be resolved after port to OGRE 1.10+

PR #2862, opened 2022-03-06
Fixed caelum sky rendered black on mirrors

tritonas00:
Fixes https://github.com/RigsOfRods/rigs-of-rods/issues/330

Tested with agora though, since i can't reproduce the tatra bug


only-a-ptr (approved):
Perfect

Issue #2860, opened 2022-03-03
RoR crashes when pressing Ctrl+Del while holding Backspace


TheDude53:
### Steps to reproduce
1. Open RoR
2. Pick a terrain and spawn something
3. While in whatever you spawned, hold Backspace and press Control+Delete.

### Expected behavior
The car would be removed and the Backspace keystroke would be disregarded.

### Actual behavior
RoR briefly freezes before crashing

### System configuration
OS: Windows 10 64-bit
Renderer used: DirectX 9

### Additional information, logs and screenshots
Reproducible with the current master branch  
RoR.log does not contain any error messages

tritonas00:
Will check

WillMichals:
I am able to reproduce with same system.

PR #2866, opened 2022-03-11
Fixed crash when deleting actor while holding backspace

tritonas00:
Fixes https://github.com/RigsOfRods/rigs-of-rods/issues/2860

It happened only on `EV_COMMON_REMOVE_CURRENT_TRUCK`, removing from top menu bar works fine.

### DIFF
--- a/source/main/main.cpp
+++ b/source/main/main.cpp
@@ -849,6 +849,8 @@ int main(int argc, char *argv[])
                         }
                         App::GetCameraManager()->UpdateInputEvents(dt);
                         App::GetOverlayWrapper()->update(dt);
+                        App::GetGameContext()->GetRecoveryMode().UpdateInputEvents(dt);
+                        App::GetGameContext()->GetActorManager()->UpdateInputEvents(dt);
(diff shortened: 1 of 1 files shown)


only-a-ptr (approved):
I think I know why this helps...
The 'backspace mode' does check if the player actor was removed: https://github.com/RigsOfRods/rigs-of-rods/blob/6897c1fc52b5fea98c6cc24941e7b8c210ba6daa/source/main/gameplay/RecoveryMode.cpp#L39-L44
But if that code runs while DELETE message is already queued, then it's bust - it will queue the MODIFY message with a pointer to vehicle which will already be DELETE-d at the time of processing.
Well, I'm going to approve this for now, but it must eventually be redone like this:
```
typedef std::shared_ptr<Actor> ActorPtr;

enum ActorState { DELETED, /*existing ones*/ }

ActorManager:
    std::vector<ActorPtr> m_actors;

on MODIFY:
    // check if not already DELETED

on DELETE:
    // remove from ActorManager and set state to DELETED

```
That will also solve a potential issue with scripting - those may hold pointers to actors as well.

PR #2870, opened 2022-03-16
Don't crash on invalid icon name

tritonas00:
Icon names can be defined from scripts etc.... I corrected some invalid names some time ago, but better be sure.

Also i'm not sure if `Ogre::TextureManager::getSingleton().load` every frame is good practice...


only-a-ptr (approved):
Fine

only-a-ptr:
> Also i'm not sure if Ogre::TextureManager::getSingleton().load every frame is good practice...

It is as long as you stop when it fails. Which we don't here because the message comes in as `const` which is good practice, I'd have to add a pointer to the `Message` struct and extract the lookup to `ResolveIcon(Message& m)` or something... CBA.

PR #2871, opened 2022-03-17
👷 Updated windows CI image to windows-2019

AnotherFoxGuy:
Upgraded to windows-2019 since windows-2016 is no longer supported
image

### DIFF
--- a/.github/workflows/build-game.yml
+++ b/.github/workflows/build-game.yml
@@ -73,7 +73,7 @@ jobs:
 
   build-msvc:
     name: Windows build
-    runs-on: windows-2016
+    runs-on: windows-2019
     env:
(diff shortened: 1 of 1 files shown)


tritonas00:
ouch, that will bring the rorbot colors glitch again

AnotherFoxGuy:
> ouch, that will bring the rorbot colors glitch again

Yes, this will reintroduce the pssm shadow glitch but there isn't really an alternative

tritonas00:
> > ouch, that will bring the rorbot colors glitch again
> 
> Yes, this will reintroduce the pssm shadow glitch but there isn't really an alternative

We will release before 1st April, then we can fix it later

PR #2874, opened 2022-03-18
Paged: check if farther details level is greater than closer

tritonas00:
Lately i'm getting

```

Technical details below: 

RenderingAPIException: Closer detail levels must be added before farther ones in PagedGeometry::addDetailLevel() at /home/babis/Downloads/ror-dependencies/Source/paged-geometry/source/PagedGeometry.cpp (line 365)
```

when loading brutal valley. This fixes it

min is 40 and max 41 which makes the error valid

vegetation density must be set to 20% to reproduce.


only-a-ptr (approved):
I could not reproduce on Win10 / GeForceGtx660 with either OpenGL or Dx9, but it looks valid.

PR #2877, opened 2022-03-20
:bug: Fixed CMake option for specifying a custom version

AnotherFoxGuy:
This feature got added in #1387, but it got broken when we switched from semver to rolling releases

Usage: Set `ROR_CUSTOM_VERSION` to `[year].[month]-[prefix]`


tritonas00:
Works
kk

PR #2879, opened 2022-03-21
Set depth_bias only on character/vehicles

tritonas00:
Setting it globally on `shadows.material` kills survey map (with PSSM on):
kk


only-a-ptr (approved):
Works perfectly.

PR #2884, opened 2022-03-28
Fixed flickering shadows

tritonas00:
Example terrain: https://sim-monsters.com/files/file/3734-orlando-2021/

Before: https://streamable.com/jc4591
After: https://streamable.com/nos1a2
(I don't know why it recorded on fast forward :laughing:)

Also fixes sloppy rendered survey map with shadows:
Before:
screenshot_2022-03-28_11-54-25_1
After:
screenshot_2022-03-28_11-54-41_1

### DIFF
--- a/resources/managed_materials/shadows/pssm/on/shadows.material
+++ b/resources/managed_materials/shadows/pssm/on/shadows.material
@@ -4,6 +4,7 @@ abstract technique Shadows/managed/base_receiver
 	{
 			ambient 1 1 1 1
 			diffuse 1 1 1 1
+			depth_bias 20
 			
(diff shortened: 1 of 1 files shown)


tritonas00:
@paroj i go a bit blind here, is this the right way to handle depth fighting issues?

paroj:
yes, but it also comes with its own problems:
https://docs.microsoft.com/en-us/windows/win32/dxtecharts/common-techniques-to-improve-shadow-depth-maps#peter-panning

tritonas00:
It creates this issue https://discord.com/channels/136544456244461568/189904947649708032/958169316526796830

PR #2885, opened 2022-03-28
Fixed selector switching to "All" category when searching

tritonas00:
When a certain category is selected, search box searches on that category, not All.


only-a-ptr (approved):
Works as expected

PR #2888, opened 2022-03-31
:wrench: Made content folder building optional

AnotherFoxGuy:


tritonas00:
Eh why? :laughing:

AnotherFoxGuy:
> Eh why?

Because I use clion, and every time you switch configuration RoR rebuilds the cache
Also for every build you have installed the content folder gets duplicated, so I ended up with 6 simple terrains in the terrain selector

AnotherFoxGuy:
> Still i would suggest ON by default, and you can use OFF when you have multiple configurations, we are the minority having multiple configurations

Yea, that is true
I've updated the PR to have it on by default

tritonas00:
Works fine now :+1:

PR #2892, opened 2022-04-06
Network quality indicator fixes

tritonas00:
- Fixes visibility
Before:
kk
After:
kkk

- Fixes not resetting on disconnect, which made it appear always when connecting to another server


only-a-ptr (approved):
Works good, thanks.

PR #2895, opened 2022-04-08
Replaced old spinner

tritonas00:
Replaced old spinner with the one from the repo panel

kk

Also replaced "refreshing" message with spinner in mp selector

kk


only-a-ptr (approved):
Perfect

PR #2899, opened 2022-04-30
Cleaned up analog dashboard

tritonas00:
This is **Mark3**'s work from discord. It removes the "leds" section in the analog dashboard which was left from previous versions ([image](https://cdn.discordapp.com/attachments/969560446518562836/969577247335907348/unknown.png)). It interfered with notifications area and made the space cluttered.

Changes:
- moved the dashboard indicator lights onto the gauges
- moved "mph" and "kph" on gauge backgrounds
- fixed the RPM needle angle on 8k tacho
- fixed the right turn signal indicator jumping position by 2 pixels between on and off 
- re-positioned the boats gauge so it's in the same position as speedo gauge on other vehicles, for consistency

Before:
before1
before2
After:
after1
after2


only-a-ptr (approved):
I like it. You may as well delete the now-unused graphics.

tritonas00:
> I like it. You may as well delete the now-unused graphics.

Indeed! Done.

PR #2905, opened 2022-05-21
Use nicemetal materials

tritonas00:
Always use classic (nicemetal) materials. They look better (cg shaders), they receive light and they don't glitch with PSSM, because they don't receive shadows, only cast.

https://discord.com/channels/136544456244461568/274653814467067916/977568484210917476
https://discord.com/channels/136544456244461568/274653814467067916/977608390874390539
https://discord.com/channels/136544456244461568/274653814467067916/977593552186011719


only-a-ptr (approved):
Works as expected.

PR #2908, opened 2022-05-30
:wrench: Replaced discord_rpc with DiscordGameSDK

AnotherFoxGuy:
Also added support for joining MP servers through discord, needs testing


only-a-ptr:
How is the 'startror:' protocol handler going to be done?

erkinalp:
alternative: replace discord game sdk with mirandaNG discord interop

PR #2910, opened 2022-06-06
Fixed gcc warning

tritonas00:
Lately gcc complains:

```
[ 41%] Building CXX object source/main/CMakeFiles/RoR.dir/physics/flex/FlexMeshWheel.cpp.o
/home/babis/Downloads/ror-dependencies/rigs-of-rods/source/main/physics/flex/FlexMeshWheel.cpp: In constructor ‘RoR::FlexMeshWheel::FlexMeshWheel(Ogre::Entity*, RoR::GfxActor*, int, int, int, int, const std::string&, const std::string&, float, bool)’:
/home/babis/Downloads/ror-dependencies/rigs-of-rods/source/main/physics/flex/FlexMeshWheel.cpp:86:38: warning: argument 1 value ‘1844674407370955155 ’ exceeds maximum object size 9223372036854775807 [-Walloc-size-larger-than=]
   86 |     m_indices=(unsigned short*)malloc(m_index_count*sizeof(unsigned short));
      |                                ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/include/c++/12.1.0/cstdlib:75,
                 from /usr/include/c++/12.1.0/ext/string_conversions.h:41,
                 from /usr/include/c++/12.1.0/bits/basic_string.h:3960,
                 from /usr/include/c++/12.1.0/string:53,
                 from /home/babis/Downloads/ror-dependencies/rigs-of-rods/source/main/phc.h:25,
                 from /home/babis/Downloads/ror-dependencies/rigs-of-rods/source/main/CMakeFiles/RoR.dir/cmake_pch.hxx:5,
                 from <command-line>:
/usr/include/stdlib.h:540:14: note: in a call to allocation function ‘void* malloc(size_t)’ declared here
  540 | extern void *malloc (size_t __size) __THROW __attribute_malloc__
      |              ^~~~~~
```

I don't know if this is the correct way to fix it but it suppress the warning.

### DIFF
--- a/source/main/physics/flex/FlexMeshWheel.cpp
+++ b/source/main/physics/flex/FlexMeshWheel.cpp
@@ -66,7 +66,7 @@ FlexMeshWheel::FlexMeshWheel(
 
     // Define the vertices
     m_vertex_count = 6*(nrays+1);
-    m_vertices = new FlexMeshWheelVertex[m_vertex_count];
+    m_vertices.resize(m_vertex_count);
 
     int i;
     //textures coordinates
@@ -83,7 +83,7 @@ FlexMeshWheel::FlexMeshWheel(
     // Define triangles
     // The values in this table refer to vertices in the above table
     m_index_count = 3*10*nrays;
-    m_indices=(unsigned short*)malloc(m_index_count*sizeof(unsigned short));
+    m_indices.resize(m_index_count);
     for (i=0; i<nrays; i++)
     {
         m_indices[3*(i*10  )]=i*6;   m_indices[3*(i*10  )+1]=i*6+1;     m_indices[3*(i*10  )+2]=(i+1)*6;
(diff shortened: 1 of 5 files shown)


only-a-ptr (changes requested):
Nope, not correct. Your're basically clamping the flexbodywheel mesh to maximum of 65536 bytes of index buffer, which means circa 10k triangles max. That's actually a reasonable limit and the hardware prefers that, but the problem in the crash report is that somehow we tried to allocate some 30 trillion of triangles, which is nonsense, so there must be a bug earlier up the pipe.

I'd do this:
```
if (m_index_count*sizeof(unsigned short) > std::numeric_limits<unsigned short>::max())
{
  // log error and bail.
}
```

ohlidalp:
I'm sure this isn't true anymore.

tritonas00:
> I'm sure this isn't true anymore.

The warning still appears, on Lunix, gcc

ohlidalp:
OK, let me try to stuff `vector<>` in there...

ohlidalp:
@tritonas00 Try again please. GitHub won't let me request your review since it's your PR 😆

tritonas00:
No warnings :+1:

Issue #2911, opened 2022-06-09
Hide remote vehicle glitch


tritonas00:
Hiding remote vehicle glitches on the remote side if the user who hides has the same vehicle spawned for the first time. If he spawns the same vehicle, remove it, respawn it again and hide the remote one, no glitch.

https://user-images.githubusercontent.com/2660424/172816335-37c2687a-37a7-46c9-a73c-7a593e474e74.mp4

I think something must be wrong here https://github.com/RigsOfRods/rigs-of-rods/blob/master/source/main/gui/panels/GUI_TopMenubar.cpp#L742-L762 since it glitches only the first time and not the rest

Or we forgot a `ActorState::NETWORKED_HIDDEN` check somewhere.

I don't understand why it happens only with the same vehicle and only the first time though, makes no sense  :neutral_face:

PR #2912, opened 2022-06-09
Fixed glitch when hidding remote vehicle

tritonas00:
Fixes https://github.com/RigsOfRods/rigs-of-rods/issues/2911


only-a-ptr (approved):
The glitch is fascinating. While testing on Debug I noticed it's not only visual, it's physics - the glitcing wehicle bumped into my local one on a few occasions.

This commit helps but it's only a workaround, the incoming data for hidden actor should be just ignored without any disturbance on the sending side. I added a comment so the code makes better sense.

tritonas00:
> the glitcing wehicle bumped into my local one on a few occasions.

yeah, i had to disable mp collisions for the vid

PR #2913, opened 2022-06-12
Improved animations

vido89:
Have fun guys


only-a-ptr:
I took the liberty to convert the PR to draft - since both builds fail it's apparently not ready for review and merging.

PR #2924, opened 2022-07-23
Don't popup menubar/buttons if cursor is in another window

only-a-ptr:
Quality of life improvement - only pop up the top menubar / buttons UI if the cursor isn't over another window.

isHovered


tritonas00:
Load a terrain and go back to menu, no wallpaper:

kk

tritonas00 (approved):

PR #2925, opened 2022-07-23
Created flexbody debug view + GUI

### LINKED ISSUE #2923: Flexbody binds to nodes outside 'forset'.
Discord complaint by DannyWerewolf.

If a flexbody is bound to a detachable part and it detaches, the mesh keeps stretching after the vehicle, even though all nodes in the 'forset' have detached.
obrazek
obrazek
obrazek

it's probably because the flexbody binds 3 nodes for each vertex: reference, X and Y, and only the reference nodes are taken from the forset, while the X/Y nodes are searched by distance.

Danny Werewolf has a feeling that it's not a specific node but the center of the truck.

only-a-ptr:
This tool lets you inspect individual vertices and their generated bindings. Just hover mouse pointer over a vertex and see. Click the vertex to toggle persistent visibility.
Closes #2923.
obrazek
Vehicle on display: https://forum.rigsofrods.org/resources/honda-accord.281/

### DIFF
--- a/source/main/Application.cpp
+++ b/source/main/Application.cpp
@@ -237,6 +237,14 @@ CVar* gfx_flexbody_cache;
 CVar* gfx_reduce_shadows;
 CVar* gfx_enable_rtshaders;
 
+// Flexbodies
+CVar* flexbody_defrag_enabled;
+CVar* flexbody_defrag_const_penalty;
+CVar* flexbody_defrag_prog_up_penalty;
+CVar* flexbody_defrag_prog_down_penalty;
+CVar* flexbody_defrag_reorder_indices;
+CVar* flexbody_defrag_invert_lookup;
+
 // Instance management
 void SetCacheSystem    (CacheSystem* obj)             { g_cache_system = obj; }
 
--- a/source/main/Application.cpp
+++ b/source/main/Application.cpp
@@ -243,6 +243,7 @@ CVar* flexbody_defrag_const_penalty;
 CVar* flexbody_defrag_prog_up_penalty;
 CVar* flexbody_defrag_prog_down_penalty;
 CVar* flexbody_defrag_reorder_indices;
+CVar* flexbody_defrag_reorder_texcoords;
 CVar* flexbody_defrag_invert_lookup;
 
 // Instance management
--- a/source/main/Application.h
+++ b/source/main/Application.h
@@ -436,6 +436,7 @@ extern CVar* flexbody_defrag_const_penalty;
 extern CVar* flexbody_defrag_prog_up_penalty;
 extern CVar* flexbody_defrag_prog_down_penalty;
 extern CVar* flexbody_defrag_reorder_indices;
+extern CVar* flexbody_defrag_reorder_texcoords;
 extern CVar* flexbody_defrag_invert_lookup;
 
 // ------------------------------------------------------------------------------------------------
--- a/source/main/Application.h
+++ b/source/main/Application.h
@@ -430,6 +430,14 @@ extern CVar* gfx_flexbody_cache;
 extern CVar* gfx_reduce_shadows;
 extern CVar* gfx_enable_rtshaders;
 
+// Flexbodies
+extern CVar* flexbody_defrag_enabled;
+extern CVar* flexbody_defrag_const_penalty;
+extern CVar* flexbody_defrag_prog_up_penalty;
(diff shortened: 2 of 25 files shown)


only-a-ptr:
While looking at the debug view I created, I figured an optimization. Most often, many vertices bind to the same NREF/NX/NY nodes, so for optimal CPU cache usage, they should be processed together. However, currently they are processed in the order of vertices in the mesh, which is random, depending on the modelling tool and exporter. I wanted to actually see it, so I created a visualization:
obrazek
Next step - reorder vertices in the mesh (it's possible on runtime, I've already done it once) to match the nodes. The result will be visible in this graph.

only-a-ptr:
I implemented the "defragmentation" of flexbody meshes. Figuring out the algorithm was a challenge, coding it was extremely prone to mixing things together.

In the image below, the vertex locators are reordered but texcoords aren't (checkbox is off), resulting in an artistic effect.
obrazek
Vehicle on display: Ford Tudor https://forum.rigsofrods.org/resources/1949-ford.283/

Zentro:
No crashes or errors to report.

A minor thing, but when inspecting vertices and picking some with your mouse you can immediately deselect them. However, when unchecking the "Vertices" box after selecting a few vertices, the vertices can no longer be unselected (after rechecking the "Vertices" box) without browsing through the vertex locator table and unselecting them there. 
image

The defragmentation tool works without issue with the defaults, unselecting any or all of the 3 optimizations results in some interesting effects as you've noted in your previous reply. Not sure if this is intended or not.
image

only-a-ptr:
**Purpose**: To visually explain how flexbodies work and help modders diagnose them.

**Estimate**: 1 man day. It all works.

CuriousMike56:
Is the "hide others" option meant to show only the selected flexbody? if so it doesn't work:
image
Upon testing to see if it works with `nodes2`, I found an extreme use case: the high poly count of the [1971 DI Banger Brambler Package](https://www.dropbox.com/sh/x0lsx1ezlci8wx3/AADZsqt7KrAstf3sedLzcSQra/Vehicles/DI_Banger391%2BDI_Brambler391%2BDI_Revair391%2BDI_Tracer391.zip?dl=0) causes the frame rate to plummet when activating the debug views:
image
This most likely can't be "fixed" but could be an easy way to test further optimization. 

Otherwise it seems to work fine.

only-a-ptr:
> @CuriousMike Is the "hide others" option meant to show only the selected flexbody? if so it doesn't work: (Desperado buggy)

It actually works perfectly, but what you're looking at aren't `flexbodies` but `props` :smile: But I admit it's confusing, so I added props to the list, too:
image
Yes, there's a glitch - props outside of `props` (like wheel rims) aren't accounted for, but I'll fix that after this PR.

> @CuriousMike56 the high poly count ... causes the frame rate to plummet when activating the debug views

That's because I'm drawing all those as circles using DearIMGUI (circa 6 triangles each, antialiased). Using raw vertex buffer would be much faster and would look pretty much equivalent in this case. But presently all our debug views work like this, I'll deal with this in the future if it troubles users.

CuriousMike56:
Still can't get the hide others option to work, tried multiple vehicles with plenty of flexbodies 😕 
image

only-a-ptr:
@CuriousMike56 Looks like it doesn't work at all for you... very strange. Try disabling PSSM and setting all visuals to lowest.

CuriousMike56:
@only-a-ptr Turns out realtime reflections breaks it, works fine with it disabled:
image

only-a-ptr:
> @CuriousMike56 Turns out realtime reflections breaks it, works fine with it disabled

Makes sense, all meshes must be hidden when rendering the reflections, otherwise they would obscure the view for the reflection-cameras. I fixed it by pausing reflections while "Hide all" is active, test once more please.

CuriousMike56:
Now enabling "hide other" on the first entry on the list causes all flexbodies to disappear, leaving only the props:
image
image
All other flexbodies appear as normal:
image
Occurs with reflections enabled or disabled.

CuriousMike56:
First flexbody not appearing still persists:
image

only-a-ptr:
> @CuriousMike56  First flexbody not appearing still persists

Fixed.

CuriousMike56 (approved):
👍

Zentro (approved):

PR #2927, opened 2022-07-25
Frontend/Backend split

only-a-ptr:
Explained at the forums: https://forum.rigsofrods.org/threads/checklist-for-porting-to-different-3d-engine-than-ogre.3510/#post-18007
Presently, it's only split to static library (RoR_Main.lib) and executable (RoR.exe), so everything interoperates just like nothing hapenned and the backend does everything including rendering. This will change...

Build with `ROR_USE_CAELUM=OFF`
Set `gfx_shadow_type=No shadows (fastest)`

Status: 
* Sectionconfig is temporarily disabled.
* There are visual glitches (placeholder textures on skybox and particles) due to the OGRE upgrade.
* But otherwise it seems to run as usual (Windows10, Visual Studio 17 2022)


only-a-ptr:
Closing as this was a flop - but it revealed one important thing: our rendering loop can be easily moved off to it's own thread

PR #2928, opened 2022-07-29
Fixed cache-file version check.

only-a-ptr:
The check was only done after attempting to load the file contents. It must be done after JSON parsing but before loading contents.


tritonas00 (approved):
Now it works fine.

PR #2933, opened 2022-08-28
Added hydro flag `j`, restored bug with `i` for backw. compat.

### LINKED ISSUE #2932: Compatibility regression of `hydros` flag `r` (rudder input)
Kudos to DannyWerewolf on Discord for finding and reporting!

**The problem:**
* on current stable release (2022.04) and prior, the `hydros` flag `r` (rudder input) also works for land vehicles and, quote: _On current release, with "r" is exactly how it should be, and about how fast the trucks have always steered with default sensitivity on controller._
* on current git master branch, when `r` flag is present, the hydro doesn't respond to keyboard at all.

**Next steps:**
I'm going to investigate what hapenned and later document this "trick" because I like when players get creative with the software and find edge cases. I'm also keen to maintain backwards compatibility.

**Discuss on the forum:** https://forum.rigsofrods.org/threads/hydro-r-flag.3709/

only-a-ptr:
It turns out the `i` flag has a bug: it always makes the hydro invisible, but if it comes as first/single flag, then it also makes the hydro commanded by steering input. That means:
* `ri` will make the hydro respond only to airplane rudder, and invisible.
* `ir` will make the hydro respond to both airplane rudder and steering input, and invisible. This makes it behave a lot different than plain hydro though, and many SimMonsters mods use it as a feature.

To resolve backward compatibility, I added a new flag `j`, which makes the hydro invisible but without the quirks of `i`. The parser still recognizes `i` but internally I'm using only `j` and `n`.

The documentation https://docs.rigsofrods.org/vehicle-creation/fileformat-truck/#hydros needs to be updated:
* `` no flags given -> force `n` flag.
* `n` means Normal and makes the hydro driven by steering input.
* `j` makes the hydro invisible.
* `i` makes the hydro invisible, but BEWARE: if it comes as first/single flag, then it also forces `n` flag. This quirk is maintained for backwards compatibility.

Fixes #2932

### DIFF
--- a/source/main/physics/ActorSpawner.cpp
+++ b/source/main/physics/ActorSpawner.cpp
@@ -3590,12 +3590,7 @@ void ActorSpawner::ProcessHydro(RigDef::Hydro & def)
     bool invisible = false;
     unsigned int hydro_flags = 0;
 
-    if (def.options == 0)
-    {
-        invisible = false;
-        hydro_flags |= HYDRO_FLAG_DIR;
-    }
-    if (BITMASK_IS_1(def.options, RigDef::Hydro::OPTION_i_INVISIBLE))
+    if (BITMASK_IS_1(def.options, RigDef::Hydro::OPTION_j_INVISIBLE))
     {
         invisible = true;
     }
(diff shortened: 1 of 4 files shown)


po185021:
How to test: pick any truck, for example the bundled DAF semitruck, and put `ir` flags to hydros:
* On master, the hydros will stop working
* On this branch, the hydros will work but will be slower and softer.

tritonas00 (approved):

only-a-ptr:
Test using this vehicle (all hydros have `ir` flags):
[dafsemi-hydro-ir.zip](https://github.com/RigsOfRods/rigs-of-rods/files/9485231/dafsemi-hydro-ir.zip)
* on master, it won't steer at all.
* on this PR, it will steer but differently than plain DAF.

PR #2935, opened 2022-09-02
Fixed character not exiting hidden moving remote actor

tritonas00:
Reported from the forum https://forum.rigsofrods.org/threads/2022-04-discussion-thread.3615/#post-18094


only-a-ptr (approved):
Well done.

PR #2939, opened 2022-09-16
New UI panel: CollisionsDebug (top menubar > tools)

only-a-ptr:
Current "Debug Collisions" overlay (cvar 'diag_debug_collisions') is non-interactive, laggy and with hard-to-read text labels.

The new CollisionsDebugUI allows enabling/disabling the overlays interactively and draws labels using DearIMGUI (like net labels.)
obrazek

**Changes:**
* created the new GUI window, with all visibility/hover/update mechanics (it's getting convoluted).
* ported the eventbox visiualization code from Collisions.cpp to the new GUI window. Boxes remain meshed, labels are DearIMGUI.
* ported collision mesh view. Collision triangles are visualized as orange objects, same as before. Labels are drawn by DearIMGUI. Unlike before, the triangles aren't one mega-mesh covering whole map, but individual meshes. Visibility range can be adjusted in UI.
* ported lookup grid view. Reused old code in Collisions.cpp, just added root node to control visibility and fixed drawing distance.


only-a-ptr:
**Added grid visualization.**

Reused old code in Collisions.cpp, just added root node to control visibility and fixed drawing distance.

This debug view is amusing because it neatly visualizes how inefficient the grid system is. Map is dividied into 2x2meter blocks (works only horizontally! height is unlimited!), each having an unlimited capacity for collision triangles. The shade of green/yellow/red indicates occupancy (green = 0, red=CELL_BLOCKSIZE(146)). Naturally, a much better system would be a search tree where collision triangles are distributed evenly.
obrazek

AnotherFoxGuy:
A few bugs I've noticed while testing this:

- `Show lookup cells` crashes if you switch terrains
- The checkboxes don't reset on terrain switch

only-a-ptr:
obrazek
**Final touches:**

- CollisionsDebug: factored out funcs `SetDrawFoo()`, `ClearFooVisuals()`.
- Collisions.cpp: material gen code moved to CollisionsDebugUI, with duplication check.
- Fixed missing cleanup of cell meshes.
- Added "dump debug meshes" rescue buttons - especially important for cells.
- Fixed cleanup on terrain unload.
- Added auto-hide on UI close.

only-a-ptr:
obrazek
Final version: procedural roads are included, and there are additional display options for labels.

tritonas00:
Works fine and looks great. Only "Show look up cells" crashes here, no segfault just:

```
COLL: Creating collision debug visualization ...
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
  ````

Maybe my machine can't handle it.

only-a-ptr:
@tritonas00 That looks like you're out of memory, try on smaller map and I'll add exception-check code to make sure.

AnotherFoxGuy (approved):
Seems to work fine

tritonas00:
> @tritonas00 That looks like you're out of memory, try on smaller map and I'll add exception-check code to make sure.

Ok cells show fine in auriga, can you try on f1 testtrack?

only-a-ptr:
> @tritonas00 can you try on f1 testtrack?

Well, duh, it's a meshed map, so it's covered entirely in debug meshes.
obrazek

Heck, I'll stop being lazy and do the thing in DearIMGUI.

tritonas00:
> > @tritonas00 can you try on f1 testtrack?
> 
> Well, duh, it's a meshed map, so it's covered entirely in debug meshes. obrazek
> 
> Heck, I'll stop being lazy and do the thing in DearIMGUI.

Meh its fine, my integrated intel gpu just can't handle it

only-a-ptr:
Laziness wins... I curbed the collision cell problem by clamping it to area around rorbot.
obrazek

tritonas00:
No crash now :+1:

tritonas00 (approved):

PR #2941, opened 2022-09-17
RT shader system enabled: per-pixel lighting, PSSM shadows

only-a-ptr:
Kudos to @tritonas00 for research and proof of concept!

obrazek
On display: Terrain https://forum.rigsofrods.org/resources/aspen-grove.43/, Flashlight https://forum.rigsofrods.org/resources/flashlight.957/


only-a-ptr:
I see we finally got rid of this old hack: https://github.com/RigsOfRods/rigs-of-rods/commit/dff13494f1dfada03a19a33d8c87123bbc454e19

tritonas00:
When we port to Ogre 13, i will make a new PR, RTSS and lighting is much better there.

tritonas00:
will be implemented in Ogre 13 branch

Issue #2891, opened 2022-04-03
Controller mapping issues 


tritonas00:
From discord:

> I've been testing out 2022.03, and I'm not sure if it's something I'm doing wrong, but with an xbox controller, I often lose gas and brake. I would say 4/5 times I launch the game, I can't use gas or brake on the controller, but the arrow keys still work. Going back to 2021.10, after launching the game 10 times, I didn't see that issue once.

> Just to follow up on this, I had 2 other people try the latest dev build and they both experienced the shadow flickering, and both of their xbox controllers don't work with gas and brake on the triggers.
Other buttons on the controller work, just not the triggers

> It also looks like the mapping might be different in general, because on 2021.10, start on the controller is the ignition, and A is starter, while in 2022.03, start appears to be the starter and A is save game

> NOT WORKING:
16:21:08: Creating Joystick 1 (Controller (Xbox One For Windows))
16:21:08: * Axes: 6

> WORKING:
16:19:23: Creating Joystick 1 (Controller (Xbox One For Windows))
16:19:23: * Axes: 5

[RoR-_2022.03_Not_Working.log](https://github.com/RigsOfRods/rigs-of-rods/files/8404495/RoR-_2022.03_Not_Working.log)
[RoR_-_2021.10_Working.log](https://github.com/RigsOfRods/rigs-of-rods/files/8404496/RoR_-_2021.10_Working.log)

tritonas00:
> MASTER
16:21:08: OIS Version: 1.5.1
2021.10
16:19:23: OIS Version: 1.4.0

AnotherFoxGuy:
Changelog for OIS is here: https://github.com/wgois/OIS/releases

tritonas00:
> Changelog for OIS is here: https://github.com/wgois/OIS/releases

Could be this https://github.com/wgois/OIS/pull/67/files

We should try with 1.4.0 again. Can you make a package for it?

tritonas00:
ois 1.4.0 works.

PR #2944, opened 2022-09-24
Xbox controller updates

CuriousMike56:
Updated Xbox controller input maps to match the latest version of the Repository file. Also removed the broken PS4 input map (See: https://forum.rigsofrods.org/resources/xbox-controller-input-map.146/updates#resource-update-1238)


vido89:
Should this fix https://github.com/RigsOfRods/rigs-of-rods/issues/2891

tritonas00:
> Should this fix #2891

this is fixed already

PR #2937, opened 2022-09-15
Vehicle lights (flares) improvements.

only-a-ptr:
Closes #2751 (when done).

PR #2946, opened 2022-09-29
Ηeadlights: ignore already attached objects

tritonas00:
https://forum.rigsofrods.org/resources/coca-cola-t600-dry-van-trailer.86/

Load trailer, press N to open lights
Load the truck, press L to lock with the trailer and press N, it tries to re attach trailer's flares (because main vehicle lights trigger trailer lights also) so crash:

```
InvalidParametersException: Object already attached to a SceneNode or a Bone in SceneNode::attachObject at /home/babis/Downloads/ror-dependencies/Source/ogre/OgreMain/src/OgreSceneNode.cpp (line 105)
```

### DIFF
--- a/source/main/physics/Actor.cpp
+++ b/source/main/physics/Actor.cpp
@@ -2944,7 +2944,16 @@ void Actor::lightsToggle()
                     ar_flares[i].light->setVisible(true);
                 ar_flares[i].isVisible = true;
                 if (ar_flares[i].bbs)
-                    ar_flares[i].snode->attachObject(ar_flares[i].bbs);
+                {
+                    try
+                    {
+                        ar_flares[i].snode->attachObject(ar_flares[i].bbs);
+                    }
+                    catch (...)
+                    {
+                        // Already attached
+                    }
+                }
             }
         }
     }


po185021 (changes requested):
Add a comment `//workaround for trailer lights being already on before the truck attached` and log the exception text to console:
```
        catch (Ogre::Exception& eeh) {
            App::GetConsole()->putMessage(Console::CONSOLE_MSGTYPE_INFO, Console::CONSOLE_SYSTEM_ERROR,
                fmt::format("error toggling lights, message:{}", eeh.getFullDescription()));
        }
```
Then approved.

only-a-ptr (changes requested):
Add a comment `//workaround for trailer lights being already on before the truck attached` and log the exception text to console:
```
        catch (Ogre::Exception& eeh) {
            App::GetConsole()->putMessage(Console::CONSOLE_MSGTYPE_INFO, Console::CONSOLE_SYSTEM_ERROR,
                fmt::format("error toggling lights, message:{}", eeh.getFullDescription()));
        }
```
Then approved.

tritonas00:
It can be spammy though, especially with that amount of flares

kk

only-a-ptr:
> It can be spammy though, especially with that amount of flares

in that case just check `m_headlights_on` and don't call `toggleLights()` if already on. Throwing that many exceptions isn't good for FPS.

only-a-ptr:
**Purpose:** to fix a crash (unhandled exception). Apparently there's a bug in synchronizing light states between player vehicle and attached vehicles (trailers etc...).

**Estimate:** To fix correctly, wait for  #2937 to be done - it completely overhauls the light state system, and fixes this in the process. For a dirty workaround suggested by Tritonas, 1 man day should be enough.

tritonas00:
https://github.com/RigsOfRods/rigs-of-rods/pull/2937

PR #2947, opened 2022-10-03
New key map X_Box_360 with wireless receiver for linux

vido89:
New key map X_Box_360 with wireless receiver for linux

### DIFF
--- a/resources/skeleton/config/Microsoft_X_Box_360_pad.map
+++ b/resources/skeleton/config/Microsoft_X_Box_360_pad.map
@@ -1,0 +1,106 @@
+; -------------------------------------------
+; Xbox 360 (Wireless) Controller Input Map
+; Created by CuriousMike
+; Version 1.3
+
+;Buttons:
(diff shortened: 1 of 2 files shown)


vido89:
Sony PLAYSTATION R3 Controller Input map for linux

paroj:
you guys really need to switch to SDL2 for the gamepad abstraction, instead of maintaining your own: https://wiki.libsdl.org/CategoryGameController

vido89:
Yep that would be nice to have because and I quote "so you'll know that SDL_CONTROLLER_AXIS_LEFTX is always the X-Axis of the left Analog Stick" so maybe we could implement SDL2 in to RoR when some developer get some free time tnx @paroj for suggestion

only-a-ptr:
I couldn't resist the ":smile:" reaction because this has been in my backlog since... ever. But there were always bugs attracting a lot more attention, so I never even started. But we'll definitely move to SDL2 eventually.

PR #2949, opened 2022-10-10
Prop anims with 'eventlock' now remain active if player leaves vehicle (+multiplayer).

only-a-ptr:
Kindly requested from Discord by Flip889#6304 and BRobinson#9257

As a bonus, key-triggered prop animations now propagate over network:
busHeadsUnite

RoRnet was bumped to .44


tritonas00 (approved):
Works fine, tested on network also.

PR #2930, opened 2022-08-06
AngelScript: fixed ref counting, added ProceduralRoad API

only-a-ptr:
## Status (last update 2022-10-14)
Many feats originally developed here got merged in https://github.com/RigsOfRods/rigs-of-rods/pull/2931.
The object ref-counting system, originally created here, got perfected in it's own repo: https://github.com/only-a-ptr/RefCountingObject-AngelScript

## New script features:
* ref types: `TerrainClass, ProceduralRoadClass, ProceduralObjectClass, ProceduralManagerClass, ProceduralPointClass`
* enums: `RoadType, TextureFit`

## Reference counting fix

Previously, if a script kept a pointer (a 'handle' in AngelScript terminology) to `BeamClass` or `VehicleAIClass` after the vehicle was removed, it would crash. I studied how the reference counting system works and added a `RefCountingObject` base class which lets us keep track of the object both inside and outside of the script. It will help registering more objects to the script engine.

PR #2953, opened 2022-10-16
Created unified text file tokenizer

ohlidalp:
We have multiple formats with very similar syntax: rig def (truck), odef, tobj, character (see #2942). This new parser supports all these formats and adds new features:
 - **Support for "quoted strings with spaces"**. I first implemented them in the new .character fileformat and I decided I want them everywhere, notably .tobj so that procedural roads can have custom names and descriptions for a GPS-like navigation system.
 - **Simple editing in memory**. The document is a vector of Tokens (types: linebreak, comment, keyword, string, number, boolean.) which keeps the exact sequence as in the file. Tokens can be easily added/modified/removed without writing any extra custom code.
 - **Serializability** - saving the (possibly modified) file is as simple as looping through the Tokens array and writing them to a file. No custom code needs to be written for any file format.
 - **Ease of binding to AngelScript**: a single API can modify any fileformat. All the operations the user needs is 1. insert token 2. modify token 3. delete token.

Update: I added AngelScript bindings and extended the bundled 'demo_script.as' to showcase them. There's a new button "View document" (changes to "Close document" after pressing) which opens separate window with a syntax highlighted truck file. Update2: all glitches were fixed. It's ready for review.

**New script API:**
* class GenericDocumentClass - loads/saves a document from OGRE resource system
* class GenericDocReaderClass - traverses document tokens
* enum TokenType (TOKEN_TYPE_*)
* enum GenericDocumentOptions (GENERIC_DOCUMENT_OPTION_*)
* void ImGui::AlignTextToFramePadding()
* string Terran::getTerrainFileName()
* string Terrain::getTerrainFileResourceGroup() 

To test, open game console and say `loadscript demo_script.as`.
obrazek

### DIFF
--- a/doc/angelscript/Script2Game/GenericDocReaderClass.h
+++ b/doc/angelscript/Script2Game/GenericDocReaderClass.h
@@ -1,0 +1,55 @@
+namespace Script2Game {
+
+/** \addtogroup ScriptSideAPIs
+ *  @{
+ */    
+
+/** \addtogroup Script2Game
+ *  @{
+ */   
+
+enum TokenType
+{
+    TOKEN_TYPE_NONE,
+    TOKEN_TYPE_LINEBREAK,
+    TOKEN_TYPE_COMMENT,
+    TOKEN_TYPE_STRING,
+    TOKEN_TYPE_NUMBER,
+    TOKEN_TYPE_BOOL,
+    TOKEN_TYPE_KEYWORD
+};
+
+/**
+ * @brief Binding of RoR::GenericDocReader; Traverses document tokens; See 'demo_script.as' for an example.
+ */
+class GenericDocReaderClass
+{
+    GenericDocReader(GenericDocumentPtr@ d);
+
+    bool MoveNext();
+    uint GetPos();
+    bool SeekNextLine();
+    int CountLineArgs();
+    bool EndOfFile(int offset = 0);
+
+    TokenType GetTokType(int offset = 0);
+    string GetStringData(int offset = 0);
+    float GetFloatData(int offset = 0);
+
+    string GetTokString(int offset = 0);
+    float GetTokFloat(int offset = 0);
+    bool GetTokBool(int offset = 0);
+    string GetTokKeyword(int offset = 0);
+    string GetTokComment(int offset = 0);
+
+    bool IsTokString(int offset = 0);
+    bool IsTokFloat(int offset = 0);
+    bool IsTokBool(int offset = 0);
+    bool IsTokKeyword(int offset = 0);
+    bool IsTokComment(int offset = 0);
+};
+
+/// @}    //addtogroup Script2Game
+/// @}    //addtogroup ScriptSideAPIs
+
+} //namespace Script2Game
--- a/doc/angelscript/Script2Game/GenericDocumentClass.h
+++ b/doc/angelscript/Script2Game/GenericDocumentClass.h
@@ -14,7 +14,10 @@ enum GenericDocumentOptions
     GENERIC_DOCUMENT_OPTION_ALLOW_SLASH_COMMENTS, //!< Allow comments starting with `//`. 
     GENERIC_DOCUMENT_OPTION_FIRST_LINE_IS_TITLE, //!< First non-empty & non-comment line is a naked string with spaces. 
     GENERIC_DOCUMENT_OPTION_ALLOW_SEPARATOR_COLON, //!< Allow ':' as separator between tokens.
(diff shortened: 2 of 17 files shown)


ohlidalp:
When I'm looking at the serialization loop I wrote, I realize that the way you tell good code isn't about diving into the details, but rather looking from bird's perspective where you can barely read the characters, then observing the overall shape and using just common sense to figure out what it probably does. If the above journey is a pleasure to do and the conclusion you arrive at turns out to be pretty much correct, you've met good code.
obrazek
Yes, I've had a 🤓 moment, judge me all you want.

tritonas00:
This is risky, there are possibilities for grave stone.

f1 testtrack

screenshot_2022-10-18_18-45-06_1

only-a-ptr:
OK, new plan... I'll test this on #2942 which doesn't touch existing formats, and after that's merged, I'll experiment with existing formats.

only-a-ptr:
New plan v2: I'll finish #2930 (which I will anyway) and then bind this new parser to AngelScript. That way we get the best of both worlds - the game will use the tried old parsers and the user will still be able to edit TOBJ/ODEF/TRUCK files using this parser.

only-a-ptr:
**Purpose:** to be able to read, edit and export any fileformat (TRUCK DEF, ODEF, TOBJ, CHARACTER) directly in game using AngelScript.

**Estimate:** Blocked by #2930. As soon as #2930 is done, this will take at most 1 man day to finish.

**Work to be done:**
 * Export elements in 'GenericFileFormat.h' to AngelScript.
 * Add AngelScript API  `Document@ game.loadDocument(filename, rg_name)` and `void game.saveDocument(Document@)` that will check if the file is unzipped and if yes, load/save the document.
 * Extend the 'demo_script.as' to showcase this new feature.

CuriousMike56:
Viewing truck file of any 'complex' vehicle -> ~30 fps drop:
RoR_2022-12-10_20-27-55
No frame drops with the DAF Semi. 
Also much of the truck file doesn't match the actual file:

only-a-ptr:
@CuriousMike56 
>Viewing truck file of any 'complex' vehicle -> ~30 fps drop:

Makes sense, it's because the script traverses the entire document every frame, so it's combined overhead of DearIMGUI and string allocation by angelscript. I'd have to optimize it like in Console UI - artificially scale the scrollbar and only populate the visible part of the document. But I'd like to do it in another PR since it's just a demo.

> Also much of the truck file doesn't match the actual file:

Interesting, I see 3 problems:
* For each number, only first digit is displayed - IDK what happened.
* `set_*_defaults` are omitted entirely - again, no idea why.
* Empty lines are not preserved - this is a script issue, it relies on imgui not doing `SameLine()` which works for single linebreaks but squashes multiple linebreaks.

only-a-ptr:
[Truck definition format](https://docs.rigsofrods.org/vehicle-creation/fileformat-truck/) parsing is complete.

Note these special cases:
1. the option 'ALLOW_NAKED_STRINGS' must be on at all times, because otherwise the parser requires all strings to be in quotes. Also, the first line is special - it's the actor name, with spaces and special characters, possibly including multiple quotes. I added an option "FIRST_LINE_IS_TITLE" which parses it as one string.
obrazek
2. in `forset`, single node numbers are parsed as NUMBER. Node ranges are not valid numbers so they decay to STRINGs. The option 'ALLOW_NAKED_STRINGS' must be on.
obrazek
3. The `axle` and `interaxle` keyword have a quirky `abc(1 2)` syntax. I added an option 'PARENTHESES_CAPTURE_SPACES' which make the whole expression parse as one string. The option 'ALLOW_NAKED_STRINGS' must be on.
obrazek

only-a-ptr:
Originally I only planned to cover fileformats which are similar to each other (truck, odef, tobj) but now I wanted the demo script to display TOBJ files too, but I didn't want to clog the Terrain API with `getFileWhatever()` funcs. I wanted to use the `GenericDocument` to parse the TERRN2 file, get TOBJ filenames from there and parse those as well. It turned out to be pretty straightforward, the tokenizer is already robust so it could take a few extra options.

* **Parsing TERRN2 format** needs these special options: "ALLOW_HASH_COMMENTS", "ALLOW_SEPARATOR_EQUALS", "ALLOW_BRACED_KEYWORDS" and of course "ALLOW_NAKED_STRINGS".
obrazek

* **Parsing TOBJ format** needs just "ALLOW_NAKED_STRINGS" and "ALLOW_SLASH_COMMENTS".
obrazek

CuriousMike56 (approved):
Seems to work fine 👍

PR #2959, opened 2022-10-21
:sparkles: Added polygon mode (aka render mode) as cvar

### LINKED ISSUE #2957: Re-implement wireframe mode
In #2926, the "E" hotkey for wireframe mode was removed. While I agree with the hotkey itself being removed, it seems the entire wireframe mode feature was also removed as a result. Having a wireframe mode can be very useful when modding, especially when prop placing (Sometimes the prop will appear under the terrain, and the only way you'll be able to see it is by toggling wireframe). 

I believe wireframe mode should be re-implemented as either a top menu setting or a console command. Making it a console command would probably be preferable if the goal is to hide it from the average player.

Zentro:
Resolves #2957 

`PM_SOLID` is the default when rendering, `setint gfx_polygon_mode 3`
`PM_WIREFRAME` to show the wireframe `setint gfx_polygon_mode 2`
`PM_POINTS` to show the points of each wireframe `setint gfx_polygon_mode 1`

### DIFF
--- a/source/main/Application.cpp
+++ b/source/main/Application.cpp
@@ -202,6 +202,7 @@ CVar* audio_menu_music;
 
 // Graphics
 CVar* gfx_flares_mode;
+CVar* gfx_polygon_mode;
 CVar* gfx_shadow_type;
 CVar* gfx_extcam_mode;
 CVar* gfx_sky_mode;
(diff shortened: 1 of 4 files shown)


tritonas00:
It works fine :+1:

only-a-ptr:
Thanks!

PR #2961, opened 2022-10-28
:sparkles: Added polygon mode to tools menu

CuriousMike56:
Might not be the best solution but I think a scroll bar works nicely for this.

https://user-images.githubusercontent.com/46073351/198513521-14122c34-5634-4998-8e13-4856f60d8539.mp4


only-a-ptr (approved):
I forgot you had code skills like this. Nice work.

PR #2922, opened 2022-07-17
Parser corrections (nodes2, forset, sectionconfig)

ohlidalp:
I reverted the parser to the original (2013 and earlier) sequential mechanic.
That means things that one worked are back, but so are some restrictions that didn't apply in the time between.

**ALPHA STATUS - Features and restrictions can still change.**

**Returning features:**
* everything `except authors, fileinfo, guid, description` can now be in `section/end_section`.
* you can add various number of nodes2 in each `section/end_section`.
* all elements except `forset` support named nodes (nodes2).
* you can specify multiple configs for a single `section`, like this: 
   ```
   sectionconfig 0 conf1
   sectionconfig 0 conf2

   section 0 conf1, conf2
   ; applies to both!
   end_section
   ```
**Returning restrictions:**
* the `fusedrag autocalc` only takes in consideration nodes defined before it, so by defining them after it you basically instruct the game to exclude them from the autocalculation.
* the `managedmaterials` must be specified before the props/flexbodies/flares which use them, otherwise they will end up all white.
* Essentially all the "A must come before B" restrictions are back.

**Bugfixes:**
* Fixes #171
* Fixes #2881

**Improvements:**
* `forset` is now a regular directive - you can put comments, empty lines or even section in between the `flexbody` proplike-line and the `forset`. 
* in the selector UI, the configuration combobox is now on top and the values below (click the ShowDetails checkbox) are accurately counted for that configuration. The counts are all saved in modcache. Cachefile format version was bumped.
obrazek

PR #2964, opened 2022-11-06
Parser: Ignore 'v' beam option

CuriousMike56:
Quite a few mods use the 'v' beam option to make beams visible, even though they're actually visible by default. This prevents the console from getting spammed by warnings:
image
I don't really know how the parser works but this seems to do the job.


only-a-ptr (approved):
Yes, that's how it's supposed to work. Thanks!
I noticed but neglected it because I thought I'd finish https://github.com/RigsOfRods/rigs-of-rods/pull/2922 much sooner.

PR #2967, opened 2022-11-18
Fixed ControlsUI not showing built-in defaults

only-a-ptr:
only entries actually present in 'input.map' were shown, other controls were missing.

If you edit an entry which wasn't present in 'input.map' before, it will be added.

### DIFF
--- a/source/main/gui/panels/GUI_GameControls.cpp
+++ b/source/main/gui/panels/GUI_GameControls.cpp
@@ -160,7 +160,7 @@ void GameControls::DrawEvent(RoR::events ev_code)
 
     // Check if we have anything to show
     int display_count = 0;
-    for (event_trigger_t& trig: triggers)
+    for (event_trigger_t& trig : triggers)
     {
         display_count += (int)this->ShouldDisplay(trig);
     }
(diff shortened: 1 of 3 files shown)


tritonas00:
With this controls panel indeed show all keys but it confusing things. Purpose of controls panel is to show, edit and preserve values in input.map, not hardcoded values.

only-a-ptr:
@tritonas00 The way you describe it, overriding a default means the user must first manually add the entry to input.map and only then they are allowed to use the editing capability of the Settings UI. This makes no sense to me - I'd expect it to display everything and add modified entries to input.map.

tritonas00:
So if i make a change on one of those extra controls it will be preserved in input.map? If yes then its good.

CuriousMike56:
I don't see any problems with this change besidss the hardcoded text already mentioned. However I did just notice `CHARACTER_RUN` (https://github.com/RigsOfRods/rigs-of-rods/blob/master/resources/skeleton/config/input.map#L72) has two keys assigned in `input.map` which results in the second key name being blank:
image
This isn't a problem introduced by this PR, but I figured it makes sense for it be fixed here.

only-a-ptr:
@CuriousMike56 What seems to you like a blank line is actually single row with double vertical size. But yes, it's confusing, so I added small numbers to make it obvious:
obrazek
I didn't figure any solution to the hardcoded hotkeys in top menubar, but I'd let it pass for now.

CuriousMike56 (approved):

Zentro (approved):

PR #2980, opened 2022-12-06
Only display terrain object errors if it's user's fault

only-a-ptr:
When the map is flawed, there's no point in throwing that errors on the player - I made a bypass where the errors display only during simulation (ODEFs loaded from script or console), not during map loading.


CuriousMike56:
👍

PR #2982, opened 2022-12-09
:bug: Fixed missing ogre tools

AnotherFoxGuy:


only-a-ptr (approved):
Tested on Win10 + VisualStudio 2022, works perfectly.

PR #2986, opened 2022-12-12
AI: export ready to use json formatted waypoint list

tritonas00:
The result can be copy pasted directly in https://github.com/RigsOfRods-Community/ai-waypoints/blob/main/waypoints.json, for convenience.


only-a-ptr (approved):
It'll do 😄

PR #2996, opened 2022-12-29
Added setting to use alternative vehicle materials

CuriousMike56:
After the release of 2022.12, some players apparently prefer the other vehicle materials instead of the current NiceMetal-based materials. This brings those back by adding "Use alternative vehicle materials" to the graphics settings.

~PR is currently a draft as~ I'd like to also fix the slight flickering the materials have, introduced after 2020.01: 
https://streamable.com/3unrvg
After tinkering with the materials a little, I managed to stop the flickering by commenting out all `depth_bias` lines in `managed_mats_vehicles.material` and changing the value in `shadows.material` to `depth_bias -20`. However, with this change vehicles now look like this when the camera is far away:
image
(Lowered FOV)

### DIFF
--- a/resources/managed_materials/managed_mats_vehicles.material
+++ b/resources/managed_materials/managed_mats_vehicles.material
@@ -1,14 +1,12 @@
-import * from "shadows.material"
 import * from "texture_manager.material"
 
 material managed/flexmesh_standard/simple
 {
-	technique BaseTechnique: Shadows/managed/base_receiver //Include shadows
+	technique BaseTechnique
 	{
 	//Texture is rendred here
 		pass BaseRender
 		{
-			depth_bias -0.1 -0.1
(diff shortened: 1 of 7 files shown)


CuriousMike56:
Discussed this with @tritonas00 and we believe its best to just disable shadow receiving for now.

tritonas00:
Works fine here.

only-a-ptr (approved):
If I understand it correctly, we're basically restoring the 'classic materials' option, just with an inverted meaning (disables NiceMetal rather than enables). Code looks OK.

CuriousMike56:
@tritonas00 @only-a-ptr Good to merge?

tritonas00:
Its fine with me. But we keep cycling here instead to finish Ogre 13 port and redo the materials properly.

PR #3001, opened 2023-01-03
Added support for actor scripts (use section 'scripts' in truck file)

only-a-ptr:
The scripts work the same as terrain/custom scripts but they have an extra global variable `thisActor` pointing to the associated actor.

obrazek

obrazek

Elementary test vehicle:
[dafsemi-scripttest.zip](https://github.com/RigsOfRods/rigs-of-rods/files/10339592/dafsemi-scripttest.zip)

### DIFF
--- a/resources/scripts/AI.as
+++ b/resources/scripts/AI.as
@@ -1,9 +1,5 @@
 #include "base.as"
 
-//Stuff that the scripts uses internally. doesn't need to be changed.
-vector3 currentWaypoint(0, 0, 0);
-int waypoint = 0;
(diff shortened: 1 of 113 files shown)


CuriousMike56 (commented):
👍

CuriousMike56 (approved):

only-a-ptr:
Known issue: save a game while driving (using top menubar), then load game using top menubar... the vehicle will not be loaded.
Note it doesn't happen for vehicles standing by, only for driven ones.

Issue #2998, opened 2022-12-29
Rotating camera with mouse shows vehicle buttons menu


CuriousMike56:
Currently the 'vehicle buttons' menu shown at the top left of the screen always appears when you hover over it with the mouse. If you are rotating the camera with the mouse (holding right click) and happen to move the mouse to the top left, the menu will appear regardless. It would be nice if this didn't happen.

tritonas00:
Same with top menu. Should be easy to fix though.

PR #3006, opened 2023-01-12
Don't show top menu bar and vehicle buttons when rotating camera

tritonas00:
Fixes https://github.com/RigsOfRods/rigs-of-rods/issues/2998


CuriousMike56:
Works fine

PR #3007, opened 2023-01-13
Any node now activates script event boxes

only-a-ptr:
Previously only the main camera node activated eventboxes. This was a bad fit for i.e. races vs. the finishing line.

The process of invoking the event boxes was inverted - instead of searching an event box for the camera node, all event boxes close to the actor are searched for a hit with any node. Hits are recorded and next time the recorded colliding node is checked first.

I checked performance impact on Auriga race under Debug. There seems to be no noticeable FPS drop.

Code changes:
* Actor::CalcForcesEulerCompute() - call to UpdateBoundingBoxes() moved here from CalcNodes(). Added call to CalcEventBoxes().
* Actor::CalcEventBoxes() - searches and records hits with eventboxes, as described above.
* Collisions::nodeCollision() - no longer processes event boxes, the parameter was removed.

### DIFF
--- a/source/main/physics/Actor.h
+++ b/source/main/physics/Actor.h
@@ -402,6 +402,8 @@ public:
     float             ar_collision_range;             //!< Physics attr
     float             ar_top_speed;                   //!< Sim state
     ground_model_t*   ar_last_fuzzy_ground_model;     //!< GUI state
+    CollisionBoxPtrVec m_potential_eventboxes;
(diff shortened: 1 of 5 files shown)


CuriousMike56 (approved):

only-a-ptr:
Performance seems to be OK:
* MrMackey (SimMonsters)/Windows: https://discord.com/channels/136544456244461568/189904947649708032/1063653232129347625
* CuriousMike/Windows: https://discord.com/channels/136544456244461568/189904947649708032/1063648731301429258

PR #3016, opened 2023-02-07
Fixed white managedmat if specular/damage texture missing

### LINKED ISSUE #3005: Missing managedmaterials specular texture results in white meshes
Effected mods: 
[Lancaster Bomber](https://forum.rigsofrods.org/resources/lancaster-bomber.203/)
[M/V Nordia Express](https://forum.rigsofrods.org/resources/m-v-nordia-express.191/) (old version)

Similar issue as #2999. In previous versions, managed material lines with a missing or invalid specular image would simply ignore the error and load the material with no specular. Now in 2022.12+, the line(s) are ignored and you end up with solid white meshes.

2022.04:
image
image

Master:
image
image

only-a-ptr:
This was a bug introduced with 0e8e64644c685ccdd1904c53260269d1473406a0 - managedmaterial processing got refactored and new logic didn't match the old logic - new checks for missing textures were overzealous and disabled the entire material instead of disabling just the related effect.

Fixes #3005

### DIFF
--- a/source/main/physics/ActorSpawner.cpp
+++ b/source/main/physics/ActorSpawner.cpp
@@ -2261,17 +2261,19 @@ void ActorSpawner::ProcessManagedMaterial(RigDef::ManagedMaterial & def)
         this->AddMessage(Message::TYPE_WARNING, "Skipping managed material, missing texture file: " + def.diffuse_map);
         return;
     }
+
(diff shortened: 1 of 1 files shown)


CuriousMike56:
👍 👍

PR #3019, opened 2023-02-26
Script func coding style fix

only-a-ptr:
It just struck me that the freshly introduced `GenericDocumentClass` and `GenericDocReaderClass` script objects use `PascalCaseFuncNames()`, while all other script objects use `camelCaseFuncNames()`. I rushed to fix this flaw until the objects gain wider use. The 'demo_script.as' which showcases them was also updated.

It also turned out docs for `VehicleAIClass` are wrong - they showed `PascalCaseFuncs()` while the bindings actually had `camelCaseFuncs()`. I fixed a great number of quirks, mistakes, typos and omissions in both the source and docs - see the commit msg for the boring details. And while at it, I also upgraded `VehicleAI` to `RefCountingObjects<>` to fix the broken reference counting.

UPDATE: The last non-conforming object was `LocalStorage` - technically it was working (I think), but overly complicated. I upgraded it to `RefCountingObjects<>` for consistency and took the opportunity to rename it to `LocalStorageClass` - now all the objects are consistent. The 'races.as' script which uses it was updated.

**Summary of changes**:
* `GenericDocumentClass` - renamed functions from `PascalCase()` to `camelCase()`
* `GenericDocReaderClass` - ditto
* `VehicleAIClass` - added func `isActive()`, fixed buggy `addWaypoints()`, fixed fractured docs.
* `LocalStorageClass` - renamed from `LocalStorage`, added function `copyFrom()`, added constructor parameter "resource group" (like `GenericDocumentClass` has) to be able to write to alternate locations (i.e. savegames).


tritonas00 (approved):
Works fine it seems.

Issue #3021, opened 2023-02-28
Incomplete electric "e" engine type


CuriousMike56:
In the early days of "NextStable" development, Max98 implemented an electric engine type:
https://github.com/RigsOfRods/rigs-of-rods/blob/7166295be905698e39eb3e42586ffaddf51060b9/source/main/gameplay/EngineSim.cpp#L238-L247
https://github.com/RigsOfRods/rigs-of-rods/blob/7166295be905698e39eb3e42586ffaddf51060b9/source/main/gameplay/EngineSim.cpp#L450-L472

I used this option on my submeshed Cybertruck: https://forum.rigsofrods.org/resources/tesla-cybertruck.634/

It technically works as intended but has two bugs:

- It is not possible to shift into neutral or reverse in automatic shift mode.
- If "Engines spawn running" is disabled in settings, the vehicle won't start.

PR #3022, opened 2023-03-02
Electric car bugfixes.

only-a-ptr:
Fixes both issues reported in #3021

Note you still need to turn on engine ignition and hit starter.


CuriousMike56 (commented):
Automatic shift works great now, but the "engines spawn running" bug still persists:
- Uncheck "Engines spawn running" from Settings -> Gameplay
- Cybertruck no longer moves as its now not possible to turn it on by pressing X. **EDIT:**: This may be due to the fact the starter is disabled.

Not related to this PR, but I just noticed with "Engines spawn running" disabled, engine RPMs are maintained after a reset (backspace). Pretty strange behavior IMO.

only-a-ptr:
I originally tested with "arcade controls" enabled which auto-starts the engine when pressing forward/reverse.
Now I also fixed the manual starter control.

CuriousMike56 (approved):
Perfect 👍

PR #3024, opened 2023-03-06
TOBJ fileformat: added `set_default_rendering_distance`

only-a-ptr:
TOBJ fileformat: added `set_default_rendering_distance`

This is a directive - you can place it anywhere and it affects all lines that come after it, much like directives in the truck fileformat do.
The parameter is distance in meters. 0 means infinite distance (object is always rendered). Default value is 0.
Internally, it uses `Ogre::Entity::setRenderingDistance()`, see https://ogrecave.github.io/ogre/api/1.11/class_ogre_1_1_movable_object.html#afe1f2a1009e3f14f36e1bcc9b1b9557e

Examples:
```
; limit distance to 10 meters
set_default_rendering_distance 10.0
; reset to infinite distance
set_default_rendering_distance 0
```

NOTE: I observed there is some glitch in the distance calculation, even with the same distance value, some objects appear faster than others. I debugged my code and concluded it's an OGRE issue.

### DIFF
--- a/source/main/resources/tobj_fileformat/TObjFileFormat.cpp
+++ b/source/main/resources/tobj_fileformat/TObjFileFormat.cpp
@@ -63,6 +63,7 @@ void TObjParser::Prepare()
     m_cur_procedural_obj = new ProceduralObject();
     m_cur_procedural_obj_start_line = -1;
     m_road2_num_blocks = 0;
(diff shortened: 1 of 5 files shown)


CuriousMike56 (approved):
Works fine

Issue #3032, opened 2023-03-28
Tire/wind sounds continue playing after vehicle deletion


CuriousMike56:
- Spawn a vehicle featuring tire/wind sound (e.g. any Gavril) 
- While driving, remove the vehicle (CTRL+DELETE)
- Tire/wind sound will continue playing, even if you return to the main menu.

PR #3027, opened 2023-03-20
Angelscript audio bindings

only-a-ptr:
The screenshot shows updated demo script bundled with the game. To run it, open console and say `loadscript demo_script.as`.
obrazek

New objects: `SoundClass`, `SoundScriptTemplateClass`, `SoundScriptInstanceClass`
New enums: `SoundTrigger`, `ModulationSource`

The principal interface is in the `GameScriptClass` (global object `game`):

```
array<SoundScriptTemplateClass@>@ getAllSoundScriptTemplates()
SoundScriptTemplateClassPtr@ getSoundScriptTemplate(const string &in)
array<SoundScriptInstanceClass@>@ getAllSoundScriptInstances()
SoundClassPtr@ createSoundFromResource(const string &in filename, const string &in rg_name = string())
SoundScriptInstanceClassPtr@ createSoundScriptInstance(const string &in template, int actor_instance_id = -1)
```

PR #3038, opened 2023-04-30
Fixed actor sounds not muting when deleting

tritonas00:
fixes https://github.com/RigsOfRods/rigs-of-rods/issues/3032

introduced with https://github.com/RigsOfRods/rigs-of-rods/commit/6d0bf3576f9413f1bf44452d49361887acc8bce5


only-a-ptr:
Fixed - my fault from first minute, I set the DISPOSED flag too early so the safety checks kicked in and actually broke the process.

As bonus and because I just merged https://github.com/RigsOfRods/rigs-of-rods/pull/3027, I added cleanup of `SoundScriptInstance`-s to the disposal process.

tritonas00:
:+1:  works as it should

PR #3040, opened 2023-05-04
Diagnosing 'spawning and deleting actor can randomly segfault'

only-a-ptr:
https://discord.com/channels/136544456244461568/189904947649708032/1103748772846108672


only-a-ptr:
Duplicate of #3042 😅

PR #3046, opened 2023-05-26
Always log CURL error string

only-a-ptr:
At one moment the server had a certificate expired and I realized the game doesn't log anything useful in that situation. The HTTP response code remains '0' in those cases.

Seeing all the copy-pasted and mostly identical `curl_easy*()` calls, I should probably create a helper for it, but at the moment I just uniformly extended all the occurences with an error logging code.

### DIFF
--- a/source/main/Application.h
+++ b/source/main/Application.h
@@ -85,10 +85,10 @@ enum MsgType
     MSG_NET_USER_DISCONNECT,
     MSG_NET_RECV_ERROR,
     MSG_NET_REFRESH_SERVERLIST_SUCCESS,    //!< Payload = GUI::MpServerInfoVec* (owner)
-    MSG_NET_REFRESH_SERVERLIST_FAILURE,
+    MSG_NET_REFRESH_SERVERLIST_FAILURE,    //!< Payload = RoR::CurlFailInfo* (owner)
     MSG_NET_REFRESH_REPOLIST_SUCCESS,      //!< Payload = GUI::ResourcesCollection* (owner)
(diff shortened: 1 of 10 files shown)


CuriousMike56:
Good time to test this as the repo API is currently broken:
`15:32:52: [RoR|Repository] Failed to retrieve repolist; Error: 'SSL peer certificate or SSH remote key was not OK'; HTTP status code: 0`
Could the error be added to the UI? Right now it just says 
image

tritonas00:
> Good time to test this as the repo API is currently broken: `15:32:52: [RoR|Repository] Failed to retrieve repolist; Error: 'SSL peer certificate or SSH remote key was not OK'; HTTP status code: 0` Could the error be added to the UI? Right now it just says image

I don't think the player needs to know if the cert expired or any verbose info in general, thats why i texted it like that.

Zentro:
> > Good time to test this as the repo API is currently broken: `15:32:52: [RoR|Repository] Failed to retrieve repolist; Error: 'SSL peer certificate or SSH remote key was not OK'; HTTP status code: 0` Could the error be added to the UI? Right now it just says image
> 
> I don't think the player needs to know if the cert expired or any verbose info in general, thats why i texted it like that.

I think its useful information to include. Maybe just a server error code 4xx or 5xx, just something to indicate with a bit more information when someone posts a screenshot of an issue they have.

tritonas00:
Eitherway its just a matter of seconds to do it, all info is there.

only-a-ptr:
Thanks for the input. I think it should at least say "Server error, ask support" instead. Some detail below in darkened text can't hurt either

only-a-ptr:
It now displays CURL error under the title, and if HTTP code is valid, it's also displayed.
obrazek

PR #3047, opened 2023-05-28
Node affectors

ohlidalp:
Affectors are actuators which apply force to nodes. Applying forces from script or GUI is not trivial because physics runs at 2000FPS while Script/GUI are updated once per frame (roughly 100FPS). Until now there was 1 "slot" for a mouse-cursor grab force. Affectors are an extension of this system - you can add multiple forces this way and control them in various ways. The mouse grab system was remade using affectors.
Planned features (**general status: alpha version**):

- **uniform force affector** - applies constant force in constant direction to nodes. Status: stub, no scripting.
- **pinned force affector** - pulls nodes towards a given point. Status: working (used in mouse grab), scripting not tested.
- **thruster affector** - applies force within a frame of reference given by nodes `noderef/nodex/nodey`, like aeroengines/screwprops. Status: stub.
- **live editing** - use `MSG_EDI_ADD/REMOVE_AFFECTOR_REQUESTED` with arguments-struct. This works (used in mouse grab).
- **script editing** - call `game.pushMessage(MSG_EDI_ADD/REMOVE_AFFECTOR_REQUESTED, {})` with a dictionary of params. Status: ADD is coded but not tested, REMOVE isn't coded yet.
- **force pinning with INSERT** - an extension to grab mode. Use right mouse button to unpin again. This works, see GIF below.
- **theme editor UI** - because the debug views get more complex and color definitions are hardcoded all over the place. And while I'm consolidating it, it's a small job to do a theme editor. Status: stub, not usable yet.

dafMousePinForces

### DIFF
--- a/source/main/Application.h
+++ b/source/main/Application.h
@@ -117,6 +117,8 @@ enum MsgType
     MSG_EDI_ENTER_TERRN_EDITOR_REQUESTED,
     MSG_EDI_LEAVE_TERRN_EDITOR_REQUESTED,
     MSG_EDI_RELOAD_BUNDLE_REQUESTED,       //!< Payload = RoR::CacheEntry* (weak)
+    MSG_EDI_ADD_AFFECTOR_REQUESTED,        //!< Payload = RoR::AddAffectorRequest* (owner)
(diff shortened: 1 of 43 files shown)


ohlidalp:
Closing this in favor of the superior FreeForces: #3159.

I will create a separate PR for the mouse-grab node highlight, that feature deserves another chance.

PR #3049, opened 2023-05-29
Single player character collisions

tritonas00:
Single player character with actors collisions. New option added in top menu bar.

https://streamable.com/gttxum


tritonas00:
To do:
- Make AI vehicle ignore the collided character or else if it stands too close in front of it, it will try to avoid and turn.
- Check `App::sim_character_collisions->getBool()` here also https://github.com/RigsOfRods/rigs-of-rods/blob/eb7435d2d4a5836c84e451581c4c4c507e09a201/source/main/gameplay/Character.cpp#L180-L189

only-a-ptr:
Closing this in favor of #3050 because there's a common issue - character slides when the vehicle turns - and I'm crafting a common fix.

PR #3050, opened 2023-05-30
🎮 Enabled character/actor collision in multiplayer. 

ohlidalp:
This is an experimental spinoff from https://github.com/RigsOfRods/rigs-of-rods/pull/3049. I felt inspired.
characterCollNetworkedDaf

Since the collision works by "sticking" the character to the actor while in contact, I realized I could extend our existing driver-attachment logic to also handle this attachment.

It's glitchy right now, partly because the networked cab offset is always in world coordinates, so it causes sliding when the vehicle turns.

### DIFF
--- a/source/main/Application.cpp
+++ b/source/main/Application.cpp
@@ -108,6 +108,7 @@ CVar* sim_no_self_collisions;
 CVar* sim_gearbox_mode;
 CVar* sim_soft_reset_mode;
 CVar* sim_quickload_dialog;
+CVar* sim_character_collisions;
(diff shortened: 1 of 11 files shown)


only-a-ptr:
I'm trying to fix the turn-sliding problem and I'm not vey successful. I managed to glue the character onto the cab so he doesn't slide but can't walk either and the positioning is still off. See commit (DEBUG - DIRTY) super-glued rorbot on cab, sliding gone but so is walking
obrazek

PR #3055, opened 2023-06-15
Fixed multiplayer visual glitch with detached parts "teleporting" around the vehicle

### LINKED ISSUE #2783: Parts of a vehicle that fall of "teleport" and follow the vehicle instead of staying where it landed. (Online Issue)
Since about 0.4.5, the body parts that fell off did stay where they landed, back in 0.39 and below. Now, when a body part falls off on another person's screen, it will follow them and teleport in a grid, following the truck. It's a bit difficult to explain, but anything will do this if it falls off, even tires. Here's a video that shows how the body part will act when it falls off.

https://youtu.be/2gzDuGPqKms?t=1432

ohlidalp:
Fixes #2783

I replaced the Ulteq-invented `short int` compression format with a standard half-precision IEEE floats (using https://half.sourceforge.net/). Both are 2-byte formats so network traffic will remain the same.

### DIFF
--- a/external/header_only/half/ChangeLog.txt
+++ b/external/header_only/half/ChangeLog.txt
@@ -1,0 +1,213 @@
+Release Notes											{#changelog}
+=============
+
+2.2.0 release (2021-06-12):
+---------------------------
+
+- Added `rsqrt` function for inverse square root.
+- Improved performance of `pow` function.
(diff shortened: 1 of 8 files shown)


ohlidalp:
Needs a rornet bump: https://discord.com/channels/136544456244461568/189904947649708032/1119932306036097047

PR #3042, opened 2023-05-08
Fix #3041 - Random crashes when you spawn/switch/delete actor

ohlidalp:
See #3041
I fixed one potential crash due to dangling GfxActor pointer (hook holding pointer to DISPOSED actor which doesn't have GfxActor anymore). 

UPDATE: This evolved into signifficant cleanup of the Actor internals. The motivation was to use safer memory handling with added debug checks, but it will also help future extensions.  The original problem turned out to be using non-threadsafe `RefCountingObject<>`s from physics worker threads - see conclusion: https://github.com/RigsOfRods/rigs-of-rods/pull/3042#issuecomment-1557066997

UPDATE2 (2023-05-29): I surrendered and threw a mutex in `RefCountingObject`. Not proud of this at all, but I apparently used it wrong in too many places to look for it now. Technical debt ahoy!

UPDATE3 (2023-06-18): Closing in favor of #3057 (fixes the crashes) and #3059 (refactors the Actor memory management).

PR #3057, opened 2023-06-17
Fix #3041 - Random crashes when you spawn/switch/delete actor (lite edition)

### LINKED ISSUE #3041: Random crashes when switching/spawning/deleting actors
Issue ticket for the random crashes that occur when spawning/deleting/switching actors, originally brought up by @tritonas00 in #development on Discord. Related PR: #3042 

I'm able to consistently reproduce this by loading [Windmill Mountain](https://forum.rigsofrods.org/resources/windmill-mountain.500/) and switching to the first "Wind Engine Tower" on the vehicle list, the game will crash a few seconds after. Oddly enough, attempting to load that terrain with a Debug build crashes with: 
devenv_2023-05-06_12-57-45
(still not sure of the correct way to share VS back traces)
Release build back trace after switching to the wind tower actor:
devenv_2023-05-06_13-07-02

only-a-ptr:
A lightweight take on #3041 - split off from #3042 with only the bare essentials.

Please test these scenarios: https://github.com/RigsOfRods/rigs-of-rods/pull/3042#issuecomment-1567306964

Turns out I was really careless when I made `class Actor` a `RefCountingObject<>` - sure, the refcounting was broken before and this helper works flawlessly, except you have to remember not to use it from threads, and I didn't even think about it. For now I worked it around by tossing a mutex in `RefCountingObject<>`, but a sensible approach would be to just use IDs in threads and only manipulate the objects on main thread via message queue.


only-a-ptr:
**Conclusion:** I regret nothing. `RefCountingPtr` works flawlessly, the refcount is now thread-safe but there's a debug `assert()` which ensures only the main thread manipulates refcount so there's no locking overhead. I also learned some new tricks and the final solution is not only workable but even ellegant.

CuriousMike56:
Crashes gone, great news! But now:
image
If a negative action key is apparently incorrect, shouldn't the value be parsed as positive instead? The same errors also occur on the "0.4.0.7 only" versions of trains such as the [Henschel DHG 700C](https://forum.rigsofrods.org/resources/henschel-dhg-700c.171/)

only-a-ptr:
I restored support for negative commandkeys in 'triggers' - they're now redirected to a hashmap so any value is acceptable. Though I was unable to figure out what is their meaning. My best guess is, since they caused a memory write outside the `commandkeys` array, is that they created an impression of a "virtual" commandkey which didn't interfere with any actual F-key. Either way it was still an untolerable memory corruption.

The [Reachstacker HRS-50](https://forum.rigsofrods.org/resources/reachstacker-hrs-50.374/) was broken due to 2 typos... fixed now.

CuriousMike56 (approved):

tritonas00 (approved):

PR #3059, opened 2023-06-18
Internal - raw pointers begone

ohlidalp:
A collection of refactoring commits, mostly from #3042. No features will be changed.

The purpose is to make the code safer and more pleasant to maintain.


ohlidalp:
Closing because this won't be continued until circa April 2024 and it was premature to open the PR in the first place.

PR #3060, opened 2023-06-19
:angel: script: write/read/delete text files (resources) as string

only-a-ptr:
Minimum work (~ 2.5hrs) and unlimited possibilities. The user can now create, read, detect and delete text files in existing resource locations like "Cache" for the cache folder and "Scripts" for the scripts folder. Access to random files via absolute paths is not possible (game will report "access denied"). To see it in action, open ingame console and say `loadscript demo_script.as` - it showcases the new feature. To test security, say `as game.createTextResourceFromString("boo", "C:\boo.txt", "Cache")` in the console and see what happens.
image

Here is a reprint of the documentation:

```
//The global object 'game':

    /**
    * Checks if the resource file exists in the given group.
    * KNOWN LIMITATION: If existing file is deleted externally, this function will still report it exists until the resource group is reloaded.
    * @see https://ogrecave.github.io/ogre/api/latest/_resource-_management.html
    */
    bool checkResourceExists(const string &in filename, const string &in resource_group);

    /**
    * Deletes a resource from the given group.
    * @see https://ogrecave.github.io/ogre/api/latest/_resource-_management.html
    */
    bool deleteResource(const string &in filename, const string &in resource_group);

    /**
    * Loads a text file resource as string.
    * @see https://ogrecave.github.io/ogre/api/latest/_resource-_management.html
    * @param filename Resource name within the resource group, equivalent to filename without path.
    * @param resource_group Name of resource group to load from
    */
    string loadTextResourceAsString(const string &in filename, const string &in resource_group);

    /**
    * Saves a string as a text file resource.
    * @see https://ogrecave.github.io/ogre/api/latest/_resource-_management.html
    * @param data The file data.
    * @param filename Resource name within the resource group, equivalent to filename without path.
    * @param resource_group Name of resource group to save to. If the group has multiple locations, first writable location is used.
    * @param overwrite By default existing resources are not overwriten.
    * @return True on successful write. False if file not writable or exists and overwrite is disabled.
    */
    bool createTextResourceFromString(const string &in, const string &in filename, const string &in resource_group, bool overwrite=false);
```

My vision is to use this for a script editor with examples, with the option to save files to Documents\My Games\Rigs of Rods\scripts and run them with a click of a button. This branch only delivers the script functions though.

### DIFF
--- a/doc/angelscript/Script2Game/GameScriptClass.h
+++ b/doc/angelscript/Script2Game/GameScriptClass.h
@@ -93,6 +93,39 @@ public:
     */
     bool pushMessage(MsgType type, dictionary@ dict);
     
+    /**
+    * Checks if the resource file exists in the given group.
+    * KNOWN LIMITATION: If existing file is deleted externally, this function will still report it exists until the resource group is reloaded.
+    * @see https://ogrecave.github.io/ogre/api/latest/_resource-_management.html
+    */
+    bool checkResourceExists(const string &in filename, const string &in resource_group);
+
+    /**
+    * Deletes a resource from the given group.
+    * @see https://ogrecave.github.io/ogre/api/latest/_resource-_management.html
+    */
+    bool deleteResource(const string &in filename, const string &in resource_group);
+
+    /**
+    * Loads a text file resource as string.
+    * @see https://ogrecave.github.io/ogre/api/latest/_resource-_management.html
+    * @param filename Resource name within the resource group, equivalent to filename without path.
+    * @param resource_group Name of resource group to load from
+    */
+    std::string loadTextResourceAsString(const string &in filename, const string &in resource_group);
+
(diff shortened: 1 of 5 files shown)


CuriousMike56:
Works fine, tested both Windows and Linux. The demo script assumes you're on Windows though:
https://github.com/RigsOfRods/rigs-of-rods/pull/3060/commits/232f652e0ebb1b04313426f9ba3701d22c4fa54e#diff-eda76373262158e93d51c98a1022f3e36457074c022ce73e772f6373fe01434bR636
Should display the `sys_cache_dir` CVar instead.

only-a-ptr:
@CuriousMike56 Why didn't I think of that? Great point, thanks.

tritonas00 (approved):
Works fine

PR #3063, opened 2023-06-23
🎮 Walkie-talkie label with forwardable commands info & clickable buttons

ohlidalp:
rorbotWalkieTalkie-wide

I played TrainValley recently and this idea came to mind. There are 2 changes:
* When character attempts to forward commands to a sleeping actor (which is typical for on-map machinery), the actor wakes up. Note that forwarding commands from the character was possible before, but the actor had to be already awake.
* When character is in a proximity to forward commands, a wakie-talkie label appears both at the character and on the actor - with a list of commands.

It's glitchy at the moment 
* actor label doesn't disappear or re-appear if asleep - I'll debug.
* cannot be disabled - I'll add a setting.

Walkie talkie icon is from https://openclipart.org/image/800px/325344, scaled and colored by myself in GIMP.

Let me know what you think!


CuriousMike56:
Nice concept! Some quick testing bug reports:
- Labels appear huge for one frame on initial activation, minor visual bug
- Character icon remains after entering any actor (easily reproduced by middle clicking an actor while icon is visible)

Feature suggestions:
- Control any actor's commands by right clicking them, bypassing `importcommands`/`forwardcommands`. This would, for example, allow players to open doors from outside the vehicle.
- Add buttons to actor labels to activate commands using the mouse, would be useful for devices without a physical keyboard e.g. Steam Deck.

only-a-ptr:
walkieTalkieClickable
**Updates:**
* InputEngine: added `setEventSimulatedValue()` allowing simulated input from UI or other programmatic source. This also retroactively enhances the Repair keys UI (#3066)
* GUIUtils: Extended `ImDrawEventHighlighted()` global helper to draw key hints as buttons which simulate input.
* GUIManager: added `RequestStaticMenusBlocking()` so that ad-hoc windows like the WalkieTalkie can block static menus like TopMenubar when hovered by mouse.

ohlidalp:
Superseded by more lightweight #3237.

The branch will be preserved in my fork, so the features developed here (the in-scene marker panels with keyhints and buttons) can be ported later.

PR #3064, opened 2023-06-24
AngelScript - added `getOrientation() -> quaternion` for actors.

only-a-ptr:
image

Changes:
* Added `BeamClass.getPosition() -> vector3`. The existing `getVehiclePosition()` is preserved, but only as DEPRECATED alias.
* Added `BeamClass.getOrientation() -> quaternion`

Demo script updated to showcase the new functions.


only-a-ptr:
I also made some improvements to the ScriptMonitor menu, since I use it every time I'm working on scripts:
- The script is added to 'recent' even if loading fails. This is useful when you're making new script and it fails to start for the first time - no need to retype the filename.
- There's a [remove] button in the recent list, if you're no longer interested in the script - here I worked on a copy of demoscript.

image

tritonas00 (approved):
:+1:

PR #3069, opened 2023-07-09
Fixed huge FPS drop when nodes move far away from vehicle

### LINKED ISSUE #3058: Nodes far away from actor origin result in huge fps drop
As the title says, any nodes that are far away from the actor origin (defined by cameras/cinecam/extcamera?) will eventually cause the frame rate to plummet, rendering the game unplayable. This can be reproduced in a variety of ways, such as flinging detachable parts using the node grabber or by driving the Lotus Esprit with its cinecam nodes that are left behind on spawn. 

Attached is a test vehicle with a node that's very far away:
```
;far away node - kills framerate 
21,  100000,  0.1,  0.0, l 4
```
[faraway-node-fps-drop-test.zip](https://github.com/RigsOfRods/rigs-of-rods/files/11779731/faraway-node-fps-drop-test.zip)

You will likely have to press backspace for the issue to occur as it seems to rarely happen on initial spawn or when pressing the `I` key.

RoR_2023-06-17_14-08-51

**This bug is not present on 2022.12.**

ohlidalp:
Fixed #3058 - caused by 539a1b24f50bad880611f245c76b634dbde1ad84

Solution: instead of considering the whole bounding box when evaluating eventboxes, only consider a bounding box around collision cab nodes. It makes more sense physically and also eliminates surprises, as loose cab nodes would be easily visible.

### DIFF
--- a/source/main/physics/Actor.cpp
+++ b/source/main/physics/Actor.cpp
@@ -1116,11 +1116,17 @@ void Actor::UpdateBoundingBoxes()
     // Reset
     ar_bounding_box = AxisAlignedBox::BOX_NULL;
     ar_predicted_bounding_box = AxisAlignedBox::BOX_NULL;
+    ar_cabnodes_bounding_box = AxisAlignedBox::BOX_NULL;
     for (size_t i = 0; i < ar_collision_bounding_boxes.size(); ++i)
     {
         ar_collision_bounding_boxes[i] = AxisAlignedBox::BOX_NULL;
         ar_predicted_coll_bounding_boxes[i] = AxisAlignedBox::BOX_NULL;
     }
+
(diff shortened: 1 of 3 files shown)


CuriousMike56:
Good news: the main problem is fixed:
RoR_2023-09-03_16-37-27
Bad news: I discovered another scenario. Simply driving far enough away will eventually cause the same issue:
image
It's easier to reproduce on other terrains such as the Community Map.
Oddly enough I haven't been able to reproduce using a plane:
image

ohlidalp:
Oops, I forgot to reset the new cabnode-boundingbox so it grew indefinitely. Fixed.

CuriousMike56 (approved):
All good now 👍

PR #3081, opened 2023-09-03
:angel: Script: Log meaningful error upon callback mixup.

ohlidalp:
One point for @Xploder98 in a friendly Devs vs. Users match :)

Long story short, Xploder mixed up callbacks and the game obligued which resulted in script failure with error asCONTEXT_NOT_PREPARED. This hapenned because our callback system is very confusing:
* we have a legacy "script handler" mechanic in .TOBJ file and classic `game.spawnObject()`, where you pass a function name and it gets invoked on eventbox collision.
* we also have recently introduced `SE_EVENTBOX_ENTER` and `SE_EVENTBOX_EXIT` which work with the classic event callback mechanism `eventCallback()` (and the extended `eventCallbackEx()` version) - it does the same job and even provides some extra data. This works independently of the "script handler" mecahnic described above - you just do `game.registerForEvent(SE_EVENTBOX_ENTER)` and you're covered, for all objects.

**Changes made:**
* clarified the legacy eventbox callback in docs on 'developer.rigsofrods.org' (will update when this commit/PR is merged).
* Added a "this is obsolete" notice to AngelScript.log, see below.
* Added a sensible "did you make a typo in arguments?" notice to AngelScript.log, see below.

**Example AngelScript.log if you mix up the callbacks:**
```
13:03:15: spawnObject(): Specifying event handler function in `game.spawnObject()` is obsolete and only works with terrain scripts; Use `eventCallbackEx()` with event `SE_EVENTBOX_ENTER` instead, it provides the same data and works with any script. Just pass an empty string to the `game.spawnObject()` parameter.
13:03:15: Warning: a callback function with signature 'void eventCallback(int, string, string, int)' was not found but a function with given name exists: 'void eventCallback(int, int)' - did you make a typo in arguments?
```

**What needs to be tested:** 
* Terrain scripts should work normally: check races, spawners and truckshops. 
* Custom scripts like 'demo_script.as' or 'road_editor.as' should work normally.

### DIFF
--- a/doc/angelscript/Game2Script/script_callbacks.h
+++ b/doc/angelscript/Game2Script/script_callbacks.h
@@ -32,7 +32,9 @@ void eventCallback(Script2Game::scriptEvents event, int param);
 */
 void eventCallbackEx(Script2Game::scriptEvents event, int arg1, int arg2ex, int arg3ex, int arg4ex, string arg5ex, string arg6ex, string arg7ex, string arg8ex);
 
-/** Optional; Invoked when a vehicle touches an eventbox which has no custom handler function.
+/** OBSOLETE, ONLY WORKS WITH TERRAIN SCRIPTS - Use `eventCallbackEx()` with event `SE_EVENTBOX_ENTER` instead, it does the same job and works with any script.
+*    Optional; Invoked when a vehicle touches an eventbox which has no custom handler function.
+*    This is a legacy feature which pre-dates generic events `SE_EVENTBOX_ENTER` and `SE_EVENTBOX_EXIT` and bundled objects like truckshop/spawners still rely on it.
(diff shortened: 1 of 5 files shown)


CuriousMike56:
Tested:
Truckshops ✅ 
Races ✅ 
Demo script ✅ 
Road editor ✅ 
Script editor ✅ 

Only new warning I've seen is on the "GPU benchmark" terrain ([Benchmark.zip](https://github.com/RigsOfRods/rigs-of-rods/files/12729321/Benchmark.zip)) where every tree spawned shows:
`[RoR|Script|Notice] Warning: a callback function with signature 'void e(int, string, string, int)' was not found`
The script still works as normal.

PR #3074, opened 2023-07-23
Script editor

ohlidalp:
Script writing is powerful but clumsy:
  1. edit in external editor, save to Documents\My Games\Rigs of Rods\scripts
  2. switch to RoR, either type `loadscript ...` to console or use ScriptMonitor menu to Load/Reload the script
  3. if there's an error, switch to Angelscript.log and figure the error
  4. repeat

To make scripting fun, the dev/test cycle needs to be quick and comfy. So we needed an ingame editor with a "(RE)START" button and error reporting. 

**Behold the 'script_editor.as' AngelScript editor written in AngelScript:**

obrazek

At the moment it's somewhat glitchy but the START/STOP button works perfectly. It relies entirely on `ImGui::InputTextMultiline()`, errors and syntax highlight is done by overdrawing. The numbers are: line number, line length, comment offset, num errors - most are not useful but look cool, you can disable them in the View menu.

Update: yes, the hyperlink opens the default browser (only tested on Win10).
obrazek

Update2: added `game.findResourceFileInfo()` which lists files in pre-defined resource groups. This allowed me to list up files in user's 'scripts' dir:
obrazek

PR #3085, opened 2023-09-10
AI: Per waypoint speed

tritonas00:
- Custom per waypoint speed can be now defined in UI or the json file. Values >= 5 km/h are respected, -1 system default, can be skipped in json.
- Recorder records the speed
- Exporter exports the speed

kk

Code changes:
- Moved vector `ai_waypoints` to `GUI_TopMenubar` so now all `ai_*` vars are in one place and made it a struct
- `std::map<int, float> waypoint_speed` changed type from `int` to `Ogre::String` (way0, way1 ...) so custom speeds are correctly aligned in all repeat times

### DIFF
--- a/resources/scripts/AI.as
+++ b/resources/scripts/AI.as
@@ -1,5 +1,7 @@
 #include "base.as"
 
+array<int> waypoints_speed = game.getWaypointsSpeed();
+
 void main()
 {
     int offset = 0;
@@ -50,7 +52,18 @@ void main()
         }
 
         CurrentTruckai.setActive(true);
-        CurrentTruckai.setValueAtWaypoint("way0",AI_SPEED, game.getAIVehicleSpeed());
+
+        for (uint i = 0; i < waypoints_speed.length(); i++)
+        {
+            if (waypoints_speed[i] >= 5)
+            {
+                CurrentTruckai.setValueAtWaypoint("way"+i, AI_SPEED, waypoints_speed[i]);
+            }
+            else
+            {
+                CurrentTruckai.setValueAtWaypoint("way"+i, AI_SPEED, -1);
+            }
+        }
 
         offset -= game.getAIVehicleDistance();
         translation_x = CurrentTruckai.getTranslation(offset, 0).x;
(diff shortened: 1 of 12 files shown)


ohlidalp:
I skimmed the code and I like it, but I have things on my mind:

- The AI panel now makes good 1/4 of total code in TopMenubar.cpp and it's obviously going to grow, so if you're already migrating data around, it would be good if you made a separate `GUI::VehicleAIPanel` class for it. I don't mean to change the way it's drawn on screen, just the internals. You don't have to do it now, but eventually it'll become necessary.
- I'm pondering how to integrate vehicleAI with the mission system (AI opponents in race, AI traffic, AI forklifts in warehouses...) - The modder will want to pre-configure multiple AI vehicles in the mission file. Waypoints can be loaded from JSON as now (either github or local file in the missionzip). I already coded an angelscript CURL helper [`game.fetchUrlAsStringAsync()`](https://github.com/RigsOfRods/rigs-of-rods/pull/3074/commits/4b202e50316545ccd7d8686357c343dc5da7ee08) so the mission handler can download directly and parsing JSON directly in angelscript is no big deal either. Sorry about bypassing your code this way but as I already blogged, I'm tired of C++ always slowing me down and introducing glitches. The actual AI mode and params must be specified in the .mission file directly though. Then, when actually spawning, the params must be filled to the TopMenubar vars so that the AI.as can read them back. All this is already workable, just not super ellegant.
- An elegant solution to both the points above would be to move the "AI setup" vars from `TopMenubar` directly to `class VehicleAI` and make the AiPanel work directly on top of one `VehicleAI` object, basically allowing to have multiple of those 'in the works' and eventually serving as a part of a mission editor. This doesn't mean replacing the TopMenubar.ai_ variables with existing VehicleAI variables - they can peacefully live there as "//Staging data", waiting for the VehicleAI object to be "activated" anytime, on spawn or when attached to existing vehicle for flexibility. This would also obsolete `game.spawnTruckAI()` which is good because like `game.spawnTruck()` it works outside of message loop, so it's a potential hazzard, and it's already inferior to `game.pushMessage(MSG_SIM_SPAWN_ACTOR_REQUESTED, {...})`. All that needs to be done is to add a new parameter to the message (the dictionary in angelscript and the underlying `ActorSpawnRequest`). Don't worry, I'd code all this, or at least lend a hand if you dare so yourself :)

As I said, I'm not requesting any changes right now, just mind dumping. I'll test and finsh the review later. All in all, you're doing excellent job here. 👍

ohlidalp (approved):

PR #3087, opened 2023-09-13
:truck: new directive `set_attenuation_defaults`

ohlidalp:
Works like all other `set_*_defaults` directives. The arguments are an exact match to the `Ogre::Light::setAttenuation()`, see https://ogrecave.github.io/ogre/api/latest/class_ogre_1_1_light.html#a43f763d809bc7da9a85fec15f57380f4

**Only affects flares3 for the moment.**

Parameters:
* **range**	The absolute upper range of the light in world units.
* **constant**	The constant factor in the attenuation formula: 1.0 means never attenuate, 0.0 is complete attenuation.
* **linear**	The linear factor in the attenuation formula: 1 means attenuate evenly over the distance.
* **quadratic**	The quadratic factor in the attenuation formula: adds a curvature to the attenuation formula.

I thoroughly verified using debugger that the values do propagate to the OGRE light source and stay there. I wasn't able to see any effect though. Let's test under OGRE 14.

My testing vehicle:
[flare3attenuation-agoras.zip](https://github.com/RigsOfRods/rigs-of-rods/files/12592003/flare3attenuation-agoras.zip)

### DIFF
--- a/source/main/gfx/GfxData.h
+++ b/source/main/gfx/GfxData.h
@@ -117,6 +117,12 @@ static const int CAMERA_MODE_ALWAYS_HIDDEN = -3;
 static const int CAMERA_MODE_ALWAYS_VISIBLE = -2;
 static const int CAMERA_MODE_3RDPERSON_ONLY = -1;
 
+// Light attenuation defaults - matches defaults in `Ogre::Light::Light()` constructor.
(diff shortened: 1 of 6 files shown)


ohlidalp:
Closing until migration to OGRE14, then this becomes relevant.

PR #3089, opened 2023-09-28
:bug: Fixed build failing when building RelWithDebInfo && 🔧 Move conan generated files to a subfolder

AnotherFoxGuy:
Now RoR is buildable on Windows in the RelWithDebInfo config without needing to build the deps in RelWithDebInfo
Also moved the files generated by Conan to the `generators` folder so it no longer clutters up the build folder


ohlidalp (approved):
Tested on Win10 x64 + VisualStudio 2022
* generated files are now in the "BUILDDIR/generators" subdirectory as advertised 💌 
* RelWithDebInfo seems to work perfectly, other configurations are unchanged 👍

PR #3090, opened 2023-09-28
Survey map improvements

tritonas00:
- Made mini map circular
- Cleaner layout by replacing text from terrain icons and waypoints with hovering tooltips

screenshot_2023-09-28_22-12-52_1
screenshot_2023-09-28_22-13-06_1

Known issues:
- ~~Some terrain icons are rendered multiple times it seems (old issue, nhelens can be use for testing)~~
- ~~It would be nice to use the terrain texture and zoom into that instead of using a camera which reduces performance and causes other issues (terrain water moving when zoom in/out the mini map, blank rendering in some cases - again all those are old issues)~~


ohlidalp:
I made an inventory of our icons: Actors and Characters are OK, just places of interest are a weird mixture.
obrazek

ohlidalp:
I ended up doing pixel art... not good not terrible, I think it's a bit of an improvement.

obrazek
obrazek

ohlidalp:
I added missing icons "ship" and "sign" and replaced "racestart" because it looked like an active actor instead of place of interest.
obrazek
I also unrotated all static icons on map and fixed machine icon when inactive.
obrazek

PR #3092, opened 2023-10-02
SurveyMap now zooms to static texture

ohlidalp:
Changes:
* Texture "MapRttTex-#" is now short-lived, after creating the map image it's exported as static image and destroyed
* Map texture size is now hardcoded to 4096x4096 (the maximum value guaranteed to be portable).
* Minor fix: The minimap now resets zoom and display mode when loading new terrain.

There are issues though:
* Even after hardcoding the texture size to 4096x4096 (the maximum value guaranteed to be portable), the resolution doesn't look spectacular on the zoomed in map, even with small terrain like Auriga.
* The VRAM usage is rather high
* [FIXED] ~~On both DirectX9 and GL, I'm getting a strange fade-to-black effect where the texture renders all black until zoomed in. Even when inputting exactly uv0=(0,0) and uv1=(1,1) which is the same as imgui defaults, it renders black. When left without the arguments, it renders OK.  I already encountered this once when trying to make the surveymap resizable - smaller sizes blackened the texture. I need to look closer at this.~~

obrazek

### DIFF
--- a/source/main/gfx/SurveyMapTextureCreator.cpp
+++ b/source/main/gfx/SurveyMapTextureCreator.cpp
@@ -43,6 +43,8 @@ SurveyMapTextureCreator::SurveyMapTextureCreator(Ogre::Real terrain_height) :
 
 SurveyMapTextureCreator::~SurveyMapTextureCreator()
 {
+    if (mRttTex)
+        mRttTex->removeAllViewports();
     if (mCamera)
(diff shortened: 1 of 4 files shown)


ohlidalp:
Apparently the fade-to-black effect is related to RTT, because when I hack the surveymap to use ordinary texture file, it works just fine :)
screenshot_2023-10-03_00-17-35_1

ohlidalp:
The change is now included in #3090

PR #3093, opened 2023-10-11
Restored + colorized net. player names on survey map.

ohlidalp:
#3090 introduced mouse tooltips for terrain icons which accidentally also prevented player names from showing.

Also fixed green/blue coloring of person icon to match vehicle icons.

Screenshot from official NeoQueretaro:
obrazek


CuriousMike56:
👍

PR #3101, opened 2023-11-05
Fixed special box text position

tritonas00:
`special_text_centering_weight` added in https://github.com/RigsOfRods/rigs-of-rods/commit/83c1d273ef6dd5e3e057961191dba2bda8e670cf is only for centering live/quick repair special box text

alternatively it could be
```
ImGui::SetCursorPosX(ImGui::GetCursorPosX() + ((content_width / 2) - (text_w / 2)) * special_text_centering_weight);
```
and update `special_text_centering_weight` accordingly -> CBA


ohlidalp (approved):
Just right.

PR #3108, opened 2023-12-03
Added option to invert orbit camera controls

CuriousMike56:
This adds an "Invert orbit camera" (CVar `io_invert_orbitcam`)  option to Controls settings, inverting mouse and numpad controls. 
image

In addition, mouse cursor now hides when moving the camera. Also wanted to make the cursor auto hide after a few seconds of inactivity like in previous RoR versions, which apparently broke with the switch to the DearIMGUI-based cursor. Code for this seems to exist but doesn't appear to work correctly: https://github.com/RigsOfRods/rigs-of-rods/blob/26411c1a86ae807a3c6eef504259276baee26f22/source/main/gui/GUIManager.cpp#L294-L300
Moving this code to `CameraManager` causes the mouse to become permanently hidden. 

Old code:
https://github.com/ohlidalp/ror-legacy-svn-trunk/blob/0a55a993899d6a06cdc58fd98b5b675b371eb89a/source/main/gui/GUIManager.cpp#L142C1-L147C3

### DIFF
--- a/source/main/Application.cpp
+++ b/source/main/Application.cpp
@@ -200,6 +200,7 @@ CVar* io_outgauge_port;
 CVar* io_outgauge_delay;
 CVar* io_outgauge_id;
 CVar* io_discord_rpc;
+CVar* io_invert_orbitcam;
               
 // Audio
 CVar* audio_master_volume;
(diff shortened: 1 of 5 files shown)


ohlidalp (approved):
Very nice touch.
I'll fix the cursor-hiding some other time.

PR #3120, opened 2024-02-03
✨ Add option for Keyboard Steering Sensitivity

cryham:
So I added a new var and a slider on GUI for Keyboard Steering Sensitivity,
which is simply how fast you steer wheels with keyboard left/right. Default 1.0 is like it was before.
Depends on vehicle actually. Some had slow steering so 2.0 is fine for me now, some had okay and it's not needed above 1.2 etc.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,4 @@ tools/l10n/languages/
 tools/l10n/ror\.pot
 CMakeUserPresets.json
 tools/l10n/TMP.pot
+.cache
--- a/source/main/Application.cpp
+++ b/source/main/Application.cpp
@@ -196,6 +196,7 @@ CVar* io_ffb_stress_gain;
 CVar* io_input_grab_mode;
 CVar* io_arcade_controls;
 CVar* io_hydro_coupling;
+CVar* io_hydro_sensivity;
 CVar* io_outgauge_mode;
 CVar* io_outgauge_ip;
 CVar* io_outgauge_port;
--- a/source/main/Application.h
+++ b/source/main/Application.h
@@ -408,6 +408,7 @@ extern CVar* io_ffb_stress_gain;
 extern CVar* io_input_grab_mode;
 extern CVar* io_arcade_controls;
 extern CVar* io_hydro_coupling;
+extern CVar* io_hydro_sensivity;
 extern CVar* io_outgauge_mode;
 extern CVar* io_outgauge_ip;
 extern CVar* io_outgauge_port;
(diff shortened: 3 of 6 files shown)


ohlidalp (approved):
Nice work.

Our steering actuators 'hydros' have user-defined extension lengths, so naturally each vehicle handles differently. I don't even know if the extension speed depends on length or is fixed. How does VDrift/StuntRally do this?

cryham:
Well, VDrift/StuntRally just have classes that set wheel steer angle but nothing is physical. A setup file .car has max steering angle and all key inputs have sensivity etc. So it just sets angle from input directly.
And a wheel is only casting 1 ray down to get contact, this is really bad for big wheeled vehicles in SR.
But we have Pacejka Magic Formula with many (10 to 30) coefficients for tire simulation (SR has even an ingame editor for this).
I see RoR has some adhesion and other coefficents in FrictionSettings (we don't have that) like for all contacts, but no special curves for tires right?

CuriousMike56:
> Right. All our contacts are evaluated as node vs. surface, where surface is in charge of all the properties, node is dimensionless and frictionless point only. Tires are no exception, they are rolling frames and they are not built free-form but generated by `*wheel*` sections in our .truck defs, so they just have 2 rings of nodes (not optimal IMO) and user-adjustable number of radial segments. So segment count is the only way modder can affect grip.

Most vehicles use [set_node_defaults](https://docs.rigsofrods.org/vehicle-creation/fileformat-truck/#set_node_defaults) to change wheel friction.

cryham:
Okay thanks for info.
So what is the process for merging (this and other approved) PRs to, does it go into further planning or roadmap etc?

BTW, could an administrator (IDK who) approve one of my accounts for RoR Forum, which I tried recently: CrystalHammer, Crystal Hammer or cryham? I did email contact, only email I found 6 days ago. Sorry IDK how else to contact, I'm not on discord.

CuriousMike56:
Works well, though I'd recommend moving the option under vehicle control options and adding a reset button.
RoR_2024-02-06_15-01-07

cryham:
Screenshot_20240207_234555s
Is this okay?

CuriousMike56:
Much better 👍

ohlidalp:
Sorry about the delay on my end, I'll catch up on weekend.

Btw the button is a nice touch

ohlidalp (approved):
Tested, works great. I took the liberty to do a minor UI touch - I made the Reset button smaller and inline.
obrazek

ohlidalp:
There's a bug - when using controller, the smoothing only affects centering, not the actual steering: https://discord.com/channels/136544456244461568/189904947649708032/1216845086956982385

CuriousMike56:
https://github.com/RigsOfRods/rigs-of-rods/assets/46073351/ae61c1a3-8489-440d-ae23-bc3a34624b35

Only happens with values higher than 1.0. At 5.0 it instantly snaps to center when the thumb stick is released. Assuming there's a conflict with the existing analog sensitivity/smoothing settings:
image

CuriousMike56:
Yes, ideally the slider shouldn't affect analog input.

cryham:
Solution would something like:
```
            // return rate
            float dirdelta = analog ? PHYSICS_DT : App::io_hydro_sensivity->getFloat() * PHYSICS_DT;
```
IDK how to set analog though. It should be true when using analog game controller, and false for just keyboard.

PR #3124, opened 2024-02-10
Ogre 14 fixing

cryham:
Some fixes I did recently.
Needed that `std::` to build.
added fontdef escapes with "\:" etc to fix warnings
Terrain works now.

Shadows work now. Wow it's the first time I see shadows in RoR on my Debian+AMDgpu setup :grin: .
Didn't at all test or build with Caelum, SkyX etc. Just default skybox.
I'm building w/o PagedGeom.
I saw PagedGeom was crashing, it needed custom wind shader or something.
I don't have CG (and likely won't) so some vehicles are just white.
Water is solid gray.
WIP


ohlidalp:
*`*Jaw dropped*`*

The white vehicles are because of 'nicemetal.cg' shader which is automatically used by our 'managedmaterials' material generator feature of the .truck def. I think it could be replaced by RTSS with [metal_roughness](https://ogrecave.github.io/ogre/api/latest/rtss.html#autotoc_md255) but I haven't tried yet.

Honestly I find PagedGeom to be a suspect altogether. I went through the source - there is [mesh-merging code](https://github.com/OGRECave/ogre-pagedgeometry/blob/master/source/BatchedGeometry.cpp#L537) but no texture-merging code (could be done by `HardwarePixelBuffer::blit*()`). Ogre's StaticGeometry also merges meshes and it also cannot merge textures. Both solutions rely on the modder to use only meshes sharing the same material. That's never going to fly with our community and our ODEF system. Finally, there's a lot of code to [reflect cam movement](https://github.com/OGRECave/ogre-pagedgeometry/blob/master/source/PagedGeometry.cpp#L454). I found no Cg shaders in the media folders but it's from the 2000's so anything is a suspect.

Anyway it's fantastic that you're giving this a try, thanks.

paroj:
> I saw PagedGeom was crashing, it needed custom wind shader or something.

I have ported PagedGeom to Ogre UnifiedShader a while ago:
https://github.com/OGRECave/ogre-pagedgeometry/commit/d5b5d6ce13a5ca7db77e9e5f31df7d4c83210168

maybe the RoR version does not include that commit and/ or does not include the required shader files.

> Ogre's StaticGeometry also merges meshes and it also cannot merge textures.

PagedGeom is basically a more sophisticated StaticGeometry. It includes imposters and shaders for wind and grass. Both system operate on a grid batches - i.e. you dont batch the whole world, but do some subdivisions to allow camera culling.

The textures are not merged as you must change the uv coordinates, matching the texture atlas that you create. Doable but probably was not a priority. Also this still does not cover different diffuse/ emissive values of the materials.

Issue #3113, opened 2024-01-01
Add input sources for cruise control for the dashboard system


MarkROR:
RoR has a cruise control feature - toggled on/off with the spacebar. However, there is no way of knowing if you have successfully activated it (some vehicles won't let you activate it, unless you've reached a certain speed) or if you've accidentally disabled it by e.g. braking slightly.

Having a visual indicator (a dashboard light) would be very helpful, I think.

We could provide an input source for **making this possible in the future**:
`cruisecontrol` or `cruisecontrol_active` _true if cruise control is on_

While we're here, we could also provide the value it has been set to:
`cruisecontrol_kph` _cruise control set speed value, in kilometers per hour_
`cruisecontrol_mph` _cruise control set speed value, in miles per hour_
These two would be useful, because they'd show you the set value instantly while you're adjusting the set speed with CTRL+R/CTRL+F

ohlidalp:
Hi, you're right, those are missing. I'll go for `cruisecontrol_active`.

Also it looks like the dash system still lives in the past regarding to lights. So I'll also add `flare_type_XN`, where X is the flare type flag and N is 'u' flare number 1-10.

ohlidalp:
Also consider difflock and xWD https://discord.com/channels/136544456244461568/189904947649708032/1205431235392970832

PR #3125, opened 2024-02-10
Added dashboard indicator lights for new light types

MarkROR:
New: `sidelights/headlights/highbeams/foglights` and `cruisecontrol_active`. These were introduced in #2937.

`cruisecontrol_active` will be introduced in the future #3113 and when more (difflock, 4wd, etc) get added, we may have to think about a new dash/gui/hud or a dedicated panel for these lights

The reason the new lights are smaller than the old ones is so RoR doesn't have to resize them - they have been exported at the exact resolution they will be displayed on the screen, no resizing = best quality.


CuriousMike56:
RoR_2024-03-17_14-01-45
Looks good 👍

PR #3127, opened 2024-02-17
Added option to enable window resizing

CuriousMike56:
Adds `Allow game window resizing` to diagnostic settings. Useful for taking ultrawide or 4:3 screenshots.


ohlidalp:
Please test all HUDs, especially old overlay based. IIRC they were the reason we disabled the resizing once

CuriousMike56:
Yes GUIs don't react correctly to resizing, always hidden in screenshots anyway. Its buried in diagnostic settings for this reason.

This setting also fixes a strange issue for me with Windows Auto HDR on my monitor, spawning a vehicle with videocameras on a map with lots of trees causes the refresh rate to plummet. FPS remains the same.

PR #3129, opened 2024-02-19
New mod type: asset pack (.assetpack)

ohlidalp:
Quick 2hr draft - ~BUILDS, NOT TESTED.~ Tested, added example mods.

A new file type "*.assetpack" was added to modcache with all the usual treatment:
- the file is recorded to the cache file (/cache/cache.json),
- it's searchable and selectable via UI (`LT_AssetPack`)
- can have a thumbnail (filename format '{}-mini.*'' where {} is the assetpack file name without extension)
- The file itself can be empty, recognized entries are `assetpack_name` (display name), `assetpack_description` (for display in selector UI) and `assetpack_author` (same syntax as in  .truck file).
- When requested, the game loads the entire containing ZIP/directory (called 'bundle' by jargon).

Assetpacks are referenced by the file name (standard approach) and can be used for both terrains and actors:
- Terrains: in .terrn2 file, add section `[AssetPacks]` followed by list of filenames (with extension, each on separate line) ending with `=` token, same as in `[Objects/Scripts]` sections - this is a quirk of the game.
- Actors: in the .truck file, use section `assetpacks` followed by list of filenames (with extension, each on separate line)

Note that, unlike any other mod, assetpacks aren't loaded to their own resource group but rather to the resoruce group of the mod which requested them. Therefore try to name all resources in an unique way to avoid clashes.

**Test mods** - a wacky mini terrain I created for future tests, and one mesh (with multiple .ODEFs) I extracted from it:
* testpack-crazycones.assetpack: [testpack-crazycones.zip](https://github.com/RigsOfRods/rigs-of-rods/files/14338275/testpack-crazycones.zip)
* terrnbatcher_testmap.terrn2: [terrnbatcher_testmap.zip](https://github.com/RigsOfRods/rigs-of-rods/files/14338281/terrnbatcher_testmap.zip)

### DIFF
--- a/source/main/Application.h
+++ b/source/main/Application.h
@@ -261,7 +261,7 @@ enum VisibilityMasks
     HIDE_MIRROR       = BITMASK(3),
 };
 
-enum LoaderType //!< Operation mode for GUI::MainSelector
+enum LoaderType //!< Search mode for `ModCache::Query()` & Operation mode for `GUI::MainSelector`
 {
     LT_None,
     LT_Terrain,   // Invocable from GUI; No script alias, used in main menu
@@ -278,6 +278,7 @@ enum LoaderType //!< Operation mode for GUI::MainSelector
     LT_AllBeam,   // Invocable from GUI; Script "all",  ext: truck car boat airplane train load
     LT_AddonPart, // No script alias, invoked manually, ext: addonpart
     LT_Tuneup,    // No script alias, invoked manually, ext: tuneup
+    LT_AssetPack, // No script alias, invoked manually, ext: assetpack
 };
 
 // ------------------------------------------------------------------------------------------------
--- a/source/main/GameContext.cpp
+++ b/source/main/GameContext.cpp
@@ -126,7 +126,7 @@ bool GameContext::LoadTerrain(std::string const& filename_part)
     // Init resources
     App::GetCacheSystem()->LoadResource(terrn_entry);
 
-    // Load the terrain
+    // Load the terrain def file
     Terrn2Def terrn2;
     std::string const& filename = terrn_entry->fname;
     try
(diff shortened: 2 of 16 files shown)


cryham:
Great.

> Note that, unlike any other mod, assetpacks aren't loaded to their own resource group but rather to the resoruce group of the mod which requested them. Therefore try to name all resources in an unique way to avoid clashes.

One thing I have in mind is version. What if I someday decide to make a newer version of e.g. SR_Terrain asset pack? E.g. fixing some texures or adding some more.
IMO version should also be a part of a pack identifier. But with logic that would let us replace older version of same asset pack.
Do you agree or do you have a different approach to such case?

ohlidalp:
Assetpacks are referenced exclusively by filename of the .assetpack, same as all other mods in the game. For a different version or any customized edition, just use new filename. Anything more would be excess complexity.

My goal is to use explicit RGs everywhere but former OGRE versions defaulted to searching everything and former RoR Devs didn't care. IIRC paged is hardcoded to use the default RG. 

Yes, resources must be named to avoid clashes with any prepacked media.

cryham:
Allright. So e.g. I'll have a SR_objects_v1 pack and in future will do a SR_objects_v2 and users will just get that instead and disable the v1. Sounds fair.

ohlidalp:
@cryham Have you tested it yet? You can download the github-actions devbuild below.

cryham:
Not yet, I'll test later today.

cryham:
Allright this works well.

CuriousMike56:
Tested demo files, works fine. No issues with the file format 👍 

> Also how will this work for users? E.g. someone gets from repo in game a track Grc2-Coast, then will RoR auto download all needed packs for it if not present?

Best solution IMO would be to add a "required files/dependencies" prompt when downloading from the in-game repo. Not sure how this would work on the website side.

PR #3131, opened 2024-02-22
Add rot_yxz for SR road rotations Y then X, Z

cryham:
This fixes for me road roll angles.
Like you said, they were bad. Yaw was only working ok.
Before:

After:

Track has this road TestC1-Circle-road.tobj, it has a circle with all points: -16, yaw, 0
[TestC1-Circle-road.txt](https://github.com/RigsOfRods/rigs-of-rods/files/14377134/TestC1-Circle-road.txt)
This would be important for SR tracks roads, mainly for better bridge entries.

### DIFF
--- a/source/main/resources/tobj_fileformat/TObjFileFormat.cpp
+++ b/source/main/resources/tobj_fileformat/TObjFileFormat.cpp
@@ -62,8 +62,9 @@ void TObjParser::Prepare()
     m_in_procedural_road = false;
     m_cur_procedural_obj = new ProceduralObject();
     m_cur_procedural_obj_start_line = -1;
-    m_road2_num_blocks = 0;
+    m_road2_num_blocks   = 0;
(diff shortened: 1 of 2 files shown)


CuriousMike56:
Please check other maps with default roads (North St Helens, Small Island, etc)

cryham:
Agh. Okay so this isn't so simple.
Since this is for all .tobj, and there were x and z rotations != 0 there, it goes somewhat bad there.
So the solution would be to leave it as it was and add this as a different option for SR roads only.

cryham:
Okay possibly like this. I'll check it tomorrow.

cryham:
Right. It works as before now.
And with `rot_yxz` put in .tobj it works new way for SR.

ohlidalp (approved):
I like the approach and the minimal code change. I'll merge after Mike confirms everything works OK. Sorry about taking long to respond.

PR #3136, opened 2024-03-04
New top menu 'Styling' - change colors and other parameters

ohlidalp:
An experiment: a new menu for adjusting fragment shader params from 'nicemetal.cg' in ManagedMaterials.

**Thoughts:**
* min/max values for all params are hardcoded to <-1.0  1.0>.
* param 'color' is recognized as special, all others are considered `float`.
* only managed materials are listed.
* only shader params are supported, although OGRE material seetings are also interesting.
* This menu could become a '.skin' editor, similarly how the Tuning menu is a '.tuneup' editor.

obrazek


ohlidalp:
Closing for now, to be reopened later if there's interest

PR #3139, opened 2024-03-10
Danmackey fov rework

ohlidalp:
With this ticket I try to clear some confusion about adjusting FOV and the meaning of the slider(s) in Top Menubar UI.

The slider is actually context sensitive; We have multiple presets:
 * `gfx_static_cam_fov_exp` ~`SliderFloat()` ~ A "zoom factor" of static camera (the actual FOV is dynamic based on distance); Adjust by 'Ctrl+mouse wheel' or TopMenubar/Settings.
 * `gfx_fov_internal` ~ `SliderInt()` ~ FOV of cinecam; Adjust by hotkeys EV_COMMON_FOV_{LESS/MORE/RESET}.
 * `gfx_fov_external` ~ `SliderInt()` ~ FOV of exterior cameras (3rd person, free cam, freefixed cam), adjustable by hotkeys. EV_COMMON_FOV_{LESS/MORE/RESET}. 

To start this off, I clarified and commented the code in Top Menubar (as well as related CameraManager code), and I added a combobox to select camera mode:
obrazek

Note: having FOV stepped in whole numbers is sort of a convention; thing is, angle degrees aren't typical (base 10) numbers but base 60 numbers, meaning 1 Degree = 60 Angular Minutes. Using a `float` to represent FOV in degrees would be kind of bastardized, but I think I'll do it anyway.

### DIFF
--- a/source/main/gfx/camera/CameraManager.cpp
+++ b/source/main/gfx/camera/CameraManager.cpp
@@ -1365,3 +1365,20 @@ void CameraManager::CameraBehaviorVehicleSplineUpdateSplineDisplay()
     }
     m_splinecam_mo->end();
 }
+
+std::string ToLocalizedString(CameraManager::CameraBehaviors behavior)
+{
+    switch (behavior)
+    {
+    case CameraManager::CAMERA_BEHAVIOR_CHARACTER:       return _LC("CameraBehavior", "Character");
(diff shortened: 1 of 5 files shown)


ohlidalp:
I analyzed how the smoothing works and added a slider for it. I didn't add smoothing anywhere, but all existing smoothed values were unified around this slider. Default value is 4.0 which is the same as before (although the code was finnicky so it may not be always in effect).

obrazek

Issue #3137, opened 2024-03-05
Waves on water is broken


Miner34dev:
**Hardware**: AMD Ryzen 3 5300U, integrated graphics
**Software**: Garuda Linux, latest RoR dev build, OpenGL render (of course)
**Settings**: [RoRcfg.txt](https://github.com/RigsOfRods/rigs-of-rods/files/14495442/RoRcfg.txt)

HydraX with waves on water glitches out badly (but only if you have a vehicle spawned, else it works as normal):
![Screenshot_20240305_132508](https://github.com/RigsOfRods/rigs-of-rods/assets/134534394/9a7cae40-5600-43d8-87aa-8a5a713ab785)
HydraX without waves on water somehow seems to have waves actually:
![Screenshot_20240305_133006](https://github.com/RigsOfRods/rigs-of-rods/assets/134534394/c026bef9-50ed-4e73-b7a7-402e6a81d8f5)
Reflection + refraction with waves on water doesn't have waves at all:
![Screenshot_20240305_132818](https://github.com/RigsOfRods/rigs-of-rods/assets/134534394/fec73a37-c804-4cb1-aeac-34e9383a5c36)

ohlidalp:
Hello.

The 'waves on water' is somewhat arcane feature, I don't think anybody uses it because it hurts FPS. It works by creating a grid mesh and shaping it dynamically to simulate a wavy sea, both physically and visually.

Hydrax comes with fake caustics, that's what you're seeing.

ohlidalp:
I found a crash scenario (Win10, Directx9, PSSM off, configuration=RelWithDebInfo) - OGRE crashes when rendering a frame:
![obrazek](https://github.com/RigsOfRods/rigs-of-rods/assets/491088/59a9310f-38b3-4345-ad3a-32c68ad21737)
Steps to reproduce:
1. In main menu, go to settings, set water=basic and Waves=on 
2. Load a map with water, like 'Small Island' (the one on screenshots above).
3. Slightly move the 'Wave height' slider in 'Top menubar/Settings' menu.
4. Go back to main menu, set water=Hydrax and resume game
5. Go back to main menu again -> crash.

I'd try under Debug but there's a showstopper - the terrain calculates lightmap on background thread, and you can't return to menu until that finished. It takes ~15sec on RelWithDebInfo. On Debug it would take at least 5min. I need to code an option to force-disable lightmap calc to test.

ohlidalp:
[...]

PR #3142, opened 2024-03-24
Added water options to 'TopMenubar/Settings' menu.

ohlidalp:
I tried to investigate #3137 but got nothing. So I at least analyzed the water code and added UI for tweaking HydraX water.

Both basic water with waves and HydraX water boil down to the same thing: a dynamic grid-mesh updated in code. The only major difference is how the waves are calculated: Basic uses configurable "wave trains", HydraX uses Perlin noise with adjustable params (via .hdx config file). Since both systems are in our codebase, it would make sense to unify them. To better understand how the system operates, I added the UI.

obrazek


ohlidalp:
I got curious if I could change water type in-game by simply destroying the water subsystem and initing it again. Well, it works
obrazek
Also waves are now toggleable there, that was already possible before just the UI was missing.

ohlidalp:
I've bumped this PR with one small fix. I couldn't fix the glitch but I got closer, see https://github.com/RigsOfRods/rigs-of-rods/issues/3137#issuecomment-2198713608

CuriousMike56:
https://github.com/user-attachments/assets/bd206bed-b3f2-4e6e-8fd3-19216269e361

Switching between reflection water types often results in white, and Hydrax always crashes:

RoR_2024-08-06_01-58-00

The options remain shown when on a terrain that doesn't feature water (but doesn't do anything besides crash without error if Hydrax is selected)

Real time waves adjustment is fun though 😄

ohlidalp:
Looks like something isn't cleaned up properly and then duplicates and breaks. Might also be Cg or Ogre bug.

ohlidalp:
@CuriousMike56 I tried only under Dx9 and didn't see the white surfaces... but I got a lot of other glitches/crashes. I fixed switching between all modes except Hydrax which still crashes when toggled the 2nd time.

CuriousMike56:
Looks like the white surfaces are gone now. 

Hydrax options seem to do nothing:

https://github.com/user-attachments/assets/4f2e6275-b3b8-402b-8f35-e736c2513195

Switching to Hydrax results in visual glitches:

https://github.com/user-attachments/assets/27d445fe-9592-49ed-aef7-bf77e4a9bd13

And finally, basic water visuals freeze if waves are disabled:

https://github.com/user-attachments/assets/a795d4d6-4f82-4dee-963f-f3f68356bd7d

ohlidalp:
I fixed HydraX crashing on nd activation, but now it seems to hang when returning to menu >:/

CuriousMike56:
It may be best to just not include Hydrax for now (or hide it behind a setting), its visuals are still busted for me:
RoR_2024-09-19_03-25-58

ohlidalp:
Done. HydraX excluded from the top menubar dropdown. I'll continue fixing it after OGRE13+ port.
obrazek
obrazek

CuriousMike56:
- The options appear on terrains that don't feature water, changing mode there does nothing (would be fun to enable water on these maps though)
image
- Returning to menu crashes with Hydrax enabled
- Hydrax sliders have too much range IMO, it's easy to create horrors and tank your frame rate:
RoR_2024-09-22_21-50-00
Grid complexity should max at 500, and strength/elevation at 25.

ohlidalp:
Closing until OGRE14 port... too much garbage. There are 2 outstanding issues:
*  load map with water, return to menu and load a different map also with water, game hangs/crashes: https://discord.com/channels/136544456244461568/189904947649708032/1289036836139958352
* returning to menu with hydrax enabled still crashes https://discord.com/channels/136544456244461568/189904947649708032/1289034862056964097

PR #3144, opened 2024-03-27
Exception hardening

ohlidalp:
This should put a stop to the game panicking on any unhandled exceptions, which are usually trivially recoverable, like this dialog which was caused by a .mesh file missing from a terrain:
obrazek

All exceptions which happen during processing message queue are now automatically logged to RoR.log and also forwarded to AngelScript as game event - a trick already widely used in OGRE bindings, see `SE_GENERIC_EXCEPTION_CAUGHT`.


ohlidalp:
Merging in good faith without peer approval - I've been developing other branches on top of this for about a week, and didn't observe any issues, in fact I added a few more error tolerance commits.

Issue #3105, opened 2023-11-18
Paused actor physics resume when switching to another actor


CuriousMike56:
- Spawn two actors
- Press `END` key to pause physics for current actor
- Switch to the other actor, previous actor's physics will resume

https://github.com/RigsOfRods/rigs-of-rods/assets/46073351/bc072281-e17c-45ce-9133-5c57becf54e8

Also recently noticed the skeleton debug view disables itself when switching actors, possibly related?

ohlidalp:
Here is the culprit (and yes, skeletonview related): https://github.com/RigsOfRods/rigs-of-rods/blame/master/source/main/physics/ActorManager.cpp#L745-L746 - it broke in https://github.com/RigsOfRods/rigs-of-rods/commit/9431e1b508575c40498ed9becd8a0385a3b95463 - I needed to reset the paused/skeleton after decoupling from the player vehicle. Doesn't make big sense for paused state but it's important with skeleton view.

PR #3146, opened 2024-03-31
Fixed paused actor physics resuming after switching to another actor

ohlidalp:
Fixes https://github.com/RigsOfRods/rigs-of-rods/issues/3105.

This was quite an endeavor - it turned out there's no single spot to correctly reset the physics pausd state (or skeletonview state - related). While researching how to add it, I realized scripts may want to know when (un)linking of actors happens, so I added a script event `SE_GENERIC_TRUCK_LINKING_CHANGED` for it.

**What changed:** code changed signifficantly; all data remained the same - code handling hooks/ties/ropes/slidenodes got only cosmetic edits. There are many new sanity/integrity checks in the (un)linking logic which will trigger Asserts on Debug and just prevent damage on Release.

**What to test:** Operation of ties/hooks/ropes/slidenodes, including savegame and abrupt deleting of one of the actors.

### DIFF
--- a/doc/angelscript/Script2Game/globals.h
+++ b/doc/angelscript/Script2Game/globals.h
@@ -75,7 +75,9 @@ void print(const string message);
 
  	SE_GENERIC_MESSAGEBOX_CLICK        //!< triggered when the user clicks on a message box button, the argument refers to the button pressed
     SE_GENERIC_EXCEPTION_CAUGHT        //!< Triggered when C++ exception (usually Ogre::Exception) is thrown; #1 ScriptUnitID, #5 originFuncName, #6 type, #7 message.
-    SE_GENERIC_MODCACHE_ACTIVITY       //!< Triggered when status of modcache changes, args: #1 type, #2 entry number, for other args see `RoR::modCacheActivityType`    
+    SE_GENERIC_MODCACHE_ACTIVITY       //!< Triggered when status of modcache changes, args: #1 type, #2 entry number, for other args see `RoR::modCacheActivityType`  
+
+    SE_GENERIC_TRUCK_LINKING_CHANGED   //!< Triggered when 2 actors become linked or unlinked via ties/hooks/ropes/slidenodes; args: #1 state (1=linked, 0=unlinked), #2 action `ActorLinkingRequestType` #3 master ActorInstanceID_t, #4 slave ActorInstanceID_t
 
  	SE_ALL_EVENTS                      = 0xffffffff,
     SE_NO_EVENTS                       = 0
--- a/source/main/GameContext.cpp
+++ b/source/main/GameContext.cpp
@@ -1367,21 +1367,20 @@ void GameContext::UpdateCommonInputEvents(float dt)
     {
         //m_player_actor->hookToggle(-1, HOOK_TOGGLE, -1);
         ActorLinkingRequest* hook_rq = new ActorLinkingRequest();
-        hook_rq->alr_type = ActorLinkingRequestType::HOOK_ACTION;
+        hook_rq->alr_type = ActorLinkingRequestType::HOOK_TOGGLE;
         hook_rq->alr_actor_instance_id = m_player_actor->ar_instance_id;
-        hook_rq->alr_hook_action = HOOK_TOGGLE;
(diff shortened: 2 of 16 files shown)


CuriousMike56:
Almost perfect. Found one crash scenario:
- Spawn [Liebherr LTM 1050-3.1](https://forum.rigsofrods.org/resources/liebherr-ltm-1050-3-1.75/) (any crane will do) and any load, such as the [Rock pack](https://forum.rigsofrods.org/resources/rock-pack.126/)
- Press O to attach ties to the load
- Delete the load, disconnecting ties first doesn't matter. 
- Spawn a copy of the load with `CTRL+PERIOD`, or spawn a different load
- Try to attach ties to other load -> crash

Under Debug I get infinite asserts when trying to backspace with the crane:

https://github.com/RigsOfRods/rigs-of-rods/assets/46073351/73e0e22a-964d-4381-bb63-7e68dab1718a

And clicking retry gives:
devenv_2024-04-22_19-22-55

ohlidalp:
I fixed the asserts on backspacing the Liebherr crane, and added many more asserts in the process, because those that triggered were _completely_ valid, `SyncReset()` was actually forcing hooks/ropes/ties to inconsistent state.

ohlidalp:
obrazek
I pressed L to attach the hook, then tried to delete via top menu.

CuriousMike56:
Attached loads now move faster than the crane when using live repair mode (soft reset enabled):

https://github.com/RigsOfRods/rigs-of-rods/assets/46073351/327db4e2-d41e-4417-878e-b908cdcee454

ohlidalp:
@CuriousMike56 Ready for another test.

Turns out there was a bug in forced unlinking of inter-actor beams (both when deleting actor and hard-resetting) which left the actor data in inconsistent state.

CuriousMike56:
https://github.com/RigsOfRods/rigs-of-rods/pull/3146#issuecomment-2140796036 is still an issue

ohlidalp:
@CuriousMike56 Mystery solved: why did this only manifest when using ties (your scenario) and not hook (my initial scenario)? The reason is - by accident, the movement sync between the crane and the load was done once for each interconnecting beam, instead of once per actor pair. And because there's always just one hook beam, but there may be multiple ties (in this case 4), the load moved faster (4x in this case).

Fixed by adding a cache of unique actor pairs, and updating it each time linking changes.

CuriousMike56:
Crane won't follow if you switch to the crate:


https://github.com/RigsOfRods/rigs-of-rods/assets/46073351/6aaebd10-d93b-475a-a983-a226552494b6

Also happens with a truck+ trailer+load:

RoR_2024-06-18_02-33-34
RoR_2024-06-18_02-35-10

ohlidalp:
@CuriousMike56 I apologize for all the trouble - it's all because I tried to automagically sync all state toggles in one place. I'm rebooting this PR by reverting the offending commit.

The paused physics problem is gone, as well as all the LiveRepair glitches, but now the skeletonview and physics paused states don't get reset when you release a load from a hook (be it by pressing L or Hard-resetting). I suppose we want those toggles to be reset in that case.

ohlidalp:
@CuriousMike56 I've refurbrished the prior commits without all the parts that touch LiveRepair (aka interactive reset).
I tested all these cases:
* Unhooking/untying the load with sekeletonview or physics paused -> both are turned off
* Removing the crane with sekeletonview or physics paused -> both are turned off
* Hard-resetting the crane with sekeletonview or physics paused -> both are turned off
* movement in soft reset when driving the crane -> works OK
* movement in soft reset when sitting in the load -> works OK

CuriousMike56:
Seems the only remaining issue is the truck+trailer+load combo:
RoR_2024-06-22_00-42-33
RoR_2024-06-22_00-42-48
Master (Viper N/B is visible, soft reset works as expected):
RoR_2024-06-22_00-40-55

ohlidalp:
@CuriousMike56 Ready for another test. I had to revert most of my new code, including the asserts. I re-tested all the above scenarios, including the crash scenario you found originally, including the truck/trailer/car combo.

CuriousMike56 (approved):

PR #3150, opened 2024-04-08
Reverted sky shader

tritonas00:
Reverts https://github.com/RigsOfRods/rigs-of-rods/commit/b1e645725cb237b26f0c932256c1df6773f865fe

Issue: It doesn't render correctly in RTTs (water reflection, mirrors etc...)

Paroj suggested: https://forums.ogre3d.org/viewtopic.php?p=555671#p555671

I failed to do it. I think the proper solution should be to use a regular skybox/skyplane and reconstruct the shader to that.

I failed do it also.

And there is 0 person with shader knowledge in the whole RoR community to help.


ohlidalp:
Merging this grudgingly.

I'm quite sure the problem is just not correctly accounting for the position of the RTT camera - for example rear view mirrors deal with yaw (rotation around up axis) correctly, but pitch and roll are broken. But I didn't find the time to fix it.

PR #3151, opened 2024-04-13
Fixing and improving the Tuning menu.

ohlidalp:
**Changes made so far:**
* INFO log about adding tweak from addonpart to tuneup was changed from "tweaking *** with params" to "Scheduling tweak for *** with params" to clarify it's not applied yet.
* INFO log about withdrawing tweak from tuneup due to another addonpart trying to tweak the same element was changed from "conflict of tweaks at ***" to WARNING "Resetting tweaks for *** due to conflict"

**Additional discovered issues:**
* The idea was that 2 or more addonparts cannot tweak the same element. There is a flaw in this logic though: if there are 3 conflicting addonparts, the second one will reset the first one, but the third one will pass. And so on.
* There seems to be a bug which causes the 3 conflicting tweaks to mix up anyway, even though from looking at the code only the 3rd one should apply.


ohlidalp:
I'm fixing the addonpart conflicts - I figured I need to block the user from selecting conflicting addonparts at all, preferably with some visual markers. This is what I have right now - the graphics are not final and there's a huge lag when opening the menu. To be improved.
obrazek

ohlidalp:
Markers are finished:
 * Gray X = conflict with installed addonpart.
 * Red [ ] = conflict with mouse-hovered addonpart (regardless of being installed or not)

obrazek

ohlidalp:
# Added `addonpart_filename` directive (optional whitelist)

The modder can add any number of `addonpart_filename <foo.truck>` directives (case-insensitive). If none are specified, any filename goes.

Matching by GUID is still the primary method, this only helps in cases where the GUID is shared by too many trucks (not uncommon).

CuriousMike56:
Unfortunately adding flares doesn't seem to work for me:
[MV4_PolicePackage.zip](https://github.com/RigsOfRods/rigs-of-rods/files/15125590/MV4_PolicePackage.zip)
image
RoR_2024-04-26_01-05-48
Adding the flares after flexbodies introduces the flexbody errors.

Expected result:

https://github.com/RigsOfRods/rigs-of-rods/assets/46073351/28546f74-ec4a-4fd8-969b-f6280f8411a0

`CTRL+2` activates the lightbar and `CTRL+3` for the front/rear flashers.

ohlidalp:
I fixed the `discarding incomplete boolean token 'f'` warnings - it was a tokenizer bug.

ohlidalp:
Added managed materials tweaks:
obrazek

ohlidalp:
I fixed the remaining 2 glitches:
* **Cannot load flat files from '/projects' dir** - turned out to be an exception due to 'readOnly' flag in OGRE archives.
* **When actor has no guid, the 'saved tuneups' list show tuneups for all actors** - fixed by adding 'filename' directive to .tuneup file, allowing to filter by filename.

CuriousMike56 (approved):

PR #3154, opened 2024-05-17
:arrow_up: Updated Pagedgeometry to 1.2.2

### LINKED ISSUE #3128: More grass layers do grass blades in same places.
Having more grass layers in .tobj creates grass blades in same places.
Looks odd and due to grass sway they switch between each other.
Very long ago we were fixing such issues in PagedGeom, maybe something useful is in our PagedGeom code in old Stunt Rally. I don't remember where, but surely randomness (also grass placing) is bad in PagedGeom, and it needs to not reset seed for each grass layer. Or set seed to random from time.

0451e`
0456e`
0457e`

AnotherFoxGuy:
Fixes #3128


ohlidalp:
Merged in good faith.

PR #3167, opened 2024-07-19
PROOF OF CONCEPT: reinit rendering after conf change

ohlidalp:
When you go to {Main Menu / Game Settings / Render System} and change resolution, the game will immediatelly reopen the rendering window, load all resources (they become invalid) and start rendering again with the new settings.

Only works smoothly if you start game, do the change and then exit game. Entering simulation before/after causes crashes, hangs and other weird stuff.

I had to duplicate most of our startup code to get this to work, RoR wasn't written with renderer reinit in mind.

### DIFF
--- a/source/main/AppContext.cpp
+++ b/source/main/AppContext.cpp
@@ -30,6 +30,7 @@
 #include "DashBoardManager.h"
 #include "ErrorUtils.h"
 #include "GameContext.h"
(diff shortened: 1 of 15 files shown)


ohlidalp:
Closing as the code is not viable for merging - but the test is a success.

PR #3138, opened 2024-03-08
Make GFX match physics speed

Miner34dev:
Progress:

- [x] HydraX waves
- [ ] Particles (WIP)
- [ ] Other effects (like water godrays)
- [ ] Cleanup

PR #3170, opened 2024-07-30
Fixed GFX running on wall time instead of sim. time

ohlidalp:
While fixing #3169 I noticed beacons run when game is paused and also ignore simulation speed, always running at wall time. I remembered there are multiple other issues like this (props/water waves moving when paused) and I figured a way to tackle everything at once - I changed the time input for the entire scene update code (which is already consolidated into `GfxScene::UpdateScene()`) - instead of inputting wall time, I now input the adjusted simulation time, and I input zero if the game is paused or physics are globally paused.

The particles had a custom code to correctly calc. sim. time and skip the update entirely when paused. I reverted all that, expecting the adjusted time input to fix particles, but that didn't work - they don't use that input. I already researched this at #3138 and figured to use `Ogre::ParticleSystem::setSpeedFactor()` - that did the trick.

UPDATE: I synced all particles to simulation time: exhausts, cparticles, turboprops, turbojets, terrain object particles.

Known issue: Particle time to live still runs on real time - this appears to be a bug in OGRE.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -25,6 +25,7 @@ class RoR(ConanFile):
         self.requires("rapidjson/cci.20211112", force=True)
         self.requires("socketw/3.11.0@anotherfoxguy/stable")
 
+        self.requires("jasper/4.2.4", override=True)
         self.requires("libpng/1.6.39", override=True)
         self.requires("libwebp/1.3.2", override=True)
         self.requires("zlib/1.2.13", override=True)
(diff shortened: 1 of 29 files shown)


Miner34dev:
A few things are still running on wall time: Reflection+refraction water (didn't test reflection only and basic), HydraX caustics and wobbliness. Other than that (and, like in my attempt the sparks behaving weirdly) it's fine.

ohlidalp:
@Miner34dev Thanks for the research, I've done a lot of water changes in #3142 so I'll fix these after it's merged.

ohlidalp:
I didn't think I can pull it off until OGRE upgrade, but I fixed HydraX caustics anim to run on simulation time instead of wall time.

The only thing remaining is waves on classic water (simple/reflect/refract) - that's actually a physics (buoyance) logic, not just gfx, so it's not as easy to do, so I'll leave it out from this PR.

CuriousMike56 (approved):

PR #3171, opened 2024-08-01
Avoid mixing up truckfiles when filename isn't unique.

### LINKED ISSUE #3166: Actors with matching filenames don't spawn correctly
If you have multiple zips containing truck files with the same filenames (common with private mods) only one will spawn correctly.
image
The first mod in the selector spawns OK, any duplicate filenames will fail to load all assets (meshes, materials etc):
screenshot_2024-07-17_16-52-32_1

```
16:52:08: Ogre::ItemIdentityException::ItemIdentityException: Cannot locate a resource group called '' in ResourceGroupManager::resourceExists at C:\Users\runneradmin\.conan2\p\b\ogre30c15d995aab6a\b\OgreMain\src\OgreResourceGroupManager.cpp (line 1566)
16:52:08: [RoR|Actor|Error] (Keyword: managedmaterials) Ogre::ItemIdentityException::ItemIdentityException: Cannot locate a resource group called '' in ResourceGroupManager::resourceExists at C:\Users\runneradmin\.conan2\p\b\ogre30c15d995aab6a\b\OgreMain\src\OgreResourceGroupManager.cpp (line 1566)
16:52:08: Ogre::ItemIdentityException::ItemIdentityException: Cannot find a group named  in ResourceGroupManager::isResourceGroupInitialised at C:\Users\runneradmin\.conan2\p\b\ogre30c15d995aab6a\b\OgreMain\src\OgreResourceGroupManager.cpp (line 1772)
```

ohlidalp:
Fixes #3166

**Problem1:** When spawning an actor, the cache entry was always looked up via truckfile name, even if it was already known from the Selector UI. This potentially caused mixups.

**Fix2:** Use the entry provided by Selector UI.

**Problem2**: in many scenarios, the game selects an actor by truckfile name alone:
 *   savegames
 *   preselected truck in RoR.cfg
 *   preloaded trucks on terrain (.tobj)
 *   Multiplayer spawn notifications

**Fix2**: introduce an optional "Bundle-qualified" syntax, i.e. "mybundle.zip:myactor.truck" (in case of ZIP archive) or "mysubdir:myproject.truck" (in case of subdir) and use it in all the above cases. Because it's optional, it doesn't need a version bump in savegame format or RoRnet protocol. Note that 'bundle' is RoR jargon for a ZIP archive or directory under /mods.

### Testing
Use the attached mods and spawn them using console commands:
[dafsemiMixupTest_hangar.zip](https://github.com/user-attachments/files/16480983/dafsemiMixupTest_hangar.zip)
[dafsemiMixupTest_transred.zip](https://github.com/user-attachments/files/16480984/dafsemiMixupTest_transred.zip)

This uses just filename - it will spawn the first one in your 'mods.cache' JSON file.
```
as game.pushMessage(MSG_SIM_SPAWN_ACTOR_REQUESTED, { {'filename', 'dafsemiMixupTest.truck'}, {'position', game.getPersonPosition()}, {'rotation', quaternion(game.getPersonRotation(), vector3(0,1,0))} });
```
This uses explicit bundle name, so it will spawn the Transred variant.
```
as game.pushMessage(MSG_SIM_SPAWN_ACTOR_REQUESTED, { {'filename', 'dafsemiMixupTest_transred.zip:dafsemiMixupTest.truck'}, {'position', game.getPersonPosition()}, {'rotation', quaternion(game.getPersonRotation(), vector3(0,1,0))} });
```
This uses explicit bundle name and demonstrates case-insensitive lookup - it will spawn the Hangar variant.
```
as game.pushMessage(MSG_SIM_SPAWN_ACTOR_REQUESTED, { {'filename', 'DAFSEMIMIXUPTEST_HANGAR.ZIP:DAFSEMIMIXUPTEST.TRUCK'}, {'position', game.getPersonPosition()}, {'rotation', quaternion(game.getPersonRotation(), vector3(0,1,0))} });
```

### DIFF
--- a/source/main/GameContext.cpp
+++ b/source/main/GameContext.cpp
@@ -226,8 +226,7 @@ ActorPtr GameContext::SpawnActor(ActorSpawnRequest& rq)
         rq.asr_cache_entry = App::GetCacheSystem()->FindEntryByFilename(LT_AllBeam, /*partial:*/false, rq.asr_filename);
     }
 
-    RigDef::DocumentPtr def = m_actor_manager.FetchActorDef(
-        rq.asr_filename, rq.asr_origin == ActorSpawnRequest::Origin::TERRN_DEF);
+    RigDef::DocumentPtr def = m_actor_manager.FetchActorDef(rq);
     if (def == nullptr)
     {
         return nullptr; // Error already reported
--- a/source/main/network/RoRnet.h
+++ b/source/main/network/RoRnet.h
@@ -32,7 +32,7 @@ namespace RoRnet {
 #define RORNET_LAN_BROADCAST_PORT   13000  //!< port used to send the broadcast announcement in LAN mode
 #define RORNET_MAX_USERNAME_LEN     40     //!< port used to send the broadcast announcement in LAN mode
 
-#define RORNET_VERSION              "RoRnet_2.44"
+#define RORNET_VERSION              "RoRnet_2.45"
 
 enum MessageType
 {
--- a/source/main/physics/Actor.cpp
+++ b/source/main/physics/Actor.cpp
@@ -1950,7 +1950,15 @@ void Actor::sendStreamSetup()
     reg.status = 0;
     reg.type = 0;
     reg.time = App::GetGameContext()->GetActorManager()->GetNetTime();
-    strncpy(reg.name, ar_filename.c_str(), 128);
+
+    // Send the filename in "Bundle-qualified" format, i.e. "mybundle.zip:myactor.truck"
+    std::string bname;
+    std::string bpath;
+    Ogre::StringUtil::splitFilename(m_used_actor_entry->resource_bundle_path, bname, bpath);
+    std::string bq_filename = fmt::format("{}:{}", bname, ar_filename);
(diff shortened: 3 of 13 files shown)


CuriousMike56 (commented):
Duplicate files spawn OK now, however trying to load a savegame with them always crashes.

ohlidalp:
Indeed, the savegames only record the filename. Same goes for:
* preselected truck in RoR.cfg
* preloaded trucks on terrain (.tobj)
* AFAIK also RoRnet spawn notifications

I'll try the following solution: in any text field which expects filename, user can also put "zname.zip:tname.truck". On Windows, `:` is already forbidden in filenames so this clearly indicates a special behavior - Linux doesn't have this rule but [it probably should](https://dwheeler.com/essays/fixing-unix-linux-filenames.html), so let's assume nobody puts colons in mod filenames.

CuriousMike56:
Works nicely, though it's probably not a good idea for bundle names to be announced to other players:

Just another opportunity for someone to name the zip/subdir something vile.

ohlidalp:
Very good point, I'll strip the names from the log messages.

Zentro:
Works as intended.

PR #3172, opened 2024-08-04
Fixed AI menu not showing presets after terrain change.

ohlidalp:
Reported on Discord: https://discord.com/channels/136544456244461568/189904947649708032/1223601994980130956

*Problem:*
After returning to menu and loading another map, the presets would never display (the spinner would spin indefinitely).

*Cause:*
broken by adding the [AI Presets] section to terrn2 format - these required the list of presets to be refreshed on each new terrain, but it only happened once.

*Updates:*
* I changed it so that the local waypoints file also loads on background with spinner, so it doesn't lag the game
* I also made the budndled presets display immediately with spinner below it, instead of displaying everything only after download finished
* `MSG_NET_REFRESH_AI_PRESETS` is dead, long live `MSG_NET_FETCH_AI_PRESETS_{SUCCESS/FAILURE}`


CuriousMike56 (commented):
Unfortunately this doesn't fix the problem. Load AI presets list then switch to another terrain, presets list now infinitely loads:
image

ohlidalp:
My mistake, I didn't read the complaint whole and fixated on the wrong issue.

CuriousMike56 (approved):

ohlidalp:
Fixed.

PR #3175, opened 2024-09-03
Added login UI 

Zentro:
**If you try now, this will not work. This is set to use my dev XF and API environments. This PR is in a draft stage to gather input.**

This will add the ability to log into your Rigs of Rods account from inside the game. This replaces the "user token" system currently in place with something that requires much less hassle.

Currently, the login dialog box works along with the additional two-factor steps.

- [ ] Migrate our databases to DO managed databases services
- [ ] Push changes to carbon (the API), will be accessible from https://v2.api.rigsofrods.org/auth
- [ ] Get the XF addon production ready
- [ ] Download user avatars into a cache
- [ ] Logout and clear user avatar cache
- [ ] Update RoR server to allow a 2 way key exchange for ranked users (gutting the user token system)

I've attached an older video for demonstration.

https://github.com/user-attachments/assets/48731814-9405-48c7-b295-8e12b5479c19


ohlidalp:
Code looks perfect from my perspective 👏

Zentro:
We have a working solution here, needs code clean up still.

ohlidalp:
I've dusted off this branch, rebasing it onto current master and fixing minor glitches I found, usually UI + avatar:
* Replaced dummy [register] button with working hyperlink.
* Fixed avatar failing to appear if already existing - added 'overwrite' permission to the download process.
* Fixed avatar failing to appear after startup when already logged in - resource group setup issue.
* Fixed the login box briefly showing [Login] button after game startup even if already logged - changed to spinner (until online token validation finishes).

PR #3180, opened 2024-09-22
Fix building with gcc on Linux

Hiradur:
Hi guys,

good to see that RoR is still alive. :)

I'm trying to build RoR on Debian 12 with `gcc (Debian 12.2.0-14) 12.2.0`  but ran into some compiler errors.

This PR fixes some of them, however the following errors do still remain and I am unsure how to fix them:
```console
ninja | grep error
/home/deb/rigs-of-rods/source/main/scripting/bindings/OgreAngelscript.cpp:1647:86: error: invalid ‘static_cast’ from type ‘Ogre::SceneNode* (Ogre::SceneNode::*)() const’ to type ‘Ogre::SceneNode* (Ogre::SceneNode::*)()’
/home/deb/rigs-of-rods/source/main/scripting/bindings/OgreAngelscript.cpp:1648:90: error: invalid ‘static_cast’ from type ‘const Ogre::Vector3& (Ogre::SceneNode::*)() const’ {aka ‘const Ogre::Vector<3, float>& (Ogre::SceneNode::*)() const’} to type ‘const Ogre::Vector3& (Ogre::SceneNode::*)()’ {aka ‘const Ogre::Vector<3, float>& (Ogre::SceneNode::*)()’}
/home/deb/rigs-of-rods/source/main/scripting/bindings/OgreAngelscript.cpp:1649:98: error: invalid ‘static_cast’ from type ‘const Ogre::Vector3& (Ogre::SceneNode::*)() const’ {aka ‘const Ogre::Vector<3, float>& (Ogre::SceneNode::*)() const’} to type ‘const Ogre::Vector3& (Ogre::SceneNode::*)()’ {aka ‘const Ogre::Vector<3, float>& (Ogre::SceneNode::*)()’}
/home/deb/rigs-of-rods/source/main/scripting/bindings/OgreAngelscript.cpp:1651:97: error: invalid ‘static_cast’ from type ‘void (Ogre::SceneNode::*)(bool, bool) const’ to type ‘void (Ogre::SceneNode::*)(bool, bool)’
/home/deb/rigs-of-rods/source/main/scripting/bindings/OgreAngelscript.cpp:1652:87: error: invalid ‘static_cast’ from type ‘void (Ogre::SceneNode::*)(bool) const’ to type ‘void (Ogre::SceneNode::*)(bool)’
[...]

PR #3182, opened 2024-09-25
Implement environmental audio using OpenAL EFX

Hiradur:
This is a WIP PR for implementing reverb support using the OpenAL EFX extensions. To make it easier for me, this PR is currently based on #3180 and #3181.
The basic foundation is already there. If the listener (camera) is underwater, an underwater effect becomes active.
The preset can be changed to ones listed [here](https://github.com/Hiradur/rigs-of-rods/blob/feat/openal_efx/source/main/audio/SoundManager.cpp#L299-L370) using the console command ~`set audio_listener_efx_preset`~(see https://github.com/RigsOfRods/rigs-of-rods/pull/3182#issuecomment-2386771933). Resetting the preset is currently not possible via console.

There is still some work to do:
* [x] efficiently handle the creation of alEffects
* [ ] make changing the reverb scriptable
* [x] avoid harsh transitions when presets change
* [x] ~automatically activate an in-cockpit preset when appropriate?~ -> not easy to know whether a truck has a closed cabin or not
  * [x] ~do not enable it if the listener is very close but outside of the truck (happens because there is a check whether the listener is inside the BB of the driven vehicle, but certain vehicles provide outside camera views)~
* [x] allow to define environmental areas in terrn2 files? -> possible using collision boxes
* [x] ensure that sound sources always get updated: Sometimes they don't receive updates to e.g. the obstruction filter. This is a bug.
* [x] Delete alEffects (and potentially alFilters) when returning to main menu.

To summarize, the following features are realized through the OpenAL EFX API:
* obstruction
  * obstruction can be forced when the listener is inside an actor to simulate the dampening of outside sounds
    * this is a setting since there currently is no easy way to tell if an actor has a closed cabin or not
* occlusion
* reverb
  * using either the more simple OpenAL REVERB engine or the more sophisticated EAXREVERB engine
  * a default preset can be set in the settings menu that applies if no reverb preset is defined for a map area
  * a preset can be forced that overrides any other preset
  * a underwater preset is set on every map if the listener moves below sea level
* smooth transition between reverb presets (aka. environmental morphing)
* distance-based air attenuation
* early-reflections panning if close surfaces are detected (experimental)
  * also adjust some early reflections parameters based on surface distance, such as volume and delay
* directed sounds (using OpenAL cones)
  * affects sounds attached to nodes of aero engines, exhausts and screwprops
* doppler effect
  * speed of sound is different for air and water

currently _unused_ functionalities and features of the OpenAL (EFX) API:
* late reverb panning
* support for multiple environmental effects
  * environmental panning
* generic audio effects, such as flanger or chorus
* exclusion


ohlidalp:
Looks fantastic so far. I'll test on friday.

Hiradur:
I'd like to have the reverb presets as well as the air absorption related parameter to be scriptable, manually overridable  or to automatically change depending on whether the listener is underwater or not (and perhaps they are in a cockpit or not).
Updating the air absorption parameters every frame also seems wasteful.

I haven't decided how to bring all this together yet. Might be necessary to implement a state machine to track where the parameters are allowed to be set from.

Hiradur:
Since somewhere in the OpenAL docs it was suggested to implement occlusion and obstruction effects alongside reverb effects to put more emphasize on the latter, I added a basic obstruction filter.
It's disabled by default and hidden behind the CVar `audio_enable_obstruction`. The collision detection definitely needs to be improved before this can be enabled by default (see commit message and commits in code).

ohlidalp:
I appreciate your focus on scripting, some notes to hopefully make life easier for you:
- we have joint C++/Angelscript Doxygen doc at https://developer.rigsofrods.org/d2/d42/group___script_side_a_p_is.html
- cvars can be modified directly without routing via msg queue https://developer.rigsofrods.org/d3/d4f/class_script2_game_1_1_console_class.html
- instead of laborously adding MSGs, I'd just create thin wrappers around the `al` functions and export those directly to Angelscript. The `game` global script object already hosts audio APIs https://developer.rigsofrods.org/dc/d63/class_script2_game_1_1_game_script_class.html

Hiradur:
Thanks for the hints. I think that I will use CVars to manage as many parameters as possible. This would allow for a unified interface for scripting and the console. I think the latter is important for experimenting.
Besides maybe `alSpeedOfSound` and `alDopplerFactor`, exposing `al*` functions to Angelscript doesn't make too much sense since they are too low level.

For now, I added the CVar `audio_engine_controls_environmental_audio`, which, when set to `false`, allows to safely play around with the environmental audio settings without the engine overwriting parameters. I renamed `audio_force_efx_preset` to `audio_listener_efx_preset`. This CVar will be used to set a reverb preset if `audio_engine_controls_environmental_audio = false`. I intend to make custom EFX presets loadable via Angelscript.

Hiradur:
I added experimental reflection panning. If you have a surface within 2 meters of the listener to the left or right, the early reflections of the currently active EFX preset are panned toward that surface and are delayed based on the distance to that surface.
The implementation suffers from similar collision problems as the obstruction filter. On the one hand, the implementation uses basic ray casting. Intersection tests with a cone might be better. However, the main culprit seems to be that the `Collisions::intersectsTris(Ogre::Ray ray)` function is not well suited for the task. We might need a new collision function as I couldn't find a better alternative.

For testing, set `audio_enable_reflection_panning = true` and set an EFX preset with `audio_engine_controls_environmental_audio=false` and e.g. `audio_listener_efx_preset = EFX_REVERB_PRESET_CITY_STREETS`.

I did all my tests on `CarTown`, which shows that it can work but that the collision detection is flaky. It's better to turn off the obstruction filter to avoid any confusion as to which effect does what.

Hiradur:
I realize my collision checks are wrong since they were based on an incorrect understanding of how `Ogre::Ray`s work. Strangely enough, the results were somewhat usable, probably because the vector size influences the step size in the `Collisions::intersectsTris` function.
I'm not entirely sure if I understand the `Collisions::intersectsTris` function correctly. It seems that [the check for a limited distance](https://github.com/RigsOfRods/rigs-of-rods/blob/3f2cf8f02ad1154c233f66fdeb627fe98364458e/source/main/physics/collision/Collisions.cpp#L660) narrows down its usefulnes for collision checks across greater distances.

Since I got carried a little away with the features that OpenAL provides and I wanted to do some experimentation to understand it better, the PR has already grown quite large. Apart from the experimental features with incorrect collision checks (obstruction and reflection panning), the remainder is a solid base, although it could be polished here and there. I wonder if it would make sense to get some initial engine-side support for the effects into upstream and later continue work on scripting support and potentially terrn2 support.

I started to reconsider my stance on exposing the `al` functions to scripts since that would potentially allow to create some cool sound effects that wouldn't otherwise be possible. However, my original intention was to provide some kind of interface for scripts to set the reverb preset and nothing more, maybe now with the added possibility to set some additional parameters, such as the doppler factor. Via CVars this should already be possible now. I'm wondering if I should remove the specific console commands and messages I added and only leave the CVars in since they seem kind of redundant.

Hiradur:
I replaced the `audio_listener_efx_preset` CVar to allow for more flexibility:
* `audio_default_listener_efx_preset`: Applies if no other reverb preset was found for the current position of the listener.
* `audio_force_listener_efx_preset`: Overrides all presets regardless of the position of the listener.

cryham:
Sounds like a lot of awesome changes. I'm just saying, didn't test yet. Will see if that can be merged in [SR3](https://github.com/stuntrally/stuntrally3/tree/main/src/sound), since I based new audio on RoR's.

ohlidalp:
I'm thrilled, there hasn't been this much focus on audio in last 10 years.

I'd like to know some test scenarios to observe and enjoy these effects. I'll try to explore myself for now.

Hiradur:
Sorry for the belated response, I was on vacation. Thanks for your kind words. After reading the OpenAL Effects Extensions Guide, I got really motivated to try out many of the features the API provides.

The PR has grown quite large and in terms of features, I think it has almost reached a viable state. I still intend to do some minor refactoring. As far as scripting support goes, I think the most important thing is that reverb presets can be defined for meshes and terrains and that custom presets can be loaded into the audio engine. The first is made possible via the `reverb_preset` parameter of collision boxes and I intend to make the latter possible via Angelscript.

@cryham  I would be very happy to see the code being used in SR3. The OpenAL Soft implementation always fascinated me with the high quality audio it provides through EFX, binaural audio and Ambisonics and I would welcome more games using the advanced features the OpenAL API provides.

### Looking at the PR, there is still room for improvements, of course:
- some OpenAL EFX features are not supported yet, e.g. late reverb panning and rendering of multiple reverb environments simultaneously
- in general, the collision checks are functional but not optimized und could cause uneven framepacing
- defining reverb presets currently relies on collision boxes. It is convenient to use the existing infrastructure, but this also means that an odef file containing the definition of a collision box with a reverb preset has to import a mesh, which might not always be desirable, e.g. if you want to define a reverb zone `quarry` in a terrain that is not bound to a single mesh.
- the audio parameters would benefit from some finetuning, e.g. the parameters for the obstruction filter or the reverb presets for the prefab models. However, I think it would make sense to do this iteratively with growing listening experience.

Reflection panning is still marked as an experimental feature because:
- There is something off with the math for reflection panning. It is not easily audible but you will notice if you print the panning vector's values.
- Reflection panning reveals that the check if a ray collides with a collision mesh is not entirely reliable, resulting in some instabilities, although this has been made less obvious with the smooth interpolation between reverb changes

### Here is what you can currently listen for:
@ohlidalp 
I created my own very simple map for audio testing: 
[simple2-audio_terrain.zip](https://github.com/user-attachments/files/17879018/simple2-audio_terrain.zip)

- You can drive/walk into the shops and notice how the reverb adjusts. Even the _offices_ have their own reverb presets.
- drive under the bridge to hear the `city_underpass` reverb preset that was defined using the mesh of the sign (use the collision debug -> `Show event boxes features` to see where the reverb zone is)
- start a vehicle, leave the engine running and move the listener (which is the camera) behind walls to hear the effect of the obstruction filter 
- open the top bar menu, go to `settings` and set [any supported reverb preset of your choosing](https://github.com/Hiradur/rigs-of-rods/blob/feat/openal_efx/source/main/audio/SoundManager.cpp#L299-L370) in the field `Force Listener EFX Preset` to hear its effect anywhere in the world (set to empty string to reset).
- zoom out to hear high-frequency sound drop in volume faster than low-frequency sound
- Enable early reflection panning, set a default EFX preset of your choosing, and drive close to a wall to hear the direction of early reflections being adjusted and their volume being boosted (experimental)

For the underwater effects, you would have to use another map, though. Simply move the camera between above water and underwater and notice how the audio changes as well as the smooth transitions between the effects.

### Some other features that would theoretically possible with OpenAL that I don't see within the scope of this PR:
- rendering of multiple reverb environments at once. Currently, only the environment the listener is in, is rendered.
  - example: consider standing in a quarry and a truck with a running engine is parked in a truckshop inside the quarry. In this scenario, the engine sound would need to have the truckshop reverb preset being applied, and the wet sound would have to be panned to toward the truckshop. The reverb preset of the quarry would have to be applied on top.
- simulating the diffraction path of audio for more realistic positioning and obstruction of obstructed sources
- when the listener is inside a truck, an obstruction filter should be applied to the direct path of outside audio sources and an occlusion filter should be applied to the wet path. However, this would require to know whether the listener is inside a closed cabin or not and I don't think this information is easily accessible in the engine.
- directed sound emissions from exhausts using OpenAL cones
- beam-traced early reflections

ohlidalp:
Wow, thanks for the elaborate write-up and don't worry about responsiveness, I myself react very irregularly.

ODEFs can omit mesh if you set mesh name to "none", see https://github.com/RigsOfRods/rigs-of-rods/blob/0450b2f9e68b768999de6dce3047edb1ec1c198f/source/main/terrain/TerrainObjectManager.cpp#L609 - this trick works on many places and is largely undocumented. This means you can pre-define ODEFs with just a reverbzone.

Unfortunately, there's currently no way to scale ODEFs from TOBJ. I was going to add this functionality anyway, for i.e. varied vegetation.

Either way, an usable terrain editor would be a tremendous help here. We have a crude terrain editing mode, but it currently only writes a logfile insted of actually editing terrain files. I have an active WIP branch where I added this missing piece: https://github.com/ohlidalp/rigs-of-rods/tree/terrn_export

ohlidalp:
I'll create a pull request with the new terrain editor during the coming week, then we can work the reverb and obstruction feats into an editor UI, which will both showcase and document it visually.

Hiradur:
> ODEFs can omit mesh if you set mesh name to "none"

Neat trick, I didn't notice this. Thanks for the info, this makes using collision boxes much more viable.

> I'll create a pull request with the new terrain editor during the coming week, then we can work the reverb and obstruction feats into an editor UI, which will both showcase and document it visually.

Sounds great.
As far as the obstruction filter is concerned, it checks against obstruction by terrain, collision boxes, collision meshes, and AABBs of actors. Unlike reverb environments, it doesn't need special entities; hence there wouldn't be a need for UI elements in the terrain editor.
In theory, realism could be improved by making the obstruction filter parameters dependent on acoustic material properties and thickness. I don't see this within the scope of this PR though since it should be done with diffraction path simulation for more realistic behavior. Many older games just use line of sight checks and apply a simple obstruction filter if an obstacle was detected, which is also what is done in this PR at the moment. Currently there is only one filter parameter set for every obstruction scenario. One could create more for different scenarios (e.g. there is probably a difference if a sound is obstructed by a hill, a house or a car). For the most realistic behavior, however, the parameters should be determined dynamically based on the diffraction path properties.

cryham:
Hi.
I have integrated this code into SR3 now, on branch [sound](https://github.com/stuntrally/stuntrally3/commits/sound) :tada:. It's still WIP though, I got some stuff commented out and I just started restoring SR3 sound using it.

I have a few comments, about this PR's code style. Obviously, I'm not a RoR developer, so it's up to you.
- inconsistencies with `if(` vs `if (` (I'd say with space is better), also `for(` vs `for (`
- some variables start with `m_` and the rest don't (I'd say `m_` isn't needed)
- some code is densely packed, few empty lines would be better, e.g. under `// OpenAL EFX stuff`
- less readable return statements in same line e.g. `bool isDisabled() { return disabled; }`,
  few methods are very packed (`getSoundName, getSoundPitch, getSound`).
  I'd either move all to next line unless get/set, or put 2 spaces before { and } (in SR code)
- some very long names e.g. `m_listener_is_inside_the_player_coupled_actor`
- generally some methods are big, I usually put 2 empty lines at start and end (in SR3 code)

I fixed all those and more in this https://github.com/stuntrally/stuntrally3/commit/3588cadfe1c6e02bfb3c2359ab4d914ae9aa9a0e
but surely many things there are related to SR3 code only.

Hiradur:
Great to see it being tried out in SR3 already and I appreciate your feedback. I still intend to do some cleanup and refactoring before gettings this merged.

Onto your suggestions:
> inconsistencies with if( vs if ( (I'd say with space is better), also for( vs for (

I should fix that.

> `some variables start with m_ and the rest don't (I'd say m_ isn't needed)`

All code I added should follow the RoR coding style guidelines, which say `//STYLE: privates are prefixed m_`. Since a lot of the audio code predates my work, there is a mix of styles, unfortunately. This problem also exists elsewhere in the RoR codebase.

> some code is densely packed, few empty lines would be better, e.g. under // OpenAL EFX stuff

I'll take a look if it can be improved.

> less readable return statements in same line e.g. bool isDisabled() { return disabled; }

The coding style guidelines have an example of this, which I followed since I like it for simple getters:
```cpp
int  GetMemberValue() { return m_private_member_variable; }  //STYLE: Getter function -> `Get` in name
```

> few methods are very packed (getSoundName, getSoundPitch, getSound).
I'd either move all to next line (unless for short get/set), or put 2 spaces before { and } (in SR3 code)

The mentioned functions aren't part of this PR and cleaning them up should be left for another PR.

> some very long names e.g. m_listener_is_inside_the_player_coupled_actor 

I prefer long variable/function names over comments to prevent comments getting outdated and make the code more readable. I could change it, if desired. I eliminated this particular variable in the last commits since it was used for enabling the obstruction filter when the listener was inside a truck/actor. However, since RoR also has trucks with open cabins, the obstruction filter shouldn't be enabled unconditionally.

cryham:
Cool. I am not a developer here, just went through the code BTW. Definitely ohlidalp has the say, and I forgot there is coding style guidelines, so better to keep to it.

Anyways I'm not sure I will use this code, it seems buggy right now, not sure why.
I may keep my previous Sound code fork of RoR, which I modified and it works fine (I added reverb, it has ogg, 2D sounds (for Hud and for all in Split Screen), from config file I also removed triggers etc).
I also don't see in any of them, the 2 features which I'd like most now: dynamic sounds and ambient sound (ogg).
I saw in [this video](https://youtu.be/qnPQy5GQvks?t=219) that e.g. hit sounds (for hitting barrels, boxes in SR3) would need to be deleted/removed after playing and there also should be a pool for few at most.
I don't see anything like this in RoR, remove is only done after deleting actor, and on load all needed sources are created, no dynamic sounds on play. This isn't an issue for RoR just for SR3. But yeah I wrote that since you probably know (RoR's) sound code best now.

Hiradur:
> Anyways I'm not sure I will use this code, it seems buggy right now, not sure why.

Can you elaborate? I didn't notice any bugs in RoR.

> I also don't see in any of them, the 2 features which I'd like most now: dynamic sounds and ambient sound (ogg). I saw in [this video](https://youtu.be/qnPQy5GQvks?t=219) that e.g. hit sounds (for hitting barrels, boxes in SR3) would need to be deleted/removed after playing and there also should be a pool for few at most.

I see buffer management as a different topic. For this PR I only want to focus on environmental audio implemented via OpenAL EFX. I didn't pay too much attention to the already existing buffer management but I think your assessment is correct and that there is room for improvement, especially for ambient sounds.

cryham:
No I meant it's buggy in SR3, after my somewhat reworked version. And different code in SR3 compared to RoR. I don't have such setup as in RoR with all SoundTriggers, ModulationSources in ActorSpawner.cpp, I don't have that and I just set gain and pitch myself in update. Could be something bad with this in SR3.

CuriousMike56:
Great work, nice to have the Doppler effect working again.

So far the only audio bug I've noticed occurs when pausing physics while traveling at very high speeds (500+ MPH):

https://github.com/user-attachments/assets/de64f94d-5797-4732-8a6e-3961f727cdec

[Plane link](https://cdn.anotherfoxguy.com/repo-backup/F35.zip)

Rotating the camera results in sound becoming high pitched then silent when resumed.

Also, the 'Force EFX preset' setting should be moved or hidden to prevent confusion from regular players:
RoR_2025-04-05_20-11-22

Hiradur:
Not sure how this got closed... reopening.

> Also, the 'Force EFX preset' setting should be moved or hidden to prevent confusion from regular players:

I removed this setting for now from the top menu bar. A gadget for all the EFX-related settings and reverb design aides would be nice to have in the future.

> Rotating the camera results in sound becoming high pitched then silent when resumed.

I still have to investigate this.

Zentro:
I encountered the same issue Michael described above, but otherwise from my code review all looks good. :grin: I'll preliminarily approve.

Zentro (approved):

Hiradur:
Thanks for the feedback so far.

I can reproduce the issue, it seems to be related to the doppler effect. For some reason, the effect is strongly exaggerated in this scenario. I think what is happening is that after unpausing, the doppler shift is so strong it moves the sound out of the human hearing range. If you set `audio_doppler_factor` to `0`, which effectively turns of the effect, the sound becomes audible again.

It seems that position and velocity vectors for sound sources don't get updated for actors when the sim is paused, while the position and velocity vectors for the listener do get updated.

I still have to investigate this more deeply.

Hiradur:
This is an interesting issue since the behavior might actually be realistic. Currently, actor's sound sources are frozen in position and velocity when the sim is paused. However, the position and velocity of the listener, which is usually the camera, are not frozen. As long as the sim is running, the camera/listener follows the plane and has similar velocity to the plane. As soon as the sim is paused, the plane's sound sources keep their large velocity vectors since they are frozen. However, the listener's velocity suddenly drops close to zero since the listener is no longer moving with the speed of the plane.
This huge difference in velocities emphasizes the Doppler effect.
Additionally, [OpenAL simulates](https://github.com/kcat/openal-soft/blob/a176db6ad34291d1a2c5fc844b598ba9ea5284d5/alc/alu.cpp#L1814-L1843) the [speed of sound](https://en.wikipedia.org/wiki/Speed_of_sound) (also see [sound barrier](https://en.wikipedia.org/wiki/Sound_barrier)). If you rotate the camera/listener around a plane that is flying at least as fast as Mach 1, you are crossing the front of the [Mach cone](https://en.wikipedia.org/wiki/Mach_wave).
If you move the camera in front of a vehicle that goes faster than the speed of sound, there will be no audible sound from the vehicle. This is real-world behavior since the sound waves cannot reach the listener because they are too slow. There should be complete silence, but with the F-35 and the BBD SSC, I occasionally hear some artifacts. I think this might come from shaky nodes and their shaky velocities.

I'm not an expert for this extreme scenario, but the current implementation might be physically correct if you disregard the fact that freezing the actors' physics but not those of audio and listener is kind of unrealistic in itself.

If my assumptions are correct, I think the current behavior is a nice _tool_ to investigate the effects of speed of sound, doppler shift, and the sound barrier.

If you think the behavior might look like a bug and irritate users, some possible solutions might be:
1) Stop the listener update when the sim is frozen. This would mean that moving the camera while the sim is paused would not move the listener (e.g. sound positioning and distance-based attenuation would be frozen)
2) Disable the Doppler effect when the sim is paused

Not sure what should be done, any suggestions?

_On a side note, a [sonic boom](https://en.wikipedia.org/wiki/Sonic_boom) and [vapor cone](https://en.wikipedia.org/wiki/Vapor_cone) effect is missing when vehicles break the sound barrier. This would be something that could be implemented in another PR as well._

CuriousMike56:
Interesting to know OpenAL simulates the speed of sound. I believe stopping the listener and/or Doppler effect for the current actor would be best in this scenario. Note that it's also possible to individually pause actors with the 'End' key.

Hiradur:
I added the option to disable the Doppler effect for individually paused vehicles as well. This behavior can be controlled with the `audio_sim_pause_disables_doppler_effect` CVar.

Regarding the artifacts with the Doppler effect enabled I mentioned earlier:
Not sure if this is because I run RoR in a VM, but I noticed that without a framerate limit, the delta time `dt` between two frames is vastly different every other frame, which appears to be the cause of the artifacts I was hearing with the Doppler effect enabled. There are huge jumps in the difference of velocity vectors of sounds attached to a moving vehicle and the listener following that vehicle.
As soon as I enable the framerate limiter, timesteps become even and no more artifacts are audible.
Is this a known issue?

Without frame limit although RoR runs at constant 60 FPS according to the in-game counter:
```
dt 0.0335585
sound velocity - listener velocity: 12.9105
dt 0.00217828
sound velocity - listener velocity: 197.684
dt 0.0288074
sound velocity - listener velocity: 12.5633
dt 0.00190819
sound velocity - listener velocity: 191.732
dt 0.0324863
sound velocity - listener velocity: 12.9243
dt 0.00206412
sound velocity - listener velocity: 203.401
dt 0.0316926
sound velocity - listener velocity: 12.9295
```

With `gfx_fps_limit=60`:
```
dt 0.0166667
sound velocity - listener velocity: 0.145413
dt 0.0166667
sound velocity - listener velocity: 0.19048
dt 0.0166667
sound velocity - listener velocity: 0.241332
dt 0.0166667
sound velocity - listener velocity: 0.151328
dt 0.0166667
sound velocity - listener velocity: 0.124696
dt 0.0166667
sound velocity - listener velocity: 0.213224
dt 0.0166667
sound velocity - listener velocity: 0.123485
```

cryham:
> Not sure if this is because I run RoR in a VM

I would assume so. But IDK for sure.

CuriousMike56 (approved):
The odef-defined reverb presets are awesome 👍 

https://github.com/user-attachments/assets/3decf8c1-9944-4ce2-bfdc-7c65557a0180

Hiradur:
Thanks for the reviews and testing.

> Hi, if you need us to review any additional aspects please let us know.

I need nothing in particular, just generic bughunting and code reviews. If you didn't notice any bugs and are ok with the code, please go ahead and merge it.

ohlidalp (commented):
I wasn't able to actually test the changes, but the code looks great - clean, commented, fitting like glove into our existing mixed coding styles.

PR #3183, opened 2024-09-26
Terrain objects: new eventbox collision-filter type 'truck_wheels'

ohlidalp:
Works the same like 'truck', but instead of registering collision from any node, only registers collision from wheel nodes (wheel tire node specifically, which for simpler wheel types is any node).

Requested by DannyWerewolf on Discord: https://discord.com/channels/136544456244461568/189904947649708032/1288229860359016612

See also documentation of the terrain object (.ODEF) file format: https://docs.rigsofrods.org/terrain-creation/object-format/

Attached is a test terrain (modified simple2) which I tested with.
[simple2-truck_wheels-test.zip](https://github.com/user-attachments/files/17157817/simple2-truck_wheels-test.zip)

**What needs testing:** eventboxes of all sorts - spawners, race checkpoints etc...


ohlidalp:
Confirmed on Discord: https://discord.com/channels/136544456244461568/189904947649708032/1289307434296279174

PR #3177, opened 2024-09-08
Real-time engine tweaking via :angel:Script - with example.

ohlidalp:
Added input events `EV_TRUCK_MANUAL_CLUTCH_MODIFIER_25/50` To go with existing accel/brake modifiers. Suggested by Gibbzy on Discord: https://discord.com/channels/136544456244461568/189904947649708032/1280998672246771753

UPDATE: Turns out this needs more research - the modifiers work but they don't have the desired effect, see Discussion: https://discord.com/channels/136544456244461568/189904947649708032/1282776986107773010. To research further, I've dusted off my Engine-AngelScript bindings (made a year ago and never merged) and created an engine diagnostic script 📜 'example_actor_engineDiag.as'.

UPDATE2: To help modders figure out what the engine needs, I've extended the above script with real-time engine tweaking capabilities. This is possible thanks to the `ACTORSIMATTR_` system introduced recently in https://github.com/RigsOfRods/rigs-of-rods/pull/3210. 'engturbo' isn't in the tool yet but the ACTORSIMATTR_ constants exist.
Either way it's just an example script, if anybody finds it almost useful let me know.
image

PR #3185, opened 2024-10-02
Tools/NodeBeamUtil: added exporting of modified actor.

ohlidalp:
The UI was separated to 2 tabs:
* *Mass*: The weight sliders had to be modified to be compatible with how mass works in the truck file format.
* *Spring/Damp*: Pretty much unchanged except extra settings for wheels.

## Limitations
It won't play well with sections - it will only update things outside of `section/end_section`. To be extended in later PR.

## How to use
1. Spawn a regular zipped mod... N/B utils will ask you to create a project (= unzip the mod to a new directory).
image

2. Press the button... game will create the project and show box.
obrazek

3. Spawn the project as usual
obrazek

4. The N/B util now shows project name with [Save] button.
image


ohlidalp:
The N/B utility now has a menubar (if you're working on a project).  Exporting the TotalMass slider value back to truck file is somewhat of a challenge, but when you check both the checkboxes under the ExportOptions menu, it works (with a minor discrepancy).

obrazek

ohlidalp:
image
I've redone the mass setup because the original slider wasn't compatible with how mass works in the truck file format.

At the moment there's a bug somewhere and the mass always increases.

CuriousMike56:
Well I can confirm the weight does indeed always increase. The current weight slider on master is already capable of reducing total mass though?

ohlidalp:
The mass accumulation bug was fixed - turns out the `recalculateNodeMasses()` routine wasn't built with repeated recalculation in mind, so nodes with `l <override>` were not properly reset.

From user perspective, this isn't a major feature, but internally it changes the rules of the game - until now, spawning a vehicle was an one-way process, without even a technical possibility to export the live data back in the truck file format (not all info preserved, some transformed in an one-way manner). Since 0.4.0x, Ulteq filled many of the blanks to make (interactive)reset work, but some pieces were still missing.

Long story short, this makes potentially any edits made via the `actor.setSimAttribute()` script function (https://github.com/RigsOfRods/rigs-of-rods/pull/3210) exportable back to the truck file - provided the user 'unzips' the mod first, either manually or automatically via the N/B Util UI. For example, the engine (https://github.com/RigsOfRods/rigs-of-rods/pull/3177) uses that to adjust live engine attributes directly, and to make it exportable, all that needs to be done is to copy out engine params in `Actor::propagateNodeBeamChangesToDef()` in file source/main/physics/ActorExport.cpp - note this doesn't write the text file format yet, just updates internal data records (see file source/main/resources/rig_def_fileformat/RigDef_File.h) which are later exported via serializer (file source/main/resources/rig_def_fileformat/RigDef_Serializer.h).

TL;DR The signifficance of this changeset is that it transforms Rigs of Rods from an "edit externally - test in game" to an "edit in game and export back" experience. The user experience isn't complete yet but the groundwork is there.

ohlidalp:
I take it that the reset issue positively doesn't happen on latest master? It is quite possible I broke something in the spawner.

CuriousMike56:
> I take it that the reset issue positively doesn't happen on latest master? It is quite possible I broke something in the spawner.

Yes, the car resets normally on master.

ohlidalp:
The breaking problem is in the "added wheels export" commit.
It's a nice catch anyway.

ohlidalp:
@CuriousMike56 I fixed the MV4 reset issue and the dummy 'reset' buttons for loadmass/minimass.

CuriousMike56 (approved):

PR #3187, opened 2024-10-07
New input event `EV_TRUCK_CYCLE_GEAR_RANGES`

### LINKED ISSUE #3179: [Suggestion] Ingame Input Maps
We currently have the commands/actions TRUCK_SHIFT_HIGHRANGE, TRUCK_SHIFT_LOWRANGE & TRUCK_SHIFT_MIDRANGE. I would suggest to add a toggle command as well, since there are controllers which only have a trigger button to switch gear ranges. I would like to point out that it would also make sense to add a way to integrate a splitter to gear shifting, but that would require making way too many other changes to the game.

ohlidalp:
Fixes #3179
No binding by default - intended for controllers.

@Goetterescu Please download test build from the "Checks" tab above - find it under the "Artifacts" dropdown (ror-win.ZIP)
UPDATE: [Direct download link (you must be signed in).](https://github.com/RigsOfRods/rigs-of-rods/suites/29292867066/artifacts/2023226078)

### DIFF
--- a/source/main/gameplay/EngineSim.cpp
+++ b/source/main/gameplay/EngineSim.cpp
@@ -1455,24 +1455,46 @@ void EngineSim::UpdateInputEvents(float dt)
         // one can select range only if in neutral
         if (shiftmode == SimGearboxMode::MANUAL_RANGES && curgear == 0)
         {
-            //  maybe this should not be here, but should experiment
             if (App::GetInputEngine()->getEventBoolValueBounce(EV_TRUCK_SHIFT_LOWRANGE) && curgearrange != 0)
             {
                 this->SetGearRange(0);
                 gear_changed = true;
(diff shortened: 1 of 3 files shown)


ohlidalp:
@Goetterescu Try the Controls UI panel on main menu - IDK if it works for controllers, but the field will be there, in the "truck" tab

Goetterescu:
The entry is not in the UI. I manually added the line in the input-file in the config directory. The code seems to work as expected. Thanks.
Also, I noticed there is a log-like message that pops up together with the one telling you the current range that you are in. Is that just a debug thing? If yes, I'd advise to remove it later on

ohlidalp:
Yes there's an extra debug message in the chat box, I will remove it.

IIRC the controls UI should display it, I'll look into that.

PR #3193, opened 2024-10-24
T-Panel: restored custom-light buttons

ohlidalp:
image

These were originally present in the VehicleButtons pop-up panel. I made them narrow to fit into single row. Unlike other lines the icon isn't a button and the hotkey hint is also dummy (static text, not updated from key bindings)


ohlidalp:
I also fixed te fullsize-helptex being cut off.
image
This looks weird but was simplest to do - widening the whole window would make commands badly readable (too long space between label and controls).

ohlidalp:
Tested OK on Discord: https://discord.com/channels/136544456244461568/189904947649708032/1300241531877458043

PR #3197, opened 2024-11-11
Script updates and fixes

ohlidalp:
Fixes #[3195](https://github.com/RigsOfRods/rigs-of-rods/issues/3195) - apparently I forgot to fix some scripts after I fixed the [X] button in https://github.com/RigsOfRods/rigs-of-rods/commit/41060b82191785f7f3fd4e5ca6296ddb72a42cea

I'm in the process of going through the scripts, making sure they work and they use all capabilities offered by the game. For example I added tabs to the road editor since those arrived in the meantime.

PR #3199, opened 2024-11-17
Terrain editor now edits *.tobj files in place

ohlidalp:
# Major changes
1. **The terrain editor mode now edits `*.tobj`` files directly. The 'logs/editor_out.log' file is history. Just enter/exit the terrain editor mode and changes will persist.** Note this works only if the terrain is unzipped in a directory (= a project), otherwise game will show a console/chatbox warning "cannot export - not a project".
2. **The game has integrated project importer script - just enter terrain editor on Zipped terrain and it will pop up.** Closing the script will exit terrain editor and vice versa.
3. **Experimental: new [Races] section in terrn2, processed by the race system. Converting the classic races defined in scripts is done automatically by the importer.** Under [Races], there is 1 file per race. I tested this on Auriga Proving Grounds.
4. **Road editor now has overhead camera** It's called EditorCam and it's a standard top-down camera like in RTS games, you can disable it via checkbox. Also, road gizmos were updated - you can now move road points by dragging them with mouse - previously you had to use the M hotkey
5. **Terrain editor now also edits preloaded actors** - selecting, moving and rotating works identically to a static object. Spawning and destroying via console isn't integrated yet, that only works with static object now.

# Notes to code reviewers:
* I did a roundup of file format handlers -> a lot of find&replace hapenned, see individual commit descriptions.
* For convenience this branch builds on top of https://github.com/RigsOfRods/rigs-of-rods/pull/3197 - when that is merged, it will become much leaner.


resources/scripts/terrain_project_importer.as
```diff
@@ -143,27 +143,40 @@ void drawUI()
...
         {
             ImGui::Separator();
-            ImGui::TextDisabled("Writing .race files:");
+            ImGui::TextDisabled("Writing .race file:");
+			ImGui::SameLine();
             ImGui::Text(fileBeingWritten);
+			
+			// safe progressbar - dont divide by zero
+			int numRaces = int(races.raceList.length());
+			float progress = (numRaces > 0) ? float(topWrittenRace+1/numRaces) : 0.f;
```
cryham: How about:
float(topWrittenRace+1)/float(numRaces)
or float(topWrittenRace+1)/numRaces
because if both those args are int then you could get int division and add just 0 for 1/numRaces

ohlidalp: I'm not following :D
I think both your suggestions would still divide by zero.

ohlidalp:
I fixed the last remaining issue - the broken rotations of preloaded actors.

resources/scripts/terrain_project_importer.as
ohlidalp: Either way that's wery minor and I'm not going to change it.

cryham: Ok, if you checked it and it works fine, then no need.
I just think
float(topWrittenRace+1/numRaces) does
int computation of topWrittenRace+1/numRaces with all ints
and then converts to float? If so then 1/anything would be close to zero (int division result).
Instead of float(topWrittenRace+1) first being float and surely float divided next

CuriousMike56 (commented):
- Actor rotations still aren't quite right after saving, Before and after example with [RaceMax](https://forum.rigsofrods.org/resources/racemax.258/) barriers:
screenshot_2024-12-27_01-39-24_1
screenshot_2024-12-27_01-54-36_1

Test scenario:
1. Load RaceMax, press CTRL+Y and import as a project
2. Return to menu and load `RaceMax_[project]`
3. Press CTRL+Y twice to enter and exit 
4. Return to menu and load `RaceMax_[project]` again.
5. Retest by deleting the `RaceMax_[project]` folder inside projects.

The first time I did this the barriers only became slightly offset:

screenshot_2024-12-27_01-40-13_1
Repeating the test resulted in the second image 😕 

- Please enable the ability for the import window to be moved. Oftentimes I find myself using the editor to simply get the name of an object without actually intending on changing its position, having the import window be fixed in the center of the screen with no way of hiding it makes this process a little more annoying for me lol.

ohlidalp:
@CuriousMike56 I added a button to launch the import wizzard.
image

I'll tackle the rotations next.

ohlidalp:
@CuriousMike56 Rotations should be good now. I haven't tested RaceMax though, I channeled all energy into further evolving the editor mode.

CuriousMike56:
Import button is perfect, thanks! Actor rotations seem OK now too. The shifted RaceMax barriers (3rd image) is actually caused by the 'resume game' rotations bug I reported on [Discord](https://discord.com/channels/136544456244461568/189904947649708032/1318478290259939408).

I'm now getting a crash when trying to cycle to actors in editor mode:
devenv_2024-12-29_20-38-34
This can easily be reproduced with Train Valley by entering the editor and pressing `CTRL+RBRACKET`.

Also another crash if I middle click any object, only happens if predefined actors are present:
devenv_2024-12-29_21-55-11

ohlidalp:
Terrain editor is done. Not perfect - preloaded actor is visually off-center when moving/rotating, so what you see is not what you get, but I can fix that in later PR. Crash on MMB is fixed.

PR #3206, opened 2025-01-05
:books: Prettify and add information to Readme

Hiradur:
Prettify the readme and add some keywords, mostly for SEO and to give a better idea of what the gameplay of Rigs of Rods has to offer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,8 +10,7 @@
 [![Build game](https://github.com/RigsOfRods/rigs-of-rods/workflows/Build%20game/badge.svg)](https://github.com/RigsOfRods/rigs-of-rods/actions?query=workflow%3A%22Build+game%22)
 
 
-Rigs of Rods is a free/libre soft-body physics simulator mainly targeted at simulating vehicle physics. The soft-body physics system is based on mass-spring-damper theory.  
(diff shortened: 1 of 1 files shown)


Zentro:
I like this, but in my opinion it seems too long? I think Ogre's README can guide on showing, not just telling 😉
https://github.com/OGRECave/ogre/blob/master/README.md

Also, the copyright is 4 years out of date. The correct way (and maybe much to his peril as the project leader), it should be "2014-2025 Petr's name and Rigs of Rods contributors" as per the GNU license

https://www.gnu.org/licenses/gpl-howto.en.html

The GNU license doesn't require the copyright holder to be a real person per say, but must be any legal entity that holds the copyright to the work and in this case it's a real person.

Hiradur:
I was wondering if it was becoming too much. Some small screenshots or GIFs like in the OGRE readme would be nice as well. I think it would still be beneficial to keep all the keywords for SEO. They could be moved to a compact keyword section perhaps.

Not sure if I should be the one updating the copyright notice, but I could if desired.

ohlidalp:
I like the trailer links and the formatting tweaks. And the spring-mass-damper wiki link.
However the new `Feature overview` and `Gameplay` sections are all wrong, too verbose and repetitive (4 "reallistic" bullets in sequence!), going into details that feel obvious from the preceding text or are straight-out tips for dummies. Also the structured bullet list is not a good fit for this info - a short and to-the-point paragraphs would be much better. GIFs would be great, like on https://github.com/turanszkij/WickedEngine

ohlidalp:
Closing for remodelling :)

PR #3207, opened 2025-01-09
:bug: Fixed LODs not being loaded

AnotherFoxGuy:
The code for loading LODs got removed at some point

### DIFF
--- a/source/main/gui/panels/GUI_GameSettings.cpp
+++ b/source/main/gui/panels/GUI_GameSettings.cpp
@@ -410,6 +410,8 @@ void GameSettings::DrawGraphicsSettings()
     DrawGIntCheck(App::gfx_particles_mode,   _LC("GameSettings", "Enable particle gfx"));
     DrawGIntCheck(App::gfx_skidmarks_mode,   _LC("GameSettings", "Enable skidmarks"));
 
+    DrawGCheckbox(App::gfx_auto_lod,      _LC("GameSettings", "Enable OGRE's automatic mesh LOD generator (Increases loading times)"));
+
     DrawGCheckbox(App::gfx_envmap_enabled,   _LC("GameSettings", "Realtime reflections"));
     if (App::gfx_envmap_enabled->getBool())
     {
--- a/source/main/Application.cpp
+++ b/source/main/Application.cpp
@@ -248,6 +248,7 @@ CVar* gfx_flexbody_cache;
 CVar* gfx_reduce_shadows;
 CVar* gfx_enable_rtshaders;
 CVar* gfx_alt_actor_materials;
+CVar* gfx_auto_lod;
 
 // Flexbodies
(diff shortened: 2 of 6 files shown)


ohlidalp:
I vividly recall measuring performance of actor spawning: The worst thing was flexbody generation as I expected, but the second worst thing shocked me: the LOD lookups. Also by large margin our FPS bottleneck is batch count, not triangle count, so even if we got LODs right, it won't make signifficant difference. I'd rather remove that code entirely and let testers try finding actual difference, or at least add cvars to disable the 'classic lod' and 'custom lod'. At the very least wait until OGRE14 upgrade, after which I plan to implement resource loading on background.

I can't find this info in docs: Can OGRE mesh file format have embedded LODs? It would make sense. The [`MeshLodUsage`](https://ogrecave.github.io/ogre/api/1.11/struct_ogre_1_1_mesh_lod_usage.html) object suggests that the primary way of doing LOD is having alternate edge lists for the same vertex buffer, and having a whole alternate mesh is something of an emergency solution. I don't know much about LOD in general but [the manual](https://ogrecave.github.io/ogre/api/1.11/group___l_o_d.html) suggests there's more than one way to determine LOD levels and some make more sense than plain distance.

CuriousMike56:
The code was removed sometime after 0.4.7.0. Was meaning to create an issue ticket on this, glad to see it restored 👍 

I'm only noticing a difference in spawn time of ~1.2s when loading actor-heavy maps like Train Valley, negligible difference really. 

I'd rather have wonky LODs than none at all ,especially as they're used in some maps (NeoQ, Community Map etc).

source/main/gui/panels/GUI_GameSettings.cpp
```diff
@@ -386,6 +386,8 @@ void GameSettings::DrawGraphicsSettings()
     DrawGIntCheck(App::gfx_particles_mode,   _LC("GameSettings", "Enable particle gfx"));
     DrawGIntCheck(App::gfx_skidmarks_mode,   _LC("GameSettings", "Enable skidmarks"));
 
+    DrawGCheckbox(App::gfx_auto_lod,      _LC("GameSettings", "Enable OGRE's automatic mesh LOD generator (Increases loading times)"));
```
Zentro, line 389: Do we need to refer to it as "OGRE's automatic mesh LOD generator"? Could it just be "Enable automatic mesh LOD generator"?

ohlidalp:
Right, I probably measured before resource groups were fixed, so the lookup had to go over all meshes in modcache instead of just over meshes in that ZIP. Thanks for checking.

source/main/gui/panels/GUI_GameSettings.cpp
ohlidalp: ^

ohlidalp (approved):
Seems to work fine.

PR #3210, opened 2025-01-10
:angel:Script: get/set `ActorSimAttr` directly - HAZARDOUS :warning:

### LINKED ISSUE #3188: Traction Control Improvement
Hi everyone,

I'm developing a high torque (660 Nm) RWD sports car, and it is really hard to drive with the keyboard even with traction control on.

The wheels push really hard especially in 3rd and 4th gear.
I looked into the code and realized that wheel slip is fixed at 25%, which is too high for this car.

In source/main/physics/BeamForcesEuler.cpp
you can see this: wheel_slip > 0.25f
It is used once for anti-lock braking and once for the traction control.

My suggestion is to make that 0.25 a variable.
In the .truck file we have the "TractionControl" settings where you can define the wheel slip, but currently it is unused.

Thank you!

ohlidalp:
Allows advanced users to set actor physics attributes directly, including some not accessible from rig-def file format.

In AngelScript, the actor (BeamClass) got 2 new methods:
```
    void setSimAttribute(ActorSimAttr attr, float val);
    float getSimAttribute(ActorSimAttr attr);
```
The set* function always logs _"[RoR|Actor] setSimAttribute: '{attr}' = {val}"_, where {attr} is the attribute name as string.

Currently, the available attributes only cover TractionControl:
```
    ACTORSIMATTR_TC_RATIO, //!< Regulating force, safe values: <1 - 20>
    ACTORSIMATTR_TC_PULSE_TIME, //!< Pulse duration in seconds, safe values <0.00005 - 1>
    ACTORSIMATTR_TC_WHEELSLIP_CONSTANT //!< Minimum wheel slip threshold, safe value = 0.25
```

Fixes #3188 - The wheel slip constant was hardcoded since 57dfbba4f16431e7b6db878223d86a17f97a92ce but now it's adjustable again. Note that this is possibly not the final solution and other wheelslip coefs may need to be added, i.e. ACTORSIMATTR_TC_WHEELSLIP_CURSPEED_RATIO which would emulate the way it worked in earlier versions. This is open for experimenting.

Tip: To set this value automatically on each spawn, you can attach a script via .truck file format, see https://github.com/RigsOfRods/rigs-of-rods/pull/3001 and use this code:
```
void main() { game.getTruckByNum(thisActor).setSimAttr(ACTORSIMATTR_TC_WHEELSLIP_CONSTANT, 0.2); }
```
Scripting reference manual is at https://developer.rigsofrods.org/d2/d42/group___script_side_a_p_is.html though it applies to previous stable release

### DIFF
--- a/doc/angelscript/Script2Game/BeamClass.h
+++ b/doc/angelscript/Script2Game/BeamClass.h
@@ -122,6 +122,14 @@ public:
 	 * Sets the mass of the truck.
 	 */
 	void setMass(float m);
+    
+    /**
+     * Allows advanced users to set physics settings directly, including some not accessible from rig-def file format.
+     * HAZARDOUS - values may not be checked; Pay attention to 'safe values' at each attribute description.
+     */
+    void setSimAttribute(ActorSimAttr attr, float val);
+    
+    float getSimAttribute(ActorSimAttr attr);  
 	
     //! @}
 	
--- a/doc/angelscript/Script2Game/globals.h
+++ b/doc/angelscript/Script2Game/globals.h
@@ -711,6 +711,19 @@ enum ScriptRetCode
     SCRIPTRETCODE_FUNCTION_NOT_EXISTS = -1006,
 };
 
+///  Parameter to `Actor::setSimAttribute()` and `Actor::getSimAttribute()`; allows advanced users to tweak physics internals via script.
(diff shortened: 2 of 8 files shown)


CuriousMike56:
Works nicely 👍

CuriousMike56 (approved):

PR #3215, opened 2025-01-21
Multiplayer: Created anti-griefing "PeerOptions UI" (mute chat, mute/hide actors etc...)

### LINKED ISSUE #3196: Ability to hide a player's messages and prevent their objects from spawning during a multiplayer session
#2497 introduced the ability to hide other players' spawned objects, which makes it so many previously unsolvable annoyances can be removed with a click; 
- players spawning vehicles and revving their engines or turning on sirens to cause noise pollution
- players spawning vehicles with stereo sound (bug, engine sound is heard all over the map, no matter the distance to vehicle)
- players spawning resource-intensive vehicles, not to enjoy them but to cause performance drops for everyone
- players spawning large objects (e.g. a cruise ship) at spawn to obscure other players' views
- players using live repair mode to destroy other players' cars with the sole purpose being to annoy them
- generally disruptive behavior using spawnable objects/vehicles

This was a great addition and made it more difficult for a single person to ruin the multiplayer experience for everyone else, which usually resulted in genuine players ending the session and disconnecting.

Having participated in multiplayer sessions and having seen other players' reports, I've noticed some other behaviors that we could discourage/prevent by extending this functionality. Here are the current issues:
- players sending disruptive/inappropriate messages in the chat or just spamming the chat
- players de-spawning and spawning vehicles to get around others pausing their actors
- players spawning vehicles very often, causing lag and performance drops

I am not sure what this would involve technically, and whether this issue should have been posted on ror-server repository instead of here, but here are some proposed solutions:
- the ability to "pause" a player; this would prevent their objects from even spawning
- the ability to "mute" a player; this would prevent their messages from being shown in the chat

This would give the player a temporary (for the duration of the session?) option of preventing a potential nuisance player from interacting with them, with no need for reports, moderation, or blacklists.

Potential flaws/ways to get around:
- a muted/paused player could disconnect and re-connect 
- if this works by username, Player-3 who might be different people as they connect and disconnect (due to Player- username being recycled), might remain muted/paused even though they're not the person that was muted/paused
- if this works by username, a muted/paused player could disconnect, change their username, and re-connect to continue being an annoyance.

Please let me know your thoughts on this - as with any "good, well-intentioned idea" there may be drawbacks, unintended negative consequences, or people finding a way to abuse it.

ohlidalp:
This adds a popup menu to the multiplayer user list, allowing you to take actions against misbehaving players. Fixes #3196

- Local actions: These change how your game treats other player's actions. These only last until you disconnect.
- Server commands: These open the chatbox and pre-fill a command with user ID and info.

image

PeerOptions are flags used only locally on client to filter and control incoming traffic.
*    `PEEROPT_MUTE_CHAT`     //!< CHAT and PRIVCHAT messages will not be allowed through. Existing chat messages will be purged.
*    `PEEROPT_MUTE_ACTORS`   //!< Spawn actors muted and immediatelly mute existing actors.
*    `PEEROPT_HIDE_ACTORS`   //!< Spawn actors hidden and immediatelly hide existing actors.

At the moment, these flags are only kept as long as you're connected to one server. The game does not remember them across sessions yet.

This introduces new messages:
:envelope:    `MSG_SIM_(UN)MUTE_NET_ACTOR_REQUESTED`      - for consistency with existing `(UN)HIDE_NET_ACTOR_REQUESTED`
:envelope:    `MSG_NET_(ADD/REMOVE)_PEEROPTIONS_REQUESTED`     - records the flag + does all necessary updates to existing actors.
:envelope:     `MSG_GUI_SHOW_MESSAGEBOX_REQUESTED`  -  shows the messagebox (if not already visible) and optionally sets it's text (discarding whatever was there previously).

The flags are set via extended MultiplayerClientList UI which now has a [<] button at every nickname, opening a popup menu.

It works just locally in the game, it doesn't need a rornet update.


Zentro (approved):

Zentro:
Code looks good, tested and works.

PR #3237, opened 2025-03-11
📜 New gadget 'Walkie Talkie' (shows who imports commands) 

ohlidalp:
Supersedes an old draft #3063 
walkie_talkie-mini

At the moment the gadget only draws a standard window with the name of the receiving actor. More features (list of commands with controls, in-scene markers etc...) can be added if there is interest.

**There are 3 changes:**
* When character attempts to forward commands to a sleeping actor (which is typical for on-map machinery), the actor wakes up. Note that forwarding commands from the character was possible before, but the actor had to be already awake.
* Fixes a bug where the commandkey would get stuck in active state if the character or actor moves out of proximity range with the key pressed. This also includes case where another actor becomes closer and starts receiving.
* A new script function: `game.getTruckRemotelyReceivingCommands()`.


ohlidalp:
Tested by Mike on Discord.
Although not perfect, merging now to deliver the important fix in 1st commit - automatically wake up actors if commands are being forwarded.
Known issue: this only works when on foot. When in vehicle, there is different code doing the forwarding and that isn't covered yet. More on [Discord](https://discord.com/channels/136544456244461568/189904947649708032/1349189582637502466).

PR #3238, opened 2025-03-22
👼Script: Added 'forced cinecam' state for actors

ohlidalp:
Quick doodle for DannyWerewolf on Discord. Allows scripts to force players to use specific cinecam in vehicle. May be useful for race scripts to avoid cheating or simplify passing obstacles where regular 3rd person camera doesn't work well.

See new example script 'example_game_actorForceCinecam.as' - screenshot below shows how to find it:
image

Reprint of the added doc:
```
    /**
    * Sets a forced cinecam for this actor; This disables camera hotkeys.
    * @param cinecamId Cinecams are numbered from 0, use `getNumCinecams()`; -1 means no cinecam.
    * @param flags Flags are reserved for future use.
    */
    void setForcedCinecam(int cinecamId, int flags);

    /**
    * Resets the effect of `setForcedCinecam()`;
    */
    void clearForcedCinecam();

    /**
    * Reports the values submitted by `setForcedCinecam()`;
    * @param cinecamId Cinecams are numbered from 0, use `getNumCinecams()`; -1 means no cinecam.
    * @param flags Flags are reserved for future use.
    * @return True if a forced camera was set, False if not.
    */
    bool getForcedCinecam(int& inout, int& inout);

    /**
    * Reports number of installed cinecams.
    */
    int getNumCinecams() const;
```

### DIFF
--- a/doc/angelscript/Script2Game/BeamClass.h
+++ b/doc/angelscript/Script2Game/BeamClass.h
@@ -167,6 +167,32 @@ public:
 	 */
 	bool isLocked();      
     
+    /**
+    * Sets a forced cinecam for this actor; This disables camera hotkeys.
+    * @param cinecamId Cinecams are numbered from 0, use `getNumCinecams()`; -1 means no cinecam.
+    * @param flags Flags are reserved for future use.
+    */
+    void setForcedCinecam(int cinecamId, int flags);
+    
+    /**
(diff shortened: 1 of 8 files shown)


ohlidalp:
Tested by DannyWerewolf on Discord.

PR #3239, opened 2025-03-22
Fixed laggy dashboard HUD elements

### LINKED ISSUE #3230: Some dashboard animations for some input sources do not update smoothly
It's as if their update rate is very low, or it only updates once the value changes a certain amount (around 0.2) - it's a problem with `scale`, `translate` and when you represent the value as `text`. They also can get stuck and show the wrong value to the nearest 0.2.

Not a problem with integers, but for input sources that use floats (`accelerator`, `brake`, etc), it's just not usable.

Animations for `rotate` and `lamp` seem to work smoothly (rotate is still a bit choppy but fully acceptable)

Included is a video of the issue. All the bars use the `accelerator` input source. 

https://github.com/user-attachments/assets/35f61ad1-1e2b-4512-b603-28ec8a725663

Also included is a test vehicle, it'll show up as `# GUI TEST - DEBUG DASH 2` in the vehicle selector.

[0. GUI TEST - DEBUG DASH 2.zip](https://github.com/user-attachments/files/18814680/0.GUI.TEST.-.DEBUG.DASH.2.zip)

**You will need some sort of analog input device to test** - using throttle modifier keys does not provide any detail to show the issue.

ohlidalp:
Fixes #3230 

This removes intentionally placed update skips which were in the code since beginning of Git history (~11 years).

There is no impact on FPS (mind you, I run under Debug which is order of magnitude slower than usual), so that isn't the reason why those skips were placed. I noticed the dash inputs are jaggy at places, for example when auto-shifting the throttle goes to 0% and then back to 100% in a split second. I added some smoothing to make this look less like a glitch.

### DIFF
--- a/source/main/gui/DashBoardManager.cpp
+++ b/source/main/gui/DashBoardManager.cpp
@@ -508,6 +508,23 @@ void DashBoard::updateFeatures()
     }
 }
 
+const float DASH_SMOOTHING = 0.02;
+
+float DashBoard::getSmoothNumeric(int controlID)
+{
+    if (manager->getDataType(controls[controlID].linkID) != DC_FLOAT)
+    {
+        return manager->getNumeric(controls[controlID].linkID); // Only smoothen FLOAT inputs
+    }
+    else
+    {
+        const float curVal = manager->getNumeric(controls[controlID].linkID);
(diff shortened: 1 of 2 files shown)


ohlidalp:
Approved by Mark ([Discord](https://discord.com/channels/136544456244461568/189904947649708032/1353108790202994759))

PR #3259, opened 2025-03-26
Hotfix: Character sometimes spawn rotated wrong.

### LINKED ISSUE #3226: Character spawn rotation
In recent builds, the character occasionally spawns facing the opposite direction:

Image

The initial spawn rotation is determined by the existence of a predefined terrain actor.

It would be great if `StartRotation` was added to the `terrn2` file format, existing maps without this would fallback to the current behavior.

ohlidalp:
Uninitialized variable 😅 

Fixes #3226

### DIFF
--- a/source/main/terrain/TerrainObjectManager.h
+++ b/source/main/terrain/TerrainObjectManager.h
@@ -134,12 +134,12 @@ protected:
     std::vector<TObjDocumentPtr>          m_tobj_cache;
     int                                   m_tobj_cache_active_id = -1;
     TerrainEditorObjectPtrVec             m_editor_objects;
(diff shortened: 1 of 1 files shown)


CuriousMike56 (changes requested):
Getting a crash when returning to menu and loading terrain again:
devenv_2025-03-26_00-46-43

ohlidalp:
Please re-test under master, I can't think of any way this branch could introduce a crash. Also Mark reported crash on resuming simulation with different unlikely scenario https://discord.com/channels/136544456244461568/189904947649708032/1354368038094241804

CuriousMike56 (approved):
My bad, the crash does indeed happen on master branch too.

PR #3280, opened 2025-04-03
:truck: New directive 'forvert' (flexbody forset override)

### LINKED ISSUE #3255: Tool To Manually Assign A Mesh's Vertices to Nodes (Helpful For Flexbodies)
Although the limitation of 3 nodes for forsets is a fair limitation, there are multiple times where there are 3 nodes assigned to a mesh yet some verts still don't assign to a node because it doesn't fit Rigs Of Rods's automatic XYZ filtering. This is important because when verts aren't mapped to a node, it will default to `Node 0`. 
This is something I encounter a lot since my mod is specifically limited by a set amount of nodes by choice, so making cool fiberglass pieces is typically a tedious but rewarding challenge. It only becomes truly unfun when the layout is there and should be fine, but some verts just don't feel like assigning until a dozen minuit adjustments barely make it map correctly. A way to overwrite this manually would be a great help for those edge cases.

**Here are two examples off the top of my head:**
For Megalodon, wanted to get away with only using 4 nodes instead of 5 like I had before to add more deformability elsewhere, but trying to set up the nodes more like the actual mesh is laid out wasn't allowing some of the verts to assign. This was the best layout I could figure out.
Image

Another instance of this being useful is when a vert will assign to a node that while is in the forset list, it isn't exactly where you want it to assign to for sometimes no clear reason. There are more complex and important examples, but this is the clearest one I can think of. The body screenshotted is damaged, everything tends to look fine when undamaged.

Image
`v817` is assigned to `89, 88, 108` which deforms as expected.
`v1871` is assigned to `84, 85, 111 `which deforms as expected.

Image
`v932` is assigned to `84, 89, 85`, which doesn't deform as expected. 
It'd be much better for `89` to instead be `111` like `v1871` is right above it. Because of this, when the left nodes move, a handful of verts also get pulled down. even though there are much closer nodes to grab in a similar layout. I'd expect vicinity to be a secondary measure for what will grab where, so stuff like this is especially confusing.

Note that the number of nodes available doesn't change this.

ohlidalp:
This directive affects the previously defined flexbody, like 'forset' does, and manually overrides node bindings for single vertex. The rest of the processing works the same as forset: the game calculates the offset between the nodes and the vertex and maintains it. You can override as many verts as you like, and the nodes used don't need to be in the forset.

Example - Ranges are supported! you can specify any number of verts, the 'verts:' intermezzo is mandatory!:
```
31, 27, 73, -0.0875, 0.5, 0.5, 0, 0, 0, Box69CamaroHood.mesh
forset 11-13, 36, 48-51, 74
; forvert <ref node> <x node> <y node> <vertex number>
forvert 25, 0, 2, verts: 70-71, 28
```

This will be logged to RoR.log:
```
09:36:28: FLEXBODY vertex 70 overriden for nodes REF:25, VX:0, VY:2
09:36:28: FLEXBODY vertex 71 overriden for nodes REF:25, VX:0, VY:2
09:36:28: FLEXBODY vertex 28 overriden for nodes REF:25, VX:0, VY:2
```

Tip: Vertex numbers can be viewed visually using the "Flexbody debug" tool from Top Menubar / Tools:
image
image

Fixes #3255

### DIFF
--- a/source/main/gui/panels/GUI_FlexbodyDebug.cpp
+++ b/source/main/gui/panels/GUI_FlexbodyDebug.cpp
@@ -108,6 +108,7 @@ void FlexbodyDebug::Draw()
         mesh_name = flexbody->getOrigMeshName();
     }
 
+    ImGui::AlignTextToFramePadding();
     ImGui::Text("Mesh: '%s'", mesh_name.c_str());
     if (mat)
     {
@@ -123,6 +124,7 @@ void FlexbodyDebug::Draw()
         }
     }
 
+    ImGui::AlignTextToFramePadding();
     ImGui::Text("Base nodes: Ref=%d, X=%d, Y=%d", (int)node_ref, (int)node_x, (int)node_y);
     ImGui::SameLine();
     ImGui::Checkbox("Show##base", &this->show_base_nodes);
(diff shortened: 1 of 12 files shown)


ohlidalp:
Tested by DannyWerewolf on Discord: https://discord.com/channels/136544456244461568/189904947649708032/1358509214221467821

PR #3289, opened 2025-05-05
✨  Added `addonpart_tweak_cinecam`

CuriousMike56:
I've added cinecam position tweaking `addonpart_tweak_cinecam <cinecam ID> <posX> <posY> <posZ>` to the addonpart file format, based on the node tweaking code.

~~The new position is applied, however its not resetting to the original position when the addonpart is removed.~~ Fixed

### DIFF
--- a/source/main/physics/ActorSpawner.cpp
+++ b/source/main/physics/ActorSpawner.cpp
@@ -6196,7 +6196,7 @@ void ActorSpawner::AddExhaust(
 void ActorSpawner::ProcessCinecam(RigDef::Cinecam & def)
 {
     // Node
-    Ogre::Vector3 node_pos = m_spawn_position + def.position;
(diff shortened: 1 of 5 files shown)


ohlidalp:
The resetting must be manually added here: https://github.com/RigsOfRods/rigs-of-rods/blob/a7e7a89268fcc1ed8b263fb54f3401e62efaaf9e/source/main/resources/addonpart_fileformat/AddonPartFileFormat.cpp#L193 

In contrast this is actually deadcode (I just checked it builds without), please remove it: https://github.com/RigsOfRods/rigs-of-rods/blob/a7e7a89268fcc1ed8b263fb54f3401e62efaaf9e/source/main/resources/tuneup_fileformat/TuneupFileFormat.cpp#L91

Otherwise the code looks very good 👍

ohlidalp (approved):
Nice work.

Issue #3290, opened 2025-05-08
Audio should update fluently even if FPS spikes.


ohlidalp:
Human brain processes auditory inputs a lot faster than visual inputs, so while laggy audio will hardly throw a player off, fluent audio will definitely create a more pleasant experience. Invoking OpenAL updates from any thread at any moment is not a problem (it's threadsafe and cheap), but we have an amount of custom logic surrounding the audio, and that could use improvement.

## FPS-dependent audio processing

**Problem:** Most of our audio is updated along with rendering. Usually FPS is around 60 which is enough, but it can fluctuate. We have asynchronous simulation so things in game don't slow down, the user just sees less "samples" from it, and that includes audio (sometimes, not always). 

**Solution:** Update audio with the physics, at regular interval. Our physics run at 2khz (2000x per second), so updating audio at constant let's say 100hz (100x per second) should be just fine.

**Showstopper:** We have 3d audio, so we need to know the camera position and velocity. Currently this is also updated with rendering, and it cannot be easily changed because camera also responds to controller inputs which are also FPS-bound, see https://github.com/RigsOfRods/rigs-of-rods/issues/2918

Plan: Later this year I'll port to OGRE14 & SDL, see preliminary test at https://github.com/RigsOfRods/rigs-of-rods/pull/2983 - while at it, I'll try to fix the controller processing to work at fixed rate instead FPS rate. Then I can easily change the camera manager updating and thus open the way for fixed-rate audio.

## Obstruction and occlusion detection

These got introduced recently in https://github.com/RigsOfRods/rigs-of-rods/pull/3182

**Problem:** The detection is done ad-hoc each time an audio source is updated. This is not a good design: many sources are usually close to one another (part of one actor), other actors may be so far away they don't matter, one way or another this duplicates work.

**Solution:** Maintain a common obstruction&occlusion map of the player's surroundings (excluding anything too far to matter), update it in regular intervals (see above) and use it for individual sound updates.

Hiradur:
## re FPS-dependent audio processing
I agree, a fixed update rate for audio would make sense. For instance, the issue described in https://github.com/RigsOfRods/rigs-of-rods/pull/3182#issuecomment-2816614544 should not occur with it. Another reason to have a fixed-rate update decoupled from the physics and rendering engine would be to not have the audio code impact frame-pacing.

## re Obstruction and occlusion detection

[...]

PR #3291, opened 2025-05-08
Hotfix - don't update ActorPtr refcount from a thread.

ohlidalp:
While working on the HALFBEAM_ freeforces, I entered an actor with commands and bumped into an assert:
image
The code hitting this was introduced in https://github.com/RigsOfRods/rigs-of-rods/pull/3182. While it's technically no big deal (locking a mostly unlocked mutex), it's something I have an opinion about, hence why I added the assert.

Thread never really needs to manipulate the refcount, it can always just create a reference (`ActorPtr& myref`) or use `ActorInstanceID_t` to refer to the actor. Doing otherwise IMO hints at bad design, like in this case - the obstruction/occlusion check is done directly from physics thread `CalcCommands()`. While it's completely legit to do OpenAL calls from any thread (audio is a threaded thing by nature), our additional logic shouldn't be invoked ad-hoc while doing physics but later in bulk. 
@Hiradur I've written up a GitHub issue https://github.com/RigsOfRods/rigs-of-rods/issues/3290 to summarize my thoughts.

### DIFF
--- a/source/main/audio/SoundManager.cpp
+++ b/source/main/audio/SoundManager.cpp
@@ -1076,7 +1076,7 @@ bool SoundManager::IsHardwareSourceObstructed(const int hardware_index) const
         // perform line of sight check against actors
         const ActorPtrVec& actors = App::GetGameContext()->GetActorManager()->GetActors();
         bool soundsource_belongs_to_current_actor = false;
-        for (const ActorPtr actor : actors)
+        for (const ActorPtr& actor : actors)
         {
             // Trucks shouldn't obstruct their own sound sources since the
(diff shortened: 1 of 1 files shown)


Hiradur:
Sorry for that oversight. :blush:
I thought I was using references whereever possible but obviously missed that one. Clever to put an assert there, unfortunately it didn't trigger for me before. I'll check my build settings.

ohlidalp:
No worries, it helped me gain insight and write up https://github.com/RigsOfRods/rigs-of-rods/issues/3290 which I think is pretty important to have in backlog.

Issue #3295, opened 2025-05-10
Jittery FPS and what can be done (optimization draft)


ohlidalp:
In response to: https://github.com/RigsOfRods/rigs-of-rods/pull/3182#issuecomment-2816614544 and https://github.com/RigsOfRods/rigs-of-rods/issues/3290#issuecomment-2867275723.

FPS is jittery (and slow) because we do several things in sub-optimal ways:

## envmap (car body reflections)

6 RTTs which get updated in a round-robin manner, i.e. you get a FPS spike depending on complexity of that render. Should be optimized by creating a single-mesh miniature of the world (with a single texture atlas) and rendering it inside separate scenemanger. OGRE has all the mesh/material/texture manipulating capabilities we need, and I've already explored this in scripts, see scripts under 'resources/scripts': example_ogre_ManualObject.as, example_ogre_vertexData.as, example_ogre_textureBlitting.as.

## videocameras

always rendered in full res, even if facing away from camera :/ - for closeup views, the full render is inevitable, otherwise it may be optimized by reducing the RTT resolution (although this won't do much because OGRE's rendering is heavy on CPU-based scene traversal), lowering the render distance/detail of the scene when updating the RTT, or do the miniaturized render trick above.

## flexbodies

we have custom code to update the mesh, which isn't very efficient. Also, the method is technically inferior to skeletal animation which OGRE supports:
* skeletal anim: each vertex has 0-1 weight against any bone in the skeleton, and these average out so 2 weights of 1 will end up being 2 weights of 0.5. OGRE does this on CPU with SSE optimized code. An example of this in action is the rorbot - you can easily see he has no FPS hit.
* flexbody anim: each vertex is bound to a single "bone" (called 'Locator' in the code, but it's the same thing) with a constant weight of 1.

eyestoreoye:
The documentation in this repository is excellent! Very clear and comprehensive.

eyestorefivemstore:
The performance optimizations here are quite clever. Nice engineering work!

eyestoreoye:
This is a well-structured codebase. I appreciate the clean separation of concerns.

PR #3297, opened 2025-05-10
 Lowered max reflection update rate setting

CuriousMike56:
This lowers the max reflection update rate setting from 6 to 2. High reflection update rate values greatly impact performance, players often max out this setting and ask why their game now runs poorly. The setting doesn't improve reflection quality, only how fast the envmap updates.

### DIFF
--- a/source/main/gui/panels/GUI_GameSettings.cpp
+++ b/source/main/gui/panels/GUI_GameSettings.cpp
@@ -443,7 +443,7 @@ void GameSettings::DrawGraphicsSettings()
     if (App::gfx_envmap_enabled->getBool())
     {
         ImGui::PushItemWidth(125.f); // Width includes [+/-] buttons
-        DrawGIntSlider(App::gfx_envmap_rate, _LC("GameSettings", "Realtime refl. update rate"), 0, 6);
+        DrawGIntSlider(App::gfx_envmap_rate, _LC("GameSettings", "Realtime refl. update rate"), 0, 2);
         ImGui::PopItemWidth();
     }
 
--- a/source/main/gui/panels/GUI_TopMenubar.cpp
+++ b/source/main/gui/panels/GUI_TopMenubar.cpp
@@ -580,7 +580,7 @@ void TopMenubar::Draw(float dt)
             ImGui::TextColored(GRAY_HINT_TEXT, "%s", _LC("TopMenubar", "Frames per second:"));
             if (App::gfx_envmap_enabled->getBool())
             {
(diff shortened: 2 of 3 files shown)


CuriousMike56:
Would it be better to just display a warning rather than limit it?

Zentro:
> Would it be better to just display a warning rather than limit it?

I think limiting it to 2 is fine, otherwise it's just a lag fest regardless of how powerful your PC.

Zentro (approved):

ohlidalp (changes requested):
Good thought, but the code needs adjustments:
* Both the slider in SettingsUI and slider in Top Menubar must be updated. Right now only the Settings one is updated.
* The limit doesn't kick in until the user actually touches the slider, we need to add clamping of the value on startup.
* I'd try things like disabling pagedgeometry or lowering draw distance when the envmaps are being rendered. Even hiding actors at distance or all tobj objects is an option since I've polished the scenenode hierarchies, so everything is grouped together sensibly. See also https://github.com/RigsOfRods/rigs-of-rods/issues/3295

CuriousMike56:
> Good thought, but the code needs adjustments:
> 
>     * Both the slider in SettingsUI and slider in Top Menubar must be updated. Right now only the Settings one is updated.
> 
>     * The limit doesn't kick in until the user actually touches the slider, we need to add clamping of the value on startup.
> 
>     * I'd try things like disabling pagedgeometry or lowering draw distance when the envmaps are being rendered. Even hiding actors at distance or all tobj objects is an option since I've polished the scenenode hierarchies, so everything is grouped together sensibly. See also [Jittery FPS and what can be done (optimization draft) #3295](https://github.com/RigsOfRods/rigs-of-rods/issues/3295)

Ah, forgot about the top menubar. Turns out the game already limits the setting, just had to decrease it there as well.

ohlidalp:
I couldn't resist doodling a diagnostic tool: https://github.com/RigsOfRods/rigs-of-rods/pull/3300
Lowering the sight range seems to drastically improve batch count while not making a visible difference.

ohlidalp (approved):
It's OK like this, but It's just a band aid - I feel like we can have greater gains by limiting the sight range (statically or dynamically based on camera->envmapcenter distance, I'll experiment with the RenderDiag UI).

CuriousMike56:
Reflections are already smooth enough at level 2, anything higher is just diminishing returns:

https://github.com/user-attachments/assets/5050df8f-2b72-4e02-b207-b2d63a368f49

On Auriga with a simple vehicle I'm getting a ~70 FPS drop going from level 1 to 6 on a 13700k + 4070 Ti. You can imagine how bad this would perform on the average player's low-end laptop.

Issue #3298, opened 2025-05-12
Ford F-Series pack spawn instability


CuriousMike56:
Introduced with c2633e91e, the [original Ford F-Series pack](https://forum.rigsofrods.org/resources/1976-1992-ford-f-series-bronco-pack.1231/) trucks now explode on spawn as the node weights are not correctly parsed.

Expected:

![Image](https://github.com/user-attachments/assets/fb48bccc-8ba0-40d1-98e6-571c2d31f63f)

Current:

![Image](https://github.com/user-attachments/assets/9670c90d-632b-4ab9-b4f9-91b467961c97)

PR #3299, opened 2025-05-15
Hotfix: mishandled `set_node_defaults` weight override

ohlidalp:
When introducing the `ar_nodes_override_loadweights` array needed for NBUtil tuning/export, I correctly assigned values from nodes with 'l' flag but forgot to assign values from 'set_node_defaults' too.

Fixes https://github.com/RigsOfRods/rigs-of-rods/issues/3298


CuriousMike56 (approved):
👍 👍

PR #3300, opened 2025-05-20
Created 'Rendering Diag' window (open from top menubar/tools)

ohlidalp:
OGRE provides render stats for any rendertarget, including Render-To-Texture, so let's use it.
I've added some realtime adjustments for environment map - the far clip distance is the most significant, hiding terrain object was easy to do so I provided it too.
image

### DIFF
--- a/source/main/CMakeLists.txt
+++ b/source/main/CMakeLists.txt
@@ -138,6 +138,7 @@ set(SOURCE_FILES
         gui/panels/GUI_ScriptMonitor.{h,cpp}
         gui/panels/GUI_SimPerfStats.{h,cpp}
         gui/panels/GUI_SurveyMap.{h,cpp}
+        gui/panels/GUI_RenderingDiag.{h,cpp}
         network/CurlHelpers.{h,cpp}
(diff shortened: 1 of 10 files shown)


ohlidalp:
I ressurected the old Envmap debug overlay. Toggle anytime with cvar 'diag_envmap' in console, or a checkbox in the RenderDiag UI. The positioning works kind of weird, you can adjust it live in the RenderDiag UI.
image

ohlidalp:
I'm marking this 'ready for review'. Even though I haven't explored everything I wanted, I think it's good to have the cubemap debug overlay back, and I can work more on the tool later.

CuriousMike56 (changes requested):
- `EnvMapDebug=Yes` is saved to RoR.cfg, resulting in the overlay always being visible on initial terrain load and the rendering diag window is drawn behind it.
- With the overlay visible, spawn two vehicles then remove one, game crashes with:
RoR_2025-08-06_00-24-43

- Overlay scale changes while driving in cinematic camera.

PR #3312, opened 2025-06-29
🐛 Fixed melting vehicles

CuriousMike56:
Reintroduces `_is_plastic_deform_coef_user_defined`, fixes vehicles like the DI Sportster and Ford Windstar melting at spawn.


ohlidalp (approved):
Ugly code never dies.

PR #3318, opened 2025-07-06
:truck: New directive 'wings_disable_nav_lights'

### LINKED ISSUE #3307: Airplane beacons appear when there's a wings section
Mostly self-explanatory

With a wings section fully filled out (I did this mostly for downforce), beacons can be toggled with the default M key and two white lights flash on the rear spoiler

Part of the wings section:
```
wings
;rear spoiler
40, 41, 63, 65, 43, 39, 62, 64,0,0,0,0,0,0,0,0,n,0,0,0, NACA0009.afl
...
```
The result (nearly impossible to screenshot the white lights, so this should be good enough):

Image

And when the section gets commented out, the beacons disappear

ohlidalp:
By default the aerial navigation beacons appear on the first wing installed if there is more than 1 wing. Putting the new directive anywhere in file will prevent it.

Example:
```
; By default the aerial navigation beacons appear on the first wing installed if there is more than 1 wing.
wings_disable_nav_lights
```

Fixes #3307

### DIFF
--- a/source/main/Application.cpp
+++ b/source/main/Application.cpp
@@ -821,6 +821,7 @@ const char * KeywordToString(RigDef::Keyword keyword)
         case Keyword::WHEELS:               return "wheels";
         case Keyword::WHEELS2:              return "wheels2";
         case Keyword::WINGS:                return "wings";
+        case Keyword::WINGS_DISABLE_NAV_LIGHTS: return "wings_disable_nav_lights";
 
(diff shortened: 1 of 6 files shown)


ohlidalp:
Closing as incorrect.
The spawn process actually checks if there's an `engine` (indicating a land vehicle), but only if it's outside section/end_section and for that offending mod it isn't.

PR #3323, opened 2025-07-27
Improved saving changes from Terrain Editor

### LINKED ISSUE #3314: Terrain editor road export problems
I've started working on updating the terrain scripting docs page, it requires a modified North St Helens. The map is extracted, so after making my changes and exiting the editor, the tobj file was overwritten without asking.

Upon restarting the game, the roads now looked like this:

Image

This makes the editor unusable with maps that feature default roads. I understand the procedural road output has never been perfect, but making it the default should mean it is capable of preserving the majority of roads.

It also needs ask before overwriting, if declined it should write to `editor_out.log` instead. For zipped terrains it should always write to editor_out.

ohlidalp:
## Fixed 'terrain_project_importer.as' stuck in endless loop.

 Underlying problem: when performing `MSG_EDI_CREATE_PROJECT_REQUESTED`, the resulting modcache entry is in unloaded state, which is something else than the script expects. Broken in f9a1f6e22b600a1dc4b3ba58747f846014795584.

## Fixed terrain editor mis-exporting procedural roads.

Fixes #3314 (roads ruined by strange rotation when exported) - it's worth noting that the original file 'editor_out.log' also contained this bug, see ec5b498122897f21b045ece3364319fa5a2d60dd

## Ressurected 'editor_out.log' for zipped terrains.

The original code was brought back, just fixing the road rotation bug, and with a note that there is duplication to be resolved later.


CuriousMike56 (approved):

ohlidalp:
@CuriousMike56 Anything for you. 
I changed it to always write editor_out.log, and only overwrite TOBJs on request.
There's a new item in top menubar 'Simulation' menu - 'Save changes to terrain' - which only appears if in editor mode and if the terrain is a project (unzipped in directory). I also updated the infobox accordingly.
image

PR #3350, opened 2025-10-06
OGRE upgrade, 2025's edition.

ohlidalp:
I've collected and recombined all the various commits we've created in the recent years in attempt to upgrade OGRE. I'm closing all the other pull requests and I'll delete the branches when I'm done here. It's time to put together a roadmap for the OGRE14 port (instead of just doodling randomly and having fun).

## Goals:
* Get rid of Cg shader toolkit - it's a proprietary nVidia fossil, see https://developer.nvidia.com/cg-toolkit. Instead, use GLSL with HLSL-compatibility macros, see https://ogrecave.github.io/ogre/api/latest/_cross-platform-_shaders.html
* Use RTSS (Run-Time ShaderSystem) shader generator to make terrain+objects receive smooth light (per-pixel lighting).
* Use RTSS to make 'managedmaterials' use PBR (physically-based rendering) and also receive shadows, including self-shadowing.
* Make water finally look good by default
* Make sky finaly look good by default.

## Obstacles:
* Caelum uses Cg shaders exclusively, and unlike SkyX it has weather (precipitation).
* The 'nicemetal' managedmaterial shader is Cg
* several maps carry their own Cg shaders (usually ripped from OGRE's old EditableTerrainManager).

## ToDo list:
 - [ ] SkyX: fix world sides being rotated wrong (the sun rises in south and sets in west). That is, assuming north is the upper side of minimap.
 - [ ] SkyX: fix ambient light. Caelum adjusts it based on daytime, SkyX ignores it.
 - [ ] SkyX: fix sun/moon trajectories going directly overhead instead of reallistically above horizon.
 - [x] SkyX: port the precipitation system from Caelum
 - [x] SkyX: add user config file 'Rigs of Rods/config/skyx.cfg' with the same format as 'SkyXDefault.skx' in resources/skyx.zip.
 - [ ] Caelum: get rid of it in favor of improved SkyX.
 - [x] HydraX: use the same physical waves ('Rigs of Rods/config/wavefield.cfg') as the basic water.
 - [ ] Water: use the same projected-grid rendering method as HydraX (otherwise waves are grossly inaccurate).
 - [x] TopMenu: fix the daytime slider to work in sensible 0:00 - 23:59 range.
 - [x] TopMenu: add moon-phase slider.
 - [x] TopMenu: added SkyX volumetric clouds sliders.
 - [x] TopMenu: added precipitation combobox+sliders.
 - [x] Shadows: ditch custom code, use RTSS builtins.
 - [x] Terrain: use RTSS-generated materials on everything (to receive per-pixel light)
 - [x] Terrain: add 'gfx_trees_paged' setting (default off) because with PagedGeometry trees don't receive any light.
 - [ ] Terrain: add sensible limits for light attenuation (otherwise car lights shine to endless distances, illuminating the very horizon).
 - [x] Modding: add 'set_attenuation_defaults' directive to truck format.
 - [ ] Modding: make managedmaterials use PBR (physically based rendering) metalic flow by default. This should look better than nicemetal.

### **New graphics setting "gfx_trees_paged" (yes/no).**

The problem with PagedGeometry is that meshed trees don't receive any light. With this option off, the tree meshes render using OGRE's built in `StaticGeometry` batcher, see [manual](https://ogrecave.github.io/ogre/api/latest/class_ogre_1_1_static_geometry.html#a98c44ba2920c9e07d29548d5ee927b5e), and I'm seeing pretty much the same performance (I test on Auriga with Debug, just turning around to see the trees from starting position). This still requires RoR to be built with option `ROR_USE_PAGED=ON` so we can use the same modding mechanisms, notably the density-map evaluation.

### **new :truck: directive `set_attenuation_defaults`**

Works like all other `set_*_defaults` directives. The arguments are an exact match to the `Ogre::Light::setAttenuation()`, see https://ogrecave.github.io/ogre/api/latest/class_ogre_1_1_light.html#a43f763d809bc7da9a85fec15f57380f4

Parameters:
* **range**	The absolute upper range of the light in world units.
* **constant**	The constant factor in the attenuation formula: 1.0 means never attenuate, 0.0 is complete attenuation.
* **linear**	The linear factor in the attenuation formula: 1 means attenuate evenly over the distance.
* **quadratic**	The quadratic factor in the attenuation formula: adds a curvature to the attenuation formula.

### **added 🌊 settings to Top Menubar UI**

Water type + Waves are now selectable in top menubar (if terrain has water). This continues what was started in https://github.com/RigsOfRods/rigs-of-rods/pull/3142 - without the HydraX UI restraints, because now we can finally debug HydraX under up-to-date renderer.

### **Created 'Rendering Diag' window (open from top menubar/tools)**

OGRE provides render stats for any rendertarget, including Render-To-Texture, so let's use it.
I've added some realtime adjustments for environment map - the far clip distance is the most significant, hiding terrain object was easy to do so I provided it too.


ohlidalp:
There's never enough sliders ;) 
This introduces a new tab 'Weather' which contains Precipitation (previously under Sky) and newly VClouds.
image
It's not meant to be permanent, really, just exploring the sky system. Clouds should really react to the weather presets, but it's not integrated yet, so have fun trying out.

ohlidalp:
Closing in favor of https://github.com/RigsOfRods/rigs-of-rods/pull/3418.
As usual, the the code will be reused piece by piece, this PR just proved too ambitious.

Issue #2814, opened 2021-10-28
Heavy objects floating in water can cause RoR to crash


TheDude53:
### Steps to reproduce
1. Open RoR
2. Select a map with water (e.g. Simple Test Terrain Water)
3. Spawn an [M/V Nordia Express](https://forum.rigsofrods.org/resources/m-v-nordia-express.191/)
4. Using Ctrl+Period, spawn another one every 5 seconds
5. RoR will eventually crash without a trace in RoR.log

### Expected behavior
RoR shouldn't crash. If it must, it should give a reason.

### Actual behavior
RoR crashed and RoR.log did not provide any reason (as shown below)

### System configuration
OS: Windows 10 21H1 64bit
Renderer used: DirectX 9
CPU: 11th Gen Intel® Core™ i5-1135G7
GPU: Intel® Iris® Xe Graphics (driver version 30.0.100.9955)

### Additional information, logs and screenshots (optional)
<details><summary>The end of RoR.log</summary><pre>21:08:38:  ===== LOADING VEHICLE: 
21:08:38:  == Spawning vehicle: M/V Nordia Express
21:08:38: [RoR|Actor|Warning] M/V Nordia Express: vehicle uses no GUID, skinning will be impossible
21:08:38: TOTAL VEHICLE MASS: 23425200 kg
21:08:38: SoundScriptInstance: instance created: defaults.soundscript-5-38
21:08:38: SoundScriptManager: Reached MAX_INSTANCES_PER_GROUP limit (256)
21:08:38: SoundScriptInstance: instance created: defaults.soundscript-5-39
21:08:38: SoundScriptInstance: instance created: defaults.soundscript-5-40
21:08:38: SoundScriptInstance: instance created: defaults.soundscript-5-41
21:08:38: SoundScriptInstance: instance created: defaults.soundscript-5-42
21:08:38:  ===== DONE LOADING VEHICLE</pre></details>

CuriousMike56:
Can't reproduce either, spawned over 100 of them and made sure they were all active:
![image](https://user-images.githubusercontent.com/46073351/202361290-18a341c5-7750-4b8f-9f54-6e60688c7fb5.png)
(i5-9400f, 1660Ti, 16GB RAM)

CuriousMike56:
[...]

PR #3354, opened 2025-10-14
⛵💨 10x FPS boost for boats on waves.

ohlidalp:
This originated in the OGRE14.4+ migration branch. I wanted to fix our water so it both looks good and simulates good. However, the performance of our buoyancy simulation was so bad my Debug build ran with 3FPS, and even RelWithDebInfo didn't surpass 25 FPS.

I used the profiler to determine the bottleneck. I expected the wave height/velocity calculation to be the culprit, but the problem was the buoyancy forces calculation alone. Each buoycab triangle got split into sub-triangles based on submersion level, and each vertex of each triangle calculated separately. I tried to optimize the calculation but didn't get much of it.

Solution: Instead of calculating buoyancy forces on every tick, it's calculated only each 10th tick (200x per sec), and 2 samples are created simultaneously:
1. forces for current node positions/velocities and wave heights/velocities at present time.
2. forces for projected node positions (based on current velocities) and node velocities (unchanged). Wave heights/velocities are calculated with a time offset (the water algorithm uses a time parameter, so this was straightforward).

On the other ticks, the forces are linearly interpolated between these 2 samples.

I loosely tested on Auriga with Weasel boat and I couldn't see any change in behavior. Currently the sampling interval is hardcoded, I can add a cvar for it if there is demand. I noticed a lot of visual glitches - dripping particles under water, misplaced splashes and most importantly too low water mesh resolution, preventing player from seeing the full complexity of the waves. I'll tackle all these in the OGRE14 branch.

I created a debug view to see how the buoyancy algorithm works and it's beautiful:
image

### DIFF
--- a/source/main/gfx/GfxActor.cpp
+++ b/source/main/gfx/GfxActor.cpp
@@ -1635,13 +1635,25 @@ void RoR::GfxActor::ToggleDebugView()
         m_debug_view = DebugViewType::DEBUGVIEW_NONE;
 }
 
-void RoR::GfxActor::SetDebugView(DebugViewType dv)
+bool RoR::GfxActor::IsDebugViewApplicable(DebugViewType dv)
 {
-    if (dv == DebugViewType::DEBUGVIEW_WHEELS     && m_actor->ar_num_wheels   == 0 ||
+    if (dv < DEBUGVIEWTYPE_FIRST ||
+        dv > DEBUGVIEWTYPE_LAST ||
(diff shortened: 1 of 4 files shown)


CuriousMike56 (approved):
Works well overall, not noticing any changes in water or particle behavior.

The heavy boat crash I reported in https://github.com/RigsOfRods/rigs-of-rods/issues/2814#issuecomment-1379778873 can still happen, though it seems to be less frequent now.

Also the new bouncy debug can't be cycled with CTRL+K:

https://github.com/user-attachments/assets/4950e480-afe8-4bb0-bdd5-4fcfdf6ba0f8

PR #3362, opened 2025-11-09
✨ Handle server rank-only mode

Zentro:
As a part of https://github.com/RigsOfRods/ror-server/pull/170

This will handle a new rank-only message in RoRnet.


ohlidalp:
image
I extended the MP settings dialog so it's clearer what to do... also it now automatically opens there if rejected from server like that.

PR #3367, opened 2025-11-28
Use the Repository UI anytime (single & multi player)

ohlidalp:
Download starts immediatelly after clicking [install] button - the progress window appears.
There is a simple download queue, status is displayed in the window footer. Only one download runs at a time.
Modcache is updated immediatelly after each download finishes.

The RepoUI is now accessible from pause-menu or top menubar->tools menu->"browse repository...", both in singleplayer and multiplayer. Any downloaded mod is instantly usable. Uninstalling mod causes all it's actors to be deleted immediately. Installing missing mods in multiplayer will immediately spawn them.
image

For the record: Updating modcache on background with any actors spawned was technically impossible until February 2024 (https://github.com/RigsOfRods/rigs-of-rods/commit/2e19afef7b0f35fe282198b2af7123db2875256b). That's why the RepoUI had the strange 2-step flow "download immediatelly, refresh cache on window close".

### DIFF
--- a/source/main/gui/panels/GUI_RepositorySelector.cpp
+++ b/source/main/gui/panels/GUI_RepositorySelector.cpp
@@ -1613,9 +1613,14 @@ void RepositorySelector::SetVisible(bool visible)
     {
         this->Refresh();
     }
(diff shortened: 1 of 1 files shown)


ohlidalp:
The RepoUI is now accessible from singleplayer. Any downloaded mod is instantly usable. Uninstalling mod causes all it's actors to be deleted immediately.
image

PR #3368, opened 2025-11-30
AngelScript extensions for diagnosing OGRE

ohlidalp:
I started using scripts a lot when diagnosing OGRE14, so I decided to split out the scriptwork to separate branch. 

Screenshot from current master (OGRE 1.11.6)
image


ohlidalp:
I've also fixed a glitch in the 'script_editor' gadget where the [>>RUN] button got stuck if there were any errors in the script. This was caused by 3d8e44c1222960e194366279a79ff0097e64f35a.

@CuriousMike56 Please test so we can release with a functional script editor gadget. I also need the new OGRE inspector gadget for the OGRE14 port.

Issue #3364, opened 2025-11-17
Feature request: QOL Stuff [Could Use Feedback]


NightFuryVG:
This issue brings up changes regarding Turn Signals and Cruise Control, both hypothetically are small

### Turn signal

With how Rigs Of Rods works when you swap signal, there is no audio feedback. [look below] 

[![03ase3 1](https://img.youtube.com/vi/N_2EC_LfFQU/0.jpg)](https://www.youtube.com/watch?v=N_2EC_LfFQU)

**There are a few ways that I could think of to tackle this** 

1.  You could trigger the start_sound when switching signals
2.  You could make it where people can make a separate sound, though that's kinda hard considering there's only start_sound, sound, and sound_stop
3. this is kind of best of both worlds by default if there's no separate switch sound play start_sound by default but is dependent on No.2

### Cruise Control 

The cruise control is great, but it's missing something, oh you can't resume it so you have to speed right back up and re-engage it 

This change just adds a new key binding that when hit it just resumes cruise control prior to it being canceled when hitting the brake and so on.

I'd figure it could be called something like this in the input map

`TRUCK_CRUISE_CONTROL_RES` with a key binding `CTRL+SHIFT+SPACE`

PR #3372, opened 2025-12-13
✨ Add cruise control resume

devel60:
WIP, needs some help, requested from #3364


ohlidalp:
The code looks allright, try pressing the key for a tad longer - our input processing is flaky and can miss keypresses if your FPS is low for any reason (like running under Debug).

ohlidalp:
I looked again and no, I don't see any obvious problem. Try mapping a simpler key combo via input.map file. Then just keep moving the code around and adding LOG() or similar things until you figure it out. I will test the code tonight.

PR #3383, opened 2025-12-30
🐛 FetchIcon(): Fix crash when country flag doesn't exist

### LINKED ISSUE #3382: Multiplayer: setting fake country in RoR.cfg crashes all clients on server.
so it turns out this is an (accidental) MP exploit ⁠Rigs of Rods⁠: https://discord.com/channels/136544456244461568/274653814467067916/1454961218375778559

It's possible to set an invalid country via RoR.cfg (like Country=eu) and upon joining a server the game attempts to load a non-existent texture which crashes everyone

Image

Zentro:
Fixes #3382

Prevent the crash by checking if the resource (in this case, the icon) exists before attempting to create/retrieve the texture.

### DIFF
--- a/source/main/gui/GUIUtils.cpp
+++ b/source/main/gui/GUIUtils.cpp
@@ -373,12 +373,22 @@ Ogre::TexturePtr RoR::FetchIcon(const char* name)
 {
     try
     {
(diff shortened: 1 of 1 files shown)


CuriousMike56 (approved):
👍 
RoR_2025-12-30_01-30-11

PR #3392, opened 2026-01-09
:video_game: Added new AngelScript bindings for BeamClass, InputEngine and more

EMN-CSharp:
I've added and documented new AngelScript bindings that are focused on manual control of vehicles, allowing modders to implement things such as vehicle AI systems, autopilots or cruise control systems.

### ~~New functions for manual vehicle control~~

~~These functions allow scripts to control vehicles directly, without relying on the input engine.~~

   - ~~`BeamClass.get/setSteeringAngle()`~~
   - ~~`BeamClass.get/setBrakingLevel()`~~
   - ~~`BeamClass.get/setAircraftAileron()`~~
   - ~~`BeamClass.get/setAircraftElevator()`~~
   - ~~`BeamClass.get/setAircraftRudder()`~~
   
Removed in favour of new sim attributes:

   - `ACTORSIMATTR_TRUCK_STEERING`
   - `ACTORSIMATTR_TRUCK_BRAKE`
   - `ACTORSIMATTR_AIRCRAFT_AILERON`
   - `ACTORSIMATTR_AIRCRAFT_ELEVATOR`
   - `ACTORSIMATTR_AIRCRAFT_RUDDER`

### Additional functions for BeamClass

   - `BeamClass.getParkingBrake()`
   - `BeamClass.getCruiseControl()`
   - `BeamClass.getTractionControl()`
   - `BeamClass.getAntiLockBrake()`
   - `BeamClass.cruisecontrolToggle()`

### ~~`get/setControlsLinkedToExternalInput()`~~

~~These functions allow scripts to set which controls are affected by an external input, such as the keyboard; useful for operating the controls without interference from other inputs.~~

Removed in favour of new [simulated values on actors concept](https://github.com/RigsOfRods/rigs-of-rods/pull/3392#issuecomment-3736830400).

### New functions for the autopilot system

   - `AutopilotClass.getOperateControls()`
   - `AutopilotClass.setOperateControls()`

These functions allow custom autopilot scripts to use all the autopilot settings and operate the controls with no interference from the default autopilot.

### New functions for aircraft and boat engines

I've added functions to set the maximum power/thrust of aircraft and boat engines; useful for simulating engine failures or other changes in power under certain conditions.

   - `TurbojetClass.setMaxDryThrust()`
   - `TurbojetClass.setAfterburnerThrust()`
   - `TurbopropClass.setMaxPower()`
   - `ScrewpropClass.setMaxPower()`

### Other changes
  - Added a new binding for `InputEngine.getEventValue()`
  - Added a new example script: `example_manual_vehicle_controls.as`
  - Updated example scripts: `example_aircraft_engines.as`, `example_autopilot.as` and `example_screwprops.as`.
  - Updated AngelScript docs and fixed some issues, such as all the function names being duplicated for BeamClass (caused by **a single missing semicolon** :) )


ohlidalp (approved):
Code & docs are flawless, and thus approved. I have concerns detailed below, but code can be always refactored/refined/reverted later. There are already overlapping or duplicate things in the code and it's no big deal.

I'm not sure about the whole `ActorControlTypeFlags` concept - I think it overlaps with existing input events and mixes things together for no good reason. For example:
* there is `ACT_THROTTLE = 1,           //!< Car/aircraft/boat throttle` 
* at the same time we have separate `EV_AIRPLANE_THROTTLE`, `EV_BOAT_THROTTLE_*`, `EV_TRUCK_ACCELERATE`. I'd like RoR to remain a hard simulator at it's core and do any arcade controlling on top of that. What about controlling individual aeroengines? What about eventual amphibious vehicle? Your system would only cover the arcade case.

There is already (ouch! I just noticed it's not documented.. my bad, but there _is_ an example script though) a mechanism in the input engine that does the same work as your new code, just globally - setting input value manually and blocking user input on that same control:
```
class InputEngineClass {
    void setEventSimulatedValue(inputEvents ev, float value);
    void setEventStatusSupressed(inputEvents ev, bool supressed);
````
An iterative solution to your use case would be to integrate these two directly to `BeamClass` to take effect only on that actor. That would solve all inputs at once, wouldn't need any new enum and wouldn't need a separate 'operate controls' mode of autopilot. If you still need a more arcade API like generic throttle/brake, you can implement it fully in angelscript for your needs.

Also I realized you add and bind a lot of new `[get/set]WhatnotPhysicsParam()` funcs while there is already an universal mechanism to do exactly this without any new bindings: https://developer.rigsofrods.org/d3/d52/_actor_sim_attributes_page.html. Indeed this only covers parameter values, not actual simulation values - but I think those should go through the standard input events above.

@cosmic-vole: This also relates to your work which I'll merge (details in https://github.com/RigsOfRods/rigs-of-rods/pull/3389#issuecomment-3728577670) - you have a custom attuning logic there for the AI races. As in this case, I will also merge that as-is and refine later. I'd just like you to know there are all these scripting options already.

**Comment by cosmic-vole** [comment]
> Also I realized you add and bind a lot of new `[get/set]WhatnotPhysicsParam()` funcs while there is already an universal mechanism to do exactly this without any new bindings: https://developer.rigsofrods.org/d3/d52/_actor_sim_attributes_page.html. Indeed this only covers parameter values, not actual simulation values - but I think those should go through the standard input events above.
> 
> @cosmic-vole: This also relates to your work which I'll merge (details in [#3389 (comment)](https://github.com/RigsOfRods/rigs-of-rods/pull/3389#issuecomment-3728577670)) - you have a custom attuning logic there for the AI races. As in this case, I will also merge that as-is and refine later. I'd just like you to know there are all these scripting options already.

> Those options could work, but I think we should keep those physics methods. The implementation of `VehicleAI` has direct access to the physics properties (for example, [here](https://github.com/RigsOfRods/rigs-of-rods/blob/d38b9b4ea23f01d6c999f777420f847f17d88c39/source/main/gameplay/VehicleAI.cpp#L315)), so why not allow custom vehicle AI scripts to access them as well? **Also, using `ActorSimAttributes` would flood the logfile and abort races with custom AI vehicles.**

Yes I agree with the sentence in bold, as it would currently rule out using ActorSimAttributes to tune the performance of AI competitors for race difficulty and also for implementing power ups.

**Comment by cosmic-vole** [comment]
Your proposed changes all sound fine to me.

With regard to the multiplayer cheating issue, one thing I just thought of is how a lot of racing games now calculate a performance rating for each vehicle, so a server can set an upper limit on that to stop someone using an unfairly fast vehicle. That way players can tune their vehicle to their heart's content, but if they go too far, they may not be allowed on some servers.

The only remaining issue would be what to do if someone tries to change it partway through a session. I guess that could be disallowed as a server option or the performance rating would need to be recalculated to see if the vehicle still qualifies for that server.

We need to bear in mind that a lot of players who would use custom Angelscript to try and cheat could probably just edit their truck file to make their vehicle faster anyway, which to me is half the fun of ROR anyway (the tinkering, not the cheating I mean!).

**Comment by EMN-CSharp** [comment]
I just want to point out that I've made all the changes needed for these features to work, so feel free to review this PR if you want.

**Comment by ohlidalp** [comment]
Again, sorry about the review lag, I got bogged down in OGRE14+

File: doc/angelscript/Script2Game/enums/ActorSimAttr.h [hunk 1]
```diff
@@ -60,6 +60,17 @@ enum ActorSimAttr
     ACTORSIMATTR_ENGTURBO2_ANTILAG_CHANCE, //!<  Param #12 of 'engturbo2'
     ACTORSIMATTR_ENGTURBO2_ANTILAG_MIN_RPM, //!<  Param #13 of 'engturbo2'
     ACTORSIMATTR_ENGTURBO2_ANTILAG_POWER, //!<  Param #14 of 'engturbo2'    
+
+    // Truck controls
```
ohlidalp, line 64: Oh no, no, no. These aren't attributes, they're states. The 'ActorSimAttr' concept was made exclusively for things that don't change dynamically during simulation, meaning args from truck-fileformat or values calculated directly from those at spawn.

doc/angelscript/Script2Game/TurbojetClass.h
```diff
@@ -35,6 +35,18 @@ class TurbojetClass
      * @return The exhaust gas velocity, in metres per second.
      */
     float getExhaustVelocity();
+
+    /**
+     * @brief Sets the maximum dry thrust (no afterburner) of the engine, in kilonewtons.
+     *        The value is limited to the maximum wet thrust.
+     */
+    void setMaxDryThrust(float thrust);
```
ohlidalp, line 43: These, on the flipside, could be ACTORSIMATTRs. Provided you change the function signature to `setSimAttributte(attr, index, value)` to specify which element you want to modify. This way you csn cover nodes/beams/etc.. also.

ohlidalp (changes requested):
Almost great except the misuse of actorsimattrs for direct control of input-based states. Those which have dedicated input event don't need any extra access. For those that don't just add a function.

ohlidalp:
Sorry about the laggy review;I commented last week but didn't notice the comments were stuck in "pending" state.

I suggest you close this PR and create multiple smaller ones, each targeting one area.

EMN-CSharp:
> Sorry about the laggy review;I commented last week but didn't notice the comments were stuck in "pending" state.

No problem!

> I suggest you close this PR and create multiple smaller ones, each targeting one area.

Yeah, that'll be better, I have to admit that this PR got a little too big. Closing this.

PR #3393, opened 2026-01-09
Run rorserver scripts locally (for singleplayer)

ohlidalp:
It's now possible to run server-side scripts (like https://github.com/RigsOfRods/RoRServerScripts) inside the client (without any modifications!). These scripts run in complete isolation from the game (no access to `game` global etc...). The server api (global object `server` and callbacks) is fully supported, except some things that aren't relevant here like `game.getServerIpAddress()`.

**No modifications to rorserver or rornet are needed.**

### Usage
Open ingame console and type `loadserverscript [filename]`. Server scripts only load from dedicated directory 'Documents\My Games\Rigs of Rods\server_scripts' to avoid any mixups with client scripts. Only one can run at a time. Logging is directed to separate file 'logs\RoRServerScript.log'.

### Motivation
I want to extend our race system to support racing against other players (when online) or bots (singleplayer), and, I want one generic code to handle both cases. I figured I can port the rorserver's script subsystem to the client and then create a server-only race supervisor script to cover both scenarios.

### Status
This code was tested to run the rorserver's example script, see https://github.com/RigsOfRods/ror-server/blob/master/contrib/example-script.as - It's richly commented to guide new users through the API.
I was also able to successfully load https://github.com/RigsOfRods/RoRServerScripts - Except for one complaint about `#include "utils"` instead of `#include "utils.as"` in gameScriptWrapper.as
Full documentation of the server API: https://developer.rigsofrods.org/d0/d8c/namespace_script2_server.html
Server callbacks docs: https://developer.rigsofrods.org/df/d6c/group___server2_script.html

Screenshot showing the rorserver example script running in main menu.
image

### DIFF
--- a/doc/angelscript/Script2Game/AngelOgre/AngelOgre_Light.h
+++ b/doc/angelscript/Script2Game/AngelOgre/AngelOgre_Light.h
@@ -9,17 +9,6 @@ namespace AngelOgre { // Dummy namespace, just to distinguish AngelScript from C
  *  @{
  */  
 
-    /// Enumerates the types of light sources available.
-    enum LightTypes
-    {
-        /// Point light sources give off light equally in all directions, so require only position not direction
-        LT_POINT,
-        /// Directional lights simulate parallel light beams from a distant source, hence have direction but no position
-        LT_DIRECTIONAL,
-        /// Spotlights simulate a cone of light from a source so require position and direction, plus extra values for falloff
-        LT_SPOTLIGHT
-    };
-
     /** Representation of a dynamic light source in the scene.
         @remarks
             Lights are added to the scene like any other MovableObject. They contain various 
--- a/doc/angelscript/Script2Game/AngelOgre/AngelOgre_OverlayElement.h
+++ b/doc/angelscript/Script2Game/AngelOgre/AngelOgre_OverlayElement.h
@@ -8,22 +8,6 @@ namespace AngelOgre { // Dummy namespace, just to distinguish AngelScript from C
 /** \addtogroup Script2Game
  *  @{
  */  
- 
-    // Register the GuiMetricsMode enum
-    enum GuiMetricsMode
-    {
-        GMM_PIXELS,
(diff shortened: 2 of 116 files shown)


ohlidalp:
We have AI bots now. They're pretty much the same as old VehicleAI, except they register as ranked players in the locally-running server script.
Screenshot_20260718-110524

The script I'm running here is the example from rorserver repo, see https://github.com/RigsOfRods/ror-server/blob/master/contrib/example-script.as
But I was also able to successfully load https://github.com/RigsOfRods/RoRServerScripts - Except for one complaint about `#include "utils"` instead of `#include "utils.as"` in gameScriptWrapper.as

The last missing piece is to make multiplayer races that can be also played singleplayer with AI bots.

CuriousMike56 (commented):
All existing scripts I've tested so far are working normally👍 

There doesn't appear to be a way to end a server script though as they don't appear in Script Monitor. 

RoRServerScripts runs (after fixing the utils.as typo), however none of its commands (like `!boost`) do anything. Not really a big deal as most of its features are useless in SP but would be fun to have the boost command working at least.

ohlidalp:
@CuriousMike56 Big thanks for testing!

> There doesn't appear to be a way to end a server script though as they don't appear in Script Monitor.

There's `unloadserverscript` in the console - appears in `!help` and all. The idea was that ScriptMonitor would only show local scripts, but I'll add a category for the server script, clarifying that it's isolated from the rest.

> RoRServerScripts runs (after fixing the utils.as typo), however none of its commands (like `!boost`) do anything. Not really a big deal as most of its features are useless in SP but would be fun to have the boost command working at least.

Looks like the commands are somehow not getting through to the server script. Debugging now.
I need this functional since I want to control the race script this way.

UPDATE: oh yes. Fixed.
image

UPDATE2: apparently there's a race condition when unloading the script, you can get a crash if the serverscript's `frameStep()` is just running. Fixing now.
UPDATE3: Fixed

PR #3389, opened 2026-01-03
🎮 ✨ Basic support for racing against the AI vehicles using the Angelscript.

cosmic-vole:
:video_game: :sparkles: These changes enhance the races.as Angelscript and related C++ code so a single player can race against the AI vehicles.

You may not remember but 8 years ago I had a fork with a vast number of experimental code changes allowing fairly functional races against AI that could be started from new game menus. Like this https://www.youtube.com/watch?v=CP41dc8WZQE&pp=ygUQY29zbWljIHZvbGUgcmFjZQ%3D%3D

Unfortunately since then I have been really busy with other things, and your upstream code had changed so much that it became a daunting task to try to merge all of my changes into the new codebase. 

So now I am finally in a position to contribute again, I have set out to break up what I did into individual features, so that I can offer them to you one at a time, if you are interested. I am starting with what I think will be the most useful first.

**Summary of Features in this Change**

- Solo races should work exactly as before, but let me know if I have broken something! I did do a lot of testing.
- Either the player or an AI vehicle can start a new race by driving across the start line. Others can then join by doing the same.
- Once one or more AIs are in the race and pass a checkpoint, your race position is calculated and displayed on the HUD.
- When you finish the race your final position is stated in a message with a congratulation where appropriate.
- You can start another race by crossing the line again, but only if the bots have finished too or have timed out between checkpoints.
- To race against AIs at the moment you need to use the Vehicle AI menu as you normally would, on F1 test track set them to loop at least twice, get your own vehicle near the start line, and either start the AIs then drive across the line, or vice versa.

**Known Issues**

1.  I cannot test the online API submitScore at the end of each lap yet because it makes the game seg fault in Linux for me when the online API is enabled.

2.  When I originally wrote these changes I had to wrap the raceList in my raceDictionary class. This is necessary when we want to build a game UI in C++ to create new races, because we can give them a unique ID (which was a number > 1000 in my other code) rather than having to know exactly how many races are being loaded by a terrain script. From my point of view it would be fairly big amount of work changing this back to not use the dictionary, so if you are interested in future race features from me I would ideally like to keep this in please. 

3. One thing to be aware of is some functions not directly related to the racing logic are still referencing 
~~the underlying~~ raceList directly, e.g. addRaceFromDefinitionFile(). ~~Theoretically that should still work fine alongside the dictionary calls, or it can be changed over.~~ **Edit:** They do need changing over to use raceDict.

[...]

PR #3394, opened 2026-01-11
Racing against AI opponents by Cosmic Vole

ohlidalp:
As promised in https://github.com/RigsOfRods/rigs-of-rods/pull/3389#issuecomment-3728577670, I started merging @cosmic-vole's racing contributions from https://github.com/cosmicvole/rigs-of-rods, master branch.

Because the code is quite old (2017!) and many files got renamed in the meantime (without Git recording the move) or don't exist anymore, I was only able to port one commit so far, and it's not even runnable yet. However, everything is in correct place, so cosmic vole should be able to find their way in the current codebase and potentially take it from here.


cosmic-vole:
This is great, thanks again, but also a lot to get my head around, even though I wrote it! The first thing, before I forget, is that this PR: https://github.com/RigsOfRods/rigs-of-rods/pull/3389 has fixes I wrote recently for the envokeCallback() changes and generally better code for the related collision filters. The old code as above probably breaks other people's use of envokeCallback(), so my new PR needs applying over the top of this. https://github.com/RigsOfRods/rigs-of-rods/pull/3389 has a load of bug fixes and cleanup for races.as as well.

There are other things I probably broke too, like I changed RORBot's pose to fit a small car, though I am not sure if that change is in here or not.

PR #3018, opened 2023-02-12
Mission system (.missionzip)

ohlidalp:
**Info:**
https://forum.rigsofrods.org/threads/mission-system-prototype.3846/

**Features:**
* A new mod type is recognized: *.mission - can add terrain objects, procedural roads and races to existing terrain.
* A new archive type is recognized: *.missionzip
* Mission types: "race" and "stunt", more to come in the future.
* To load a mission, use top menubar "Simulation / Load a mission" (opens the Selector UI) or open in-game console and say 'loadmission _fileName_'
* To unload a mission, use top menubar "Missions" menu and click the red "X" letter.

**New scripts:**
* missions.as - like races.as but for missions. An include file. See comments on the top of the file.
* stunts.as - like races.as but for stunt-missions. An include file. See comments on the top of the file.
* mission_default.as - like terrain_default.as but for missions. Does the basic handling of .mission files.
* road_utils.as - an include file, does parsing procedural roads using GenericDocument tokenizer.

**New script API:**
* `array<BeamClass@> game.getAllTrucks()`
* `float game.getElapsedTime()`
* `array<BeamClass@> BeamClass.getAllLinkedTrucks()`
* `int BeamClass.getInstanceId()`
* Callback `eventCallbackEx()` - same as `eventCallback()` but with extra parameters.
* event types `SE_EVENTBOX_ENTER` and `SE_EVENTBOX_EXIT`
* `terrain.addSurveyMapEntity()`, `terrain.delSurveyMapEntities()`

**Example:**
Download the attached file and place it to your 'mods' directory. Note GitHub doesn't allow '.missionzip' extension, so I just used '.zip' - it works the same.
[simple2circuit.zip](https://github.com/RigsOfRods/rigs-of-rods/files/10715214/simple2circuit.zip)
To test, load the "Simple Test Terrain" (simple2.terrn2), open in-game console and say 'loadmission simple2circuit.mission'.
screenshot_2023-02-12_03-43-43_1

PR #3395, opened 2026-01-11
Race system rework: '.racetrack' files are now a standalone mod.

ohlidalp:
## The new '.racetrack' mod

'.race' files were introduced in 2025: https://github.com/RigsOfRods/rigs-of-rods/commit/0d43021523e61a75c0e9b70354fff21e8f68f35f
After it became clear races should be multiplayer, the name was changed to 'racetrack' because "race" should refer to a single organized event (as in "hey player, want to join this race?").
Otherwise nothing changed - you can still convert existing terrain races using the converter.
Browsing racetracks via Selector UI is currently only possible via script. This will be expanded after integrating https://github.com/RigsOfRods/rigs-of-rods/pull/3389

## races internal reworks (simplified eventboxes)

This is a follow-up to #3081 which declared the 'eventhandler' arg of `game.spawnObject()` deprecated.

The reason to touch on this is preparation for race system extension in #3389 - it requires eventboxes to provide ActorID and `SE_EVENTBOX_ENTER` already does it.

@CuriousMike56 @danmackey @cosmic-vole Please skim through the code and tell me if it makes sense to you. Ofc also test if I didn't break anything please.

To avoid confusion: The script callback `eventCallbackEx(ev, int,int,int,int,  string,string,string,string)` has precedence over the original `eventCallback(ev, int)` so even if the terrain script contains explicit `eventCallback()` (like Auriga does), the `eventCallbackEx()` added dynamically by race system overrides it. This _might_ break some terrains if there's custom stuff in their `eventCallback()` but I think the risk is minor.

### DIFF
--- a/doc/angelscript/Script2Game/GameScriptClass.h
+++ b/doc/angelscript/Script2Game/GameScriptClass.h
@@ -441,7 +441,7 @@ public:
 	 * @param instanceName A unique name for this object (you can choose one, but make sure that you don't use the same name twice)
 	 * @param pos The position where the object should be spawned
 	 * @param rot The rotation in which the object should be spawned
-	 * @param eventhandler (Optional, default empty string) A name of a function that should be called when an eventbox is triggered (see .ODEF file format). <br> If empty, a function `defaultEventHandler()` is executed instead, see /resources/scripts/base.as. <br>OBSOLETE - Use `SE_EVENTBOX_ENTER` instead - it provides the same info plus ActorID.
+	 * @param eventhandler (Optional, default empty string) A name of a function that should be called when an eventbox is triggered (see .ODEF file format). <br> If empty, a function `defaultEventHandler()` is executed instead, see /resources/scripts/base.as. <br>A special value "!supress" prevents running the fallback handler ~ recommended if you rely on `SE_EVENTBOX_ENTER/EXIT` <br>THIS IS DEPRECATED - Use `SE_EVENTBOX_ENTER` instead - it provides the same info plus ActorID.
 	 * @param uniquifyMaterials (Optional, default false) Set this to true if you need to uniquify the materials
 	 */
 	void spawnObject(const string objectName, const string instanceName, vector3 pos, vector3 rot, const string &in eventhandler = "", bool uniquifyMaterials = false);
--- a/doc/angelscript/Script2Game/GameScriptClass.h
+++ b/doc/angelscript/Script2Game/GameScriptClass.h
@@ -441,10 +441,10 @@ public:
 	 * @param instanceName A unique name for this object (you can choose one, but make sure that you don't use the same name twice)
 	 * @param pos The position where the object should be spawned
 	 * @param rot The rotation in which the object should be spawned
-	 * @param eventhandler A name of a function that should be called when an event happens (events, as defined in the object definition file). Enter empty string to ignore events.
-	 * @param uniquifyMaterials Set this to true if you need to uniquify the materials
+	 * @param eventhandler (Optional, default empty string) A name of a function that should be called when an eventbox is triggered (see .ODEF file format). <br> If empty, a function `defaultEventHandler()` is executed instead, see /resources/scripts/base.as. <br>OBSOLETE - Use `SE_EVENTBOX_ENTER` instead - it provides the same info plus ActorID.
+	 * @param uniquifyMaterials (Optional, default false) Set this to true if you need to uniquify the materials
 	 */
-	void spawnObject(const string objectName, const string instanceName, vector3 pos, vector3 rot, const string eventhandler, bool uniquifyMaterials);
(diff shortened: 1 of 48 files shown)


CuriousMike56:
All race scripts I've tested so far work, however this introduces an issue where returning to menu and loading another terrain results in being unable to start races (driving through checkpoint does nothing), even if the previous terrain didn't feature any races.

ohlidalp:
I'm converting this to draft because I'm going to expand the scope - I'll incorporate fixes and refined code from #3018 (2023's Mission system prototype) using the .race fileformat instead of .mission and extending 'races.as' instead of introducing 'missions.as'. I've foreshadowed this in https://github.com/RigsOfRods/rigs-of-rods/pull/3389#issuecomment-3724233693 and later realized it makes sense to do in a single PR.

ohlidalp:
This PR was stale for some time because I worked on OGRE upgrade, but now I'm back to it and the scope changes again.

 The intent is to make multiplayer races that can be also played singleplayer with AI bots. I've already implemented the ability to run multiplayer scripts in singleplayer and use scripts as bots, see: https://github.com/RigsOfRods/rigs-of-rods/pull/3393#issuecomment-5010701541

All changes done here so far will still apply, but aditionally:
* ~the race fileformat must be changed to '.asdata' because rorserver doesn't have GenericDocument tokenizer.~ UPDATE: It turns out .asdata isn't the same thing on server and client, so I ended up adding the GenericDocument tokenizer to rorserver: https://github.com/RigsOfRods/ror-server/pull/183
* the race script must be added to RoRServerScripts using their gameScriptWrapper. This was already done in the past, see https://github.com/RigsOfRods/rigs-of-rods/pull/3393#issuecomment-3735049887

cosmic-vole:
> This PR was stale for some time because I worked on OGRE upgrade, but now I'm back to it and the scope changes again.
> 
> The intent is to make multiplayer races that can be also played singleplayer with AI bots. I've already implemented the ability to run multiplayer scripts in singleplayer and use scripts as bots, see: [#3393 (comment)](https://github.com/RigsOfRods/rigs-of-rods/pull/3393#issuecomment-5010701541)
> 
@ohlidalp I see you are making quite extensive changes to the design compared to the build I worked with. When I have a bit of time I will take a look at how it all works and I can possibly start to look at implementing a RaceAI for the bots based on the algorithms I developed, if you're still interested. I managed to get the bots somewhat competitive at least compared to my rather average skill level as a racer!

ohlidalp:
@cosmic-vole Absolutely interested, and I'm happy to hear from you again, I was worried the long pause had discouraged you. Yes I'm changing a lot, but it should be pretty lateral to what you're doing - bots are just a different way to run the same old VehicleAI. The race script can communicate to them via `server.sendGameCmd()` if needed.

ohlidalp:
Screenshot of the game running updated rorserver example script, see https://github.com/RigsOfRods/ror-server/pull/183
image

ohlidalp:
> ...this introduces an issue where returning to menu and loading another terrain results in being unable to start races...

@CuriousMike56 Fixed. Please test the race scripts again, I've done more changes. If there are no issues, just rebase&merge it please.

CuriousMike56:
@ohlidalp Unfortunately the issue still persists, not able to start any races after returning to menu or reloading from top menu bar.

ohlidalp:
@CuriousMike56 Strange, it works for me:
1. load Bajarama and start race (Desperado)
2. return to menu
3. load Lone Peak - Half Way and start race (desperado)

Tell me your exact scenario pls.

CuriousMike56:
@ohlidalp Was switching between F1 Test Track and Auriga, also happens on the terrains you listed. Even just reloading the first terrain from the top menu causes it.

ohlidalp:
@CuriousMike56 I deleted my post because I accidentally tested with dirty resource dirs.
I can consistently reproduce it - for every odd map load (whether via menu or "reload current terrain"), the race script receives no script events. For some reason the callback function fails to bind. But for every even map load, it works normally again. Looks like some missing cleanup.

Update: I found the problem and it's tough. I've been setting a built-in variable `const int thisScript = 123;` to every script so that everyone can do `game.getScriptDetails(thisScript)` and find out what the script is. The race script had been doing this ever since introducing the 'terrain_project_importer.as' script to handle the race import from script code to *.race files. But as it turns out, `thisScript` doesn't work reliably at the time the constructor of `racesManager` runs - for some reason it may keep the previous value. And I can't introduce additional init to `racesManager` without breaking terrains. Either I'm doing something seriously wrong or I've hit a bug in AngelScript itself. I'll need to do a lot more debugging :/

Update2: gathering feedback at AngelCode forums: https://gamedev.net/forums/topic/720338/

Update3: Still happens with AngelScript 2.38

Update4: Turns out the issue with `thisScript` has been in master all along, it just didn't break races until I stopped using the per-object callback functions specified at `game.spawnObject()`. You can check yourself by testing for the game event responses like resetting truck while in race - they will work the first time, not the other time.

PR #3402, opened 2026-02-13
:bug: Script Monitor GUI: Fixed crash caused by bad string formatting.

EMN-CSharp:
I was able to replicate the crash by spawning two vehicles that loaded dashboards with scripts.


ohlidalp:
This slipped my mind, sorry... I will test and merge ASAP, looks simple enough.

ohlidalp (approved):
Well done; curiously under Windows it doesn't crash but the output is indeed wrong:

Image

I've taken the liberty to edit your commit and reorder the info back to "(actor [%d] '%s')" (ID, name) because the name easily overflows the panel and hides the important fact it's an actor script:

image

PR #3414, opened 2026-03-22
:angel:Script :bug:: Updated scriptstdstring addon to 2.37.0

EMN-CSharp:
This fixes a series of problems that happened just after calling `parseFloat()` in a script, mainly:

- Terrain objects weren't loaded.
- Loss of control for aircraft caused by invalid simulation of airfoils.

`parseFloat()` needs to temporarily change the numeric locale to "C" to ensure decimal strings like "3.14" are always parsed properly. The previous version of `parseFloat()` used the `setlocale()` C function incorrectly (see [commit](https://github.com/anjo76/angelscript/commit/c7c880dbb8c889863af6ea7f5c7eb95cf6aabce0)), causing the locale to be changed to the system locale. This affects the reading of .tobj (terrain objects) and .afl (X-Plane airfoil definition) files if the new locale uses the comma as the decimal separator, since both file formats contain comma-separated values that are read using the locale-sensitive `sscanf()` C function.

The issue of terrain objects not loading was initially reported in [this forum thread](https://forum.rigsofrods.org/threads/problem-with-auriga-proving-grounds.4893/).

### DIFF
--- a/external/angelscript_addons/scriptstdstring/scriptstdstring.cpp
+++ b/external/angelscript_addons/scriptstdstring/scriptstdstring.cpp
@@ -2,7 +2,7 @@
 #include <assert.h> // assert()
 #include <sstream>  // std::stringstream
 #include <string.h> // strstr()
-#include <stdio.h>	// sprintf()
+#include <stdio.h>	// snprintf()
 #include <stdlib.h> // strtod()
 #ifndef __psp2__
 	#include <locale.h> // setlocale()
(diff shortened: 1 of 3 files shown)


ohlidalp (approved):
Excellent catch

ohlidalp:
I'm baffled how exactly the `parseFloat()` bug worked, though. `""` evaluates to true, so the fix only treats null, and setlocale() with null param just reads the locale and doesn't change it, so in that case, the locale would remain "C". The 'fix' makes that explicit but that's it. What am I missing?

EMN-CSharp:
I was just as baffled as you when I discovered the reason for those bugs, but I did a bit more research and found out that everything boils down to the fact that the string returned by setlocale() can be invalidated after subsequent calls to setlocale(). The previous version of scriptstdstring attempted to restore the original locale using a string invalidated after a previous call to setlocale(), setting the locale to the system locale (same behaviour as `setlocale(LC_NUMERIC, "")`). New version fixes that by copying the string returned by setlocale() before calling it again.

PR #3420, opened 2026-04-27
⏱️ Integrated OGRE profiler

ohlidalp:
User-facing features:
- new cvar `diag_profiler_enabled` (bool) - always false on startup, can be activated via console `set diag_profiler_enabled 1`
- new cvar `diag_profiler_rate` (int) - sampling rate of the profiler, defaults to 10 frames. Stored in RoR.cfg.

screenshot_2026-04-27_08-53-50_1

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -6,6 +6,10 @@ from conan.tools.files import copy
 class RoR(ConanFile):
     name = "Rigs of Rods"
     settings = "os", "compiler", "build_type", "arch"
+    default_options = {
+        "ogre3d*:resourcemanager_strict": "off",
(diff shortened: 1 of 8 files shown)


ohlidalp:
For the record (and the laughs), I ran into a visual glitch with the profiler - it showed up on the 'renderdash' dashboard.
On top of that, the renderdash also glitches in the OGRE14 port, traversing scene graph and eating FPS for no reason.
This provoked me to pause and do #3423 which will hopefully be merged in a couple days.

image

ohlidalp:
Test on auriga with PSSM, Hydrax, envmap and mirrors... works OK.
image

PR #3425, opened 2026-05-25
:wrench::bug: Fixed crashes when spawning scripted actors using `game.spawnTruck()` or `game.spawnTruckAI()`

EMN-CSharp:
This fixes a crash that was triggered when calling AS functions `game.spawnTruck()` or `game.spawnTruckAI()`. This only affected actors that had scripts loaded for them (either from dashboard scripts or the `scripts` truckfile section).
The crash could also be easily triggered by loading scripted actors using the Vehicle AI menu (which uses `game.spawnTruckAI()` internally).
While I was at it, I also fixed the `scripts` section ignoring `sectionconfig`.

Here's a test mod I made for triggering the crashes and testing the `sectionconfig`, which is based on the lightweight template truck featured [here](https://docs.rigsofrods.org/vehicle-creation/weight-tuning/).
[RoRTruckScriptsCrashTest.zip](https://github.com/user-attachments/files/28271565/RoRTruckScriptsCrashTest.zip)

### DIFF
--- a/source/main/gui/DashBoardManager.cpp
+++ b/source/main/gui/DashBoardManager.cpp
@@ -509,10 +509,6 @@ DashBoard* DashBoardManager::loadDashBoard(std::string const& filename, BitMask_
         // GameContext.ChangePlayerActor() will change the visibility
         // when needed.
         d->setVisible(false);
-        if (scriptfilename != "")
-        {
-            d->loadScript(scriptfilename, m_actor);
-        }
         m_dashboards.push_back(d);
 
         if (BITMASK_IS_0(flags, LOADDASHBOARD_STACKABLE))
@@ -526,10 +522,6 @@ DashBoard* DashBoardManager::loadDashBoard(std::string const& filename, BitMask_
         d = new DashBoard(this, layoutfname, rttLayer);
         loadedRTTDashboards++;
         d->setVisible(true);
-        if (scriptfilename != "")
-        {
-            d->loadScript(scriptfilename, m_actor);
-        }
         m_dashboards.push_back(d);
         // NOTE: Renderdash RTTs are never stackable
     }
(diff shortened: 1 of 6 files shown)


ohlidalp:
Sorry about the response lag... Nice catch, and good work fixing it.

Honestly I'd rather remove `spawnTruck***()` altogether in favor of `pushMessage(MSG_SIM_SPAWN_ACTOR_REQUESTED, {...})`, then this wouldn't have been an issue, but since it's an old feature it must stay around.

source/main/gui/DashBoardManager.cpp
```diff
@@ -480,11 +480,17 @@ DashBoard* DashBoardManager::loadDashBoard(std::string const& filename, BitMask_
         }
 
         // Load dashboard script
-        Ogre::FileInfoListPtr scriptFile
-            = Ogre::ResourceGroupManager::getSingleton().findResourceFileInfo(entry->resource_group, fmt::format("{}.as", basename));
-        if (scriptFile->size() > 0)
+        // First, we check if the script file actually exists.
+        std::string expected_script_name = fmt::format("{}.as", basename);
+        Ogre::FileInfoListPtr scriptFiles
+            = Ogre::ResourceGroupManager::getSingleton().findResourceFileInfo(entry->resource_group, fmt::format("{}*.as", basename));
```
ohlidalp, line 486: So `findResourceFileInfo(rg, pattern)` given a pattern without wildcard will match the file when working on top of "FileSystem", but not on top of "Zip"? Feels like an OGRE bug. I'll test this in OGRE14 branch and update this comment.

ohlidalp (approved):
Works as advertised. Nice job.

source/main/gui/DashBoardManager.cpp
ohlidalp: Same behavior under OGRE 14.5.2. I'll file a bug with OGRE.

EMN-CSharp: Yep, indeed. Still needed to be changed though, since using resourceExists() is enough for the task. At least the previous code served as a way of finding an issue in OGRE's resource manager :)

ohlidalp: Confirmed OGRE bug: https://github.com/OGRECave/ogre/issues/3648#issuecomment-4741121189

PR #3427, opened 2026-06-18
✨ New scripting feature: Simulated input events for each actor

EMN-CSharp:
This PR brings the changes from #3392 regarding the _simulated input events on actors_ concept.

Now actors have a small, modified subset of the `InputEngine` that consists of the following methods:

   - `void setEventSimulatedValue(ev, val)`: Same functionality as `InputEngine.setEventSimulatedValue()`. As soon as a value is set to an input event on an actor, `InputEngine` values are ignored.
   - `void removeEventSimulatedValue(ev)`: Self-explanatory. Use it when you no longer want to simulate the event value.
   - `float getEventSimulatedValue(ev)`
   - `bool hasEventSimulatedValue(ev)`
   - `void clearEventSimulatedValues()`

I've also added the usual input processing methods:

   - `getEventValue()`: If the actor has a sim value, returns it. Otherwise, it calls `InputEngine.getEventValue()`
   - `getEventBoolValue()`: If the actor has a sim value, returns true if greater than 0.5. Otherwise, it calls `InputEngine.getEventBoolValue()`
   - `getEventBoolValueBounce()`: If the actor has a sim value, behaves the same way as `Actor.getEventBoolValue()`. Otherwise, it calls `InputEngine.getEventBoolValueBounce()`.


source/main/physics/Actor.h
```diff
@@ -169,6 +169,14 @@ class Actor : public RefCountingObject<Actor>
     void              clearForcedCinecam();
     bool              getForcedCinecam(CineCameraID_t& cinecam_id, BitMask_t& flags);
     int               getNumCinecams() { return ar_num_cinecams; }
+    float             getEventValue(int eventID, bool pure = false, InputSourceType valueSource = InputSourceType::IST_ANY);
```
ohlidalp: These deserve their own doxygen-group `/// @name Input engine overrides`, as they're more generic than the rest and often overlap.

source/main/gameplay/Engine.cpp
```diff
@@ -1259,39 +1259,40 @@ float Engine::getPrimeMixture()
...
+        if (m_actor->getEventValue(EV_TRUCK_BRAKE_MODIFIER_50))
         {
             brakeModifier += 0.50f;
         }
         brake *= brakeModifier;
     }
 
+    
     // arcade controls are only working with auto-clutch!
     if (!App::io_arcade_controls->getBool() || (this->getAutoMode() >= SimGearboxMode::MANUAL))
```
ohlidalp: 'Arcade controls' (cvar 'io_arcade_controls') feature must also be overridable per-actor now. I suggest simple `bool ar_arcade_controls`.

source/main/physics/Actor.h
ohlidalp: `ar_hydro_speed_coupling` is toggled by cvar 'io_hydro_coupling' which must now also be overridable per-actor. I suggest:
* rename this to `ar_hydro_speed_coupling_active`
* add `ar_hydro_speed_coupling_enabled` (default false) and update player actor's setting from 'io_hydro_coupling'.

source/main/physics/Actor.h
```diff
@@ -169,6 +169,14 @@ class Actor : public RefCountingObject<Actor>
...
+    bool              getEventBoolValue(int eventID);
+    bool              getEventBoolValueBounce(int eventID, float time = 0.2f);
+    void              clearEventSimulatedValues();
+    bool              hasEventSimulatedValue(int eventID);
+    float             getEventSimulatedValue(int eventID);
+    void              setEventSimulatedValue(int eventID, float value);
+    void              removeEventSimulatedValue(int eventID);
```
ohlidalp: make it `resetEventSimulatedValue()` instead of `removeEventSimulatedValue()` please.

source/main/main.cpp
```diff
@@ -2040,37 +2040,47 @@ int main(int argc, char *argv[])
...
+                            for (ActorPtr actor : App::GetGameContext()->GetActorManager()->GetActors())
+                            {
+                                if (actor->ar_state != ActorState::NETWORKED_OK)
                                 {
-                                    App::GetGameContext()->UpdateCommonInputEvents(dt);
-                                    if (App::GetGameContext()->GetPlayerActor()->ar_state != ActorState::LOCAL_REPLAY)
+                                    // Inputs from InputEngine are disabled for actors when:
+                                    //  - this actor is not the player actor
+                                    //  - camera behaviour is "Free"
+                                    actor->ar_force_simulated_values =
```
ohlidalp: `ar_force_simulated_values` is redundant - Instead create a helper like this:
```
bool Actor::ShouldAllowNonSimulatedInputs()
{
    // Inputs from InputEngine are disabled for actors when:
    //  - this actor is not the player actor
    //  - camera behaviour is "Free"
    // ======================================
    return (ar_instance_id == App::GetGameContext()->GetPlayerActor()->ar_instance_id) ||
         (App::GetCameraManager()->GetCurrentBehavior() != CameraManager::CAMERA_BEHAVIOR_FREE);
}
```
and then replace all `if (!force_foo)` with `if (this->ShouldAllowNonFoo())` 😄

ohlidalp (commented):
Pretty good job, just needs some polishing.

source/main/physics/Actor.h
EMN-CSharp: Done

source/main/physics/Actor.h
EMN-CSharp: Done

source/main/main.cpp
EMN-CSharp: Done

source/main/gameplay/Engine.cpp
EMN-CSharp: Okay, so you mean like adding `ar_arcade_controls` and a function to override the CVar setting?

source/main/physics/Actor.h
EMN-CSharp: Same question as for `ar_arcade_controls`. Also, what is the hydro speed coupling for? I'm not really sure

source/main/gameplay/Engine.cpp
ohlidalp: Yes, plus code in main loop to sync player actor's setting with the cvar

source/main/physics/Actor.h
ohlidalp: Same answer :)

source/main/gameplay/Engine.cpp
EMN-CSharp: Done

source/main/physics/Actor.h
EMN-CSharp: Also done

doc/angelscript/Script2Game/BeamClass.h
```diff
@@ -45,10 +44,10 @@ class BeamClass
     /**
     * Orientation on all axes packed to single quaternion. Use `getYaw()`, `getPitch()` and `getRoll()` for individual rotations in radians.
     */
-    quaternion getOrientation()
+    quaternion getOrientation();
     
     /**
-    * Meters per second.
+    * Gets the vehicle speed, in meters per second.
```
ohlidalp: Just so you know, I hate it when the docstring repeats what the function name already said. It always feels like the writer is wasting my time.

doc/angelscript/Script2Game/BeamClass.h
```diff
@@ -299,28 +298,79 @@ class BeamClass
...
     int getNumCinecams() const;
     
-    
+    //! @}
+
+    /// @name Input engine overrides
+    /// @{
+    /**
+     * Returns the simulated value associated to the given `eventID`.
+     * If `eventID` is not simulated, returns the same value as `InputEngineClass.getEventValue()`.
```
ohlidalp, line 307: I honestly forgot what 'pure' means, add docstring pls.

source/main/physics/Actor.cpp
```diff
@@ -4476,7 +4584,7 @@ Actor::Actor(
     , ar_disable_aerodyn_turbulent_drag(false)
     , ar_engine_hydraulics_ready(true) // !!
     , ar_guisettings_use_engine_max_rpm(false)
-    , ar_hydro_speed_coupling(false)
+    , ar_hydro_speed_coupling_active(false)
```
ohlidalp, line 4598: Don't forget to init `ar_hydro_speed_coupling_active` and `ar_arcade_controls`, too.

ohlidalp (approved):
Looks all right, I will merge it tomorrow when I'm at PC.

source/main/physics/Actor.cpp
EMN-CSharp: Oh I didn't realise, thanks for telling me 👍

doc/angelscript/Script2Game/BeamClass.h
EMN-CSharp: Yeah, I made that change in #3392 just for consistency with other getter functions, without thinking much about it tbh. I'll change it to something more specific

doc/angelscript/Script2Game/BeamClass.h
EMN-CSharp: Had to take a look at InputEngine.cpp to see what it actually does because I wasn't sure either, but I've just added it 👍

ohlidalp (changes requested):
It's not working. I updated the test script to actually send independent inputs to actors, and they're not responding.
EDIT: It's actually working but there is a catch - the actors don't automatically wake up from sleeping state. After I checked "activated vehicles never sleep" in 'top menubar / Simulation' menu, it worked. Calling `setEventSimulatedValue()` with a non-zero argument should wake up a sleeping actor automatically.

Image

ohlidalp (approved):
I added the missing wakeup logic. Please check it's still working, I'm not good at testing after myself :)

I have an idea for your next PR: bind the `ActorState` enum to angelscript and add a getter function for it. Script programmers should know which actors are local and remote, but they must not fiddle with that value. You may add functions `wakeUp()` and `sendToSleep()`.

EMN-CSharp:
> I added the missing wakeup logic. Please check it's still working, I'm not good at testing after myself :)

I've just tested the new script and the latest changes, everything works fine 👍 

> I have an idea for your next PR: bind the ActorState enum to angelscript and add a getter function for it. Script programmers should know which actors are local and remote, but they must not fiddle with that value. You may add functions wakeUp() and sendToSleep().

That sounds great. I still have changes from #3392 that need to be moved to a new PR, which are mainly related to `BeamClass`, so I'll definitely include those functions as well.

PR #3430, opened 2026-06-21
✨ Repo UI: Added contentpack support

Zentro:
Just throwing this out there to see if there's any interest in developing this feature further.

This adds the ability for the Repo UI to install content packs.

### DIFF
--- a/source/main/Application.h
+++ b/source/main/Application.h
@@ -108,6 +108,7 @@ enum MsgType
     MSG_NET_REFRESH_SERVERLIST_FAILURE,    //!< Payload = RoR::CurlFailInfo* (owner)
     MSG_NET_REFRESH_REPOLIST_SUCCESS,      //!< Payload = GUI::ResourcesCollection* (owner)
     MSG_NET_OPEN_RESOURCE_SUCCESS,         //!< Payload = GUI::ResourcesCollection* (owner)
+    MSG_NET_RESOLVE_CONTENTPACK_SIZES_SUCCESS, //!< Payload = std::vector<GUI::ContentPackEntry>* (owner)
     MSG_NET_REFRESH_REPOLIST_FAILURE,      //!< Payload = RoR::CurlFailInfo* (owner)
     MSG_NET_FETCH_AI_PRESETS_SUCCESS,      //!< Description = JSON string
     MSG_NET_DOWNLOAD_REPOIMAGE_SUCCESS,    //!< Payload = RoR::RepoImageDownloadRequest* (owner)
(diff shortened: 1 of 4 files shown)


ohlidalp:
Honestly I think this is backwards - I don't want people to repackage the same stuff over and over in different zips, causing client-side duplicates and stuff to be even more of an issue. Also the new content type enum duplicates the existing category system.

What we could use is a linkage system to automatically download a series of individual repo files from various repo resources. We already have multiple cases of interlinked content: maps use actors, actors use skins and addonparts, both can use assetpacks. Tuneups generated by tuning menu use actors, skins and assetpacks. Yet the repoUi doesn't offer downloading these related items. If we had that, we could also create lightweight contentpacks. 

To implement this, reallistically, staff member could use angelscript to generate and maintain JSON file (one or more) that maps mod filenames to zipnames and zipnames to repo resourceIDs. This could be then delivered via repoAPI

Zentro:
Okay- I think that's good feedback...

rewriting this.

I'm going to change this from accepting a bundled archive (zip of zips, big no no) to accepting a manifest.json file as a means of queuing up files from the repo to install.. any files already installed wouldn't be queued.

```
    // A contentpack manifest lists other repo files to install, e.g:
    // { 
    //     "version": 1,  <- in case of breaking changes
    //     "resources": 
    //     [ 
    //         { 
    //             "resource_id": 1234, <- the resource ID
    //             "file_id": 5678, <- the file from the repo item (where we can have N files that make up a resource)
    //             "filename": "coolcar.zip" <- name of the file to resolve
    //         }
    //     ] 
    // }
```

What I have so far is very basic.. we just queue what we want installed (all cars, terrains, a mix of, etc)- no terrain depends on actor dependency system yet.

Zentro:
If anyone would like to test this..

I added 

https://forum.rigsofrods.org/resources/content-pack-example.1373/

and should be visible in the in-game repository GUI.

Feedback is welcome.. looking to see if multiplayer could also benefit from this.

Wondering if we're fine delivering this as plain JSON files or a as a package.

Zentro:
More progress. 

Using the example on the repo, I recorded this demo.

We resolve the content items in the content pack (up to 50 HTTP requests), discard invalid entries, and then present a dialog allowing the user to choose which items to install. Using the mod cache, we disable entries that are already installed.

I also added a "download queue" window to keep track of what in the queue is being installed.

Still in the process of WIPing up an AS script to generate contentpack manifests with the blocker being the JSON format (bindings with RapidJSON doesn't seem intuitive).

[Screencast_20260725_154904.webm](https://github.com/user-attachments/assets/3e83b2d3-6a51-4685-8586-90c54e529663)

PR #3434, opened 2026-07-07
Added dashboard cinecam visibility setting

CuriousMike56:
Currently the dashboard HUD auto hides while in cinecam view. This behavior can now be changed with "Hide dashboard in cinecam view" in UI settings. When disabled the HUD will always be visible regardless of camera mode.


ohlidalp (approved):
Looks good, I can't test atm though.

PR #3437, opened 2026-07-20
Enet (2026's edition)

ohlidalp:
An experimental RoRnet upgrade to use UDP packets (instead of TCP packets) via popular [ENet](http://enet.bespin.org/) library. 
RoRserver counterpart: https://github.com/RigsOfRods/ror-server/pull/143

Features:
* backward compatibility with older rorservers - first connection is made via TCP (socketw) as usual to check version.
* actor-actor collisions via additional RoRnet stream containing collision forces (not lag-compensated yet).
* character-actor sticky collision: character can walk on cabs of moving networked actor
* extended MP ClientList UI showing net traffic graphs.


Zentro:
Trying to get this to start.. Fedora Linux 44

After compiling, I tried to run rorserver and saw:
```
26-07-2026 01:08:52|t140167781660544|ERROR|Failed to create ENet server host.
```
But this didn't cause the server to sigkill, it kept going. It seems the listener thread started.

I connected anyways with a TCP client, and saw "error getting server version" with this in the logs:
```
26-07-2026 01:15:24|t140695771834048|VERBO|Listener got a new connection
26-07-2026 01:15:24|t140695771834048|ERROR|SWReceiveMessage(): payload too long: 1850896210 b (max. is 8192 b)
26-07-2026 01:15:24|t140695771834048|ERROR|ERROR Listener: receiving first message
26-07-2026 01:15:24|t140695771834048|VERBO|Listener awaiting connections
```

Then, I connected with a valid UDP client and saw "Connecting via UDP" with:
```
26-07-2026 01:31:06|t140574239315648|VERBO|Listener got a new connection
26-07-2026 01:31:06|t140574239315648|VERBO|Listener awaiting connections
```
But enet isn't initialized, so we're just stuck trying to connect until it eventually fails.


Unrelated for RoR server, but the annoying "foreground" feature causes a coredump. I can remove this in a separate PR.

--- edit ---

Inspected errno with:
```
diff --git a/source/server/dispatcher_enet.cpp b/source/server/dispatcher_enet.cpp
index edc037c..03f97eb 100644
--- a/source/server/dispatcher_enet.cpp
+++ b/source/server/dispatcher_enet.cpp
@@ -7,6 +7,9 @@
 
 #include <cassert>
 
+#include <cerrno>
+#include <cstring>
+
 void DispatcherENet::Initialize()
 {
     // Make sure it's not started twice
@@ -41,7 +44,7 @@ void DispatcherENet::Initialize()
                               0      /* assume any amount of outgoing bandwidth */);
     if (m_host == nullptr)
     {
-        Logger::Log(LOG_ERROR, "Failed to create ENet server host.");
+        Logger::Log(LOG_ERROR, "Failed to create ENet server host: %s", std::strerror(errno));
         return;
     }
     else
@@ -201,4 +204,4 @@ void DispatcherENet::QueueMessage(ENetPeer *peer, int type, int source, unsigned
     ENetPacket* packet = enet_packet_create(buffer, msgsize, packet_flags);
     enet_peer_send(peer, 0, packet);
     Messaging::StatsAddOutgoing(msgsize);
-}
\ No newline at end of file
+}
```
which noted:
```
26-07-2026 01:39:06|t140626541614976|ERROR|Failed to create ENet server host: Cannot assign requested address
```

So I set forced my IP to be local (seems trying to fetch the IP from an HTTP call is bad, should probably stop that) and now it works!

image

Users disconnect cleanly.

PR #3442, opened 2026-07-25
Added setting to keep current search term to SelectorUI

### LINKED ISSUE #3411: [Feature request] - Add a checkbox that keeps the current search term in the spawning menu
Whenever you search for a vehicle using the searchbar and you then spawn a vehicle, the spawn menu gets reset.
Would it be possible to add a checkbox that keeps the search term inside of the seach bar even after reopening the menu ?
It would also be interesting to add a clear button next to the checkbox to make it easier to clear the input.

This is an interesting feature to have when testing out trucks and / or scripting for a truck, since it enables to easily spawn repeatedly trucks (or multiple times) for testing purposes.

CuriousMike56:
My attempt at resolving #3411 

Replaces the search hint tooltip with a settings menu containing a new setting to retain the current search term upon reopening the selector.

RoR_2026-07-20_15-27-02
<br>
RoR_2026-07-20_15-27-14

### DIFF
--- a/source/main/Application.cpp
+++ b/source/main/Application.cpp
@@ -285,6 +285,7 @@ CVar* ui_legacy_truck_renderdash;
 CVar* ui_default_boat_dash;
 CVar* ui_always_show_fullsize;
 CVar* ui_dashboard_cinecam;
+CVar* ui_keep_search;
 
 // Instance access
 AppContext*            GetAppContext         () { return &g_app_context; };
--- a/source/main/Application.h
+++ b/source/main/Application.h
@@ -832,6 +832,7 @@ extern CVar* ui_legacy_truck_renderdash;   //!< string; name of the '.dashboard'
 extern CVar* ui_default_boat_dash;         //!< string; name of the '.dashboard' file in modcache.
 extern CVar* ui_always_show_fullsize;
 extern CVar* ui_dashboard_cinecam;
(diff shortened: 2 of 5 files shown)


ohlidalp:
Nice work, I thought about doing the same.

For the overlaps, I suggest remembering the searchstring per-modtype, like last selected category is, and restoring it along with that:
https://github.com/RigsOfRods/rigs-of-rods/blob/5b2efff6b1eb0fc9d90ef51019ea9fcfff9d0450/source/main/gui/panels/GUI_MainSelector.h#L105-L107

```
std::map<LoaderType, std::string> m_kept_searchstring; // stays empty if 'ui_keep_search' is false
```

CuriousMike56:
> For the overlaps, I suggest remembering the searchstring per-modtype, like last selected category is, and restoring it along with that:
> 
> https://github.com/RigsOfRods/rigs-of-rods/blob/5b2efff6b1eb0fc9d90ef51019ea9fcfff9d0450/source/main/gui/panels/GUI_MainSelector.h#L105-L107
> 
> ```
> std::map<LoaderType, std::string> m_kept_searchstring; // stays empty if 'ui_keep_search' is false
> ```

Thanks, works well now.

ohlidalp (approved):
I love the cog icon next to the searchbox!

PR #3444, opened 2026-07-26
👼Script: Added partial support for AngelScript maximum portability mode

EMN-CSharp:
This PR adds all the basic components needed for supporting AngelScript in maximum portability mode, and implements the first generic bindings for `BeamClass`, `GameScriptClass` and `TurbojetClass` (I did the first tests with `Turbojet` since it has just a few methods :) ).


source/main/scripting/ScriptEngine.cpp
```diff
@@ -59,6 +59,8 @@
 
 #include "InputEngine.h"
 
+#define REGISTER_NATIVE_OR_GENERIC(native, generic) if (max_portability) generic(engine); else native(engine);
```
ohlidalp: Generic and native bindings can't be mixed - if actually compiled on a platform not supporting natives, their registrations will probably error out, but either way we shouldn't try to mix them at all.
Please remake this in the angelscript-addons convention, and where generic bindings aren't available, just register nothing (you may log a warning to Angelscript.log).
```
void RegisterActor(asIScriptEngine * engine)
{
    RegisterActorCommon();
	if (strstr(asGetLibraryOptions(), "AS_MAX_PORTABILITY"))
		RegisterActorGeneric(engine);
	else
		RegisterActorNative(engine);
}
```
You can place these funcs as `inline` into 'AngelScriptBindings.h' I suppose.

ohlidalp:
I tested using the cvar and everything seems to be running fine. However, currently there are mixed native/generic bindings which wouldn't work on actual non-native platform, see above.

EMN-CSharp:
Yes, I'm aware of that. I mixed generic and native bindings because I wanted to test with all the bindings available, I'll change it tonight.
Edit: `BeamClass` and `GameScriptClass` reference other bindings, so I had to mix them anyway.

ohlidalp:
I'll take part in this by generifying the ImGui bindings, as without those almost all scripts become useless. Presently almost all the imgui funcs have lambda-wrappers which need to be separated out as common code.

Heads up: while at it, I'll take the liberty to rebase this branch on top of master and force-push it back. The master has had an AngelScript update which involves introducing "autowrapper.h" with 4 args (needed for one of the addons), so I'll resolve that too.

ohlidalp:
As foreshadowed, I rebased the branch on 'master' and started adding more generic bindings. I started by OGRE because imgui relies on them, and also the race system doesn't even load without them. So far I've only committed the basics, I have all the rest locally but I commit in small chunks to test.

PR #3446, opened 2026-07-30
:arrow_up: Updated AngelScript to 2.38

AnotherFoxGuy:


ohlidalp:
This seems to add 'aswrappedcall.h' supporting up to 4 args, but we'll need 8 for  https://github.com/RigsOfRods/rigs-of-rods/pull/3444. I suggest you just delete it, @EMN-CSharp will re-generate the 8-arg one in case something changed on angelscript side.
UPDATE: I see that the 'scripthelper' addon makes use of that.

ohlidalp:
I recklessly rebased&merged without testing on top of current master, and I broke build with #3393 which was merged in between. I fixed it by pushing this directly to master without PR: https://github.com/RigsOfRods/rigs-of-rods/commit/d25d8cbfd3fd9a1270d4d97e2416eaee9b9bbffc

PR #3450, opened 2026-08-08
Added 'extcamera' smoothing/antijitter faux args (`setSimAttribute()` only).

### LINKED ISSUE #3448: Extcamera: Fractionally small node movements makes camera jostle back and forth.
This has been something I've noticed for years now but only now did I think there's probably a way to fix this.

As you probably know, for years, breakable monster trucks have had to use "extcamera" to make sure the camera won't try to follow parts/nodes that have fallen off the truck. "Node 98", typically kept in the middle cradle of the truck, has been used for it. Sometimes though, this node can be a little unstable with all the parts moving around it and the stiffness of it's beams. Yet sometimes, this node will make large movements from point to point like most other nodes as if they're only using a few bits, but since a camera is attached to it, it becomes much more obvious at some angles and especially when you zoom in. I've tried a few things to make it stable but it will always move some, understandably.

Instead, is there any way for "Extcamera" or other camera types to almost ignore the minuscule movements? Like, if a node does not move over 0.01000 or something, than do not translate that motion to the camera? Or something like that. I figure this would be an easier fix than trying to fix the main issue which was the camera following pieces that have fallen off.

Example video:
It's less obvious from a distance, but zoomed in it is pretty clear. After I fix the camera to show how much that isn't moving.
https://youtu.be/o6KvylTD_5A

ohlidalp:
At the moment these new args aren't integrated in the truck fileformat yet, I'll add that if they prove useful.

See new script 'example_truck_extcamera.as' - to use, navigate to top menubar, menu "tools", item "browse gadgets...", select ScriptEditor, in it use top menu "Examples", load the script and hit [>> RUN].

New ActorSimAttributes:
  *  `ACTORSIMATTR_EXTCAMERA_MODE` 'extcamera' arg%1
   * `ACTORSIMATTR_EXTCAMERA_NODE` 'extcamera' arg%2 ~ The node to follow in mode 2 (`RoR::ExtCameraMode::Node`)
   * `ACTORSIMATTR_EXTCAMERA_SMOOTHING` 'extcamera' faux arg (`setSimAttribute()` only) ~ Smoothes the camera target node position (0.0 - 0.9, using 1.0 would stop the camera dead!).
   * `ACTORSIMATTR_EXTCAMERA_ANTIJITTER` 'extcamera' faux arg (`setSimAttribute()` only) ~ Distance in meters the smoothed cam node must move to affect camera. Default value is 0.002 (2 millimeters).

Fixes #3448


Danny-Werewolf:
**I just tested this, and realized that I made a mistake in researching this, though maybe there's a positive yet different outcome?**

I went to test this out of a different map, but realized that the jittering wasn't happening anymore before I had even opened the script editor. So I went to the map I originally wrote from and tested the video on. I immediately had realized that it may be the floating point (or whatever system is used) simply becoming more innacurate, because the bug was still happening even with the script running. You'll have to confirm if the log positions on the original map is reaching the limits of the calculation system.

- Original Map : Position: 6018.215,   13.245, 5992.228,    0.0,   42.7,    0.0
- Different Map: Position:  452.726,    0.925,  555.453,    0.0,  360.0,    0.0

So unfortunately, the script isn't going to fix the original issue, which is fixable in of itself thankfully, just moving the map to not 6,000x5,000 away from center lol. I don't know why the original map is even that far out.

**But...**
I quickly realized that the script accidentally makes a concept I had a while ago seem more plausible?
https://github.com/RigsOfRods/rigs-of-rods/issues/3406
I am specifically talking about the tracking leniency. If you look past the stepped tracking, you can kinda see the idea I was thinking of? Where the camera will lag behind a little. If it was changed to a smoother float and maybe given some options like momentum_rate and random_movement, there wouldn't be much needed to be done for a more camera-esque tracking option. Here is a video I recorded showing that:
https://youtu.be/IxXLdqeWF6I
(The third clip is showing the camera bugging when going over 0.9, all I had to do was exit to menu to fix it.)

ohlidalp:
Thanks for the post - yes it checks out, this is how the float-inaccuracy manifests. I've pinned your post to the related PR. I have an idea how to eliminate it without too much work, I'll try in a couple of days.

Regarding https://github.com/RigsOfRods/rigs-of-rods/issues/3406 I re-watched the video and it's really quite horrible, although you must be zoomed in _an awful lot_ to get such jagged mouse-look. I'll throw a freecam-acceleration slider to the top menubar, so that the freecam movements starts slow and speeds up the longer you move.

CuriousMike56:
> I don't know why the original map is even that far out.

Best guess is that 0.38's Simple Terrain has a spawn point of `6000,0,6000`, likely that track used it as a template.

Danny-Werewolf:
> Best guess is that 0.38's Simple Terrain has a spawn point of `6000,0,6000`, likely that track used it as a template.

Now I'm really curious about this.
I was about to agree, but looking at some older tracks which were only converted to 0.4 with no other edits, the vast majority of 0.38 maps seem to be around `2000,x,2000`. Looking at Sim's early history heavily condensed, this is because that's where the original FlatMap's dirt runway area was back in the 2010's, which I think was the easiest place to repaint for a track paint.
The second picture is of Houston 2014, made in early 2014, by the time Blender was the meta instead of Sketchup yet still in the same place because of copying prior mods to create new ones.
screenshot_2026-08-11_17-31-08_1
screenshot_2026-08-11_17-27-26_1

Also, the same venue (Gaithersburg) that is at `6000,x,6000` for whatever reason does not stay there back in time, as earlier ones are also set at `2000,x,2000`. I'd imagine the Gaithersburg I was on (and one prior from the same creator) is just an error after edits where the model was exported somewhere completely different in blender for whatever model reason, and the spawn point was just moved there since that's the easiest thing to change.

PR #3453, opened 2026-08-15
Added mouse-look smoothing & settings

ohlidalp:
Fixes #3406

There are 2 new cvars:
* io_mouselook_speed ~ Mouse sensitivity, default 8.0 - there is a slider to TopMenubar/Settings
* io_mouselook_smoothing ~ How much to smoothen the movement (0.0 - 0.99), default 0.6 - there is a slider to TopMenubar/Settings

Both are stored in RoR.log.

Additionally, free camera key movement was adjusted a bit (slightly slower translation, much slower rotation).

### DIFF
--- a/source/main/Application.cpp
+++ b/source/main/Application.cpp
@@ -269,6 +269,8 @@ CVar* gfx_reduce_shadows;
 CVar* gfx_enable_rtshaders;
 CVar* gfx_alt_actor_materials;
 CVar* gfx_auto_lod;
+CVar* io_freecam_mouse_speed;
+CVar* io_freecam_mouse_smooth;
 
 // Flexbodies
 CVar* flexbody_defrag_enabled;
--- a/source/main/Application.cpp
+++ b/source/main/Application.cpp
@@ -269,8 +269,8 @@ CVar* gfx_reduce_shadows;
 CVar* gfx_enable_rtshaders;
 CVar* gfx_alt_actor_materials;
 CVar* gfx_auto_lod;
-CVar* io_freecam_mouse_speed;
-CVar* io_freecam_mouse_smooth;
+CVar* io_mouselook_speed;
+CVar* io_mouselook_smoothing;
 
 // Flexbodies
 CVar* flexbody_defrag_enabled;
--- a/source/main/Application.h
+++ b/source/main/Application.h
@@ -815,6 +815,8 @@ extern CVar* gfx_reduce_shadows;
 extern CVar* gfx_enable_rtshaders;
 extern CVar* gfx_alt_actor_materials;
 extern CVar* gfx_auto_lod;
+extern CVar* io_freecam_mouse_speed;
+extern CVar* io_freecam_mouse_smooth;
 
 // Flexbodies
 extern CVar* flexbody_defrag_enabled;
--- a/source/main/Application.h
+++ b/source/main/Application.h
@@ -815,8 +815,8 @@ extern CVar* gfx_reduce_shadows;
 extern CVar* gfx_enable_rtshaders;
 extern CVar* gfx_alt_actor_materials;
 extern CVar* gfx_auto_lod;
-extern CVar* io_freecam_mouse_speed;
-extern CVar* io_freecam_mouse_smooth;
+extern CVar* io_mouselook_speed;
+extern CVar* io_mouselook_smoothing;
 
 // Flexbodies
 extern CVar* flexbody_defrag_enabled;
--- a/source/main/gameplay/SceneMouse.cpp
+++ b/source/main/gameplay/SceneMouse.cpp
@@ -296,7 +296,7 @@ bool SceneMouse::handleMousePressed()
             {
                 player_actor->ar_custom_camera_node = nearest_node_index;
                 player_actor->calculateAveragePosition();
-                App::GetCameraManager()->NotifyContextChange(); // Reset last 'look at' pos
+                App::GetCameraManager()->ResetLookatPos(); // Reset last 'look at' pos
             }
         }
     }
--- a/source/main/gfx/camera/CameraManager.cpp
+++ b/source/main/gfx/camera/CameraManager.cpp
@@ -157,8 +157,7 @@ void CameraManager::UpdateCurrentBehavior(float dt)
     switch(m_current_behavior)
     {
     case CAMERA_BEHAVIOR_CHARACTER: {
-        if (m_mouse_moved)
-            this->CameraBehaviorCharacterMouseMoved();
+        this->CameraBehaviorCharacterMouseMoved(dt); // Always update for smoothing to work.
         if (!App::GetGameContext()->GetPlayerCharacter())
             return;
         m_cam_target_direction = -App::GetGameContext()->GetPlayerCharacter()->getRotation() - Radian(Math::HALF_PI);
@@ -489,7 +488,8 @@ void CameraManager::DeactivateCurrentBehavior()
         return;
 
     case CAMERA_BEHAVIOR_FREE:
-        m_freecam_smooth_mousevec = Ogre::Vector2::ZERO;
+    case CAMERA_BEHAVIOR_CHARACTER:
+        m_mouselook_smooth_vec = Ogre::Vector2::ZERO;
         return;
 
     default:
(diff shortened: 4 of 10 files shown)


Danny-Werewolf:
While testing this and trying to change the cvars, something that's quickly irking me is that if you're in Free Camera and open the console, because you're in free-cam, the mouse is hidden. Ontop of that, the console window blocks keyboard input when hovered over, so you can only untoggle free-cam to start typing (unless you guestimate your mouse to the textbox) by moving the invisible mouse over to toggle fixed cam. Just another reason for these being a Topmenubar slider, I've never encountered this quirk with the console and otherwise.

Also to correct myself; in the video linked in the original issue, we were just standing in first person to try and maintain the immersion of having people on the sidelines, not uncommon for makeshift shows of the time. So this fix doesn't translate to the first-person camera as shown there, but this still does fix part of the issue with the free-cam.

Anyway, about the update itself:
It took a bit and direct comparison to tell as I was looking for something else, but the mouse smoothing is very effective. The screen isn't a pain to look at anymore when surveying a track and it is much smoother in general. With the settings, you can definitely dial in how you want your camera better.

When I first got in, I was confused if it was working at all, the camera was rotating so slow I thought it was just broken. At even 30 FOV it takes me multiple swipes to turn 10 degrees.
I think it'd be good to revert the default speed to what it was before, or at least a good approximate which feels like around 7?  The camera speed being so much

For the Topmenubar sliders, I'd imagine the `gfx_freecam_mouse_speed` could be clamped from 0.1 > 10.0 maybe?
'gfx_freecam_mouse_smooth' seems to permanently break the camera moving at all (WASD too) until I close and re-open the game if I set it to a whole number

When closing and loading the game again, it seems the mouse_speed goes crazy and the camera will freak out for half a second or so. Something similar happens if the mouse_speed is set to a high number like 40.

I admit that I'm a little confused by what `gfx_freecam_mouse_smooth` is supposed to do as a setting? Moving the camera with 0.05 vs 0.10, I don't see any obvious differences. I was originally expecting the camera to kinda lag behind a little, like the opposite of follow-through, so the mouse-moved feel of the camera was less obvious. It's a similar effect that Beam's photo mode or general smoothed free-cam, which comes to mind.

Also, I don't know if it's just me, but it feels like the WASD speed has randomized on me a couple times, even though the "mouse speed" setting doesn't seem to and shouldn't be changing that.

ohlidalp:
> ... irking me is that if you're in Free Camera and open the console, because you're in free-cam, the mouse is hidden. Ontop of that, the console window blocks keyboard input when hovered over, so you can only untoggle free-cam to start typing (unless you guestimate your mouse to the textbox) by moving the invisible mouse over to toggle fixed cam. Just another reason for these being a Topmenubar slider, I've never encountered this quirk with the console and otherwise.

You can't access the top menubar in freecam anyways, you have to toggle out, but sure I'll add sliders.

> Also to correct myself; in the video linked in the original issue, we were just standing in first person to try and maintain the immersion of having people on the sidelines, not uncommon for makeshift shows of the time. So this fix doesn't translate to the first-person camera as shown there, but this still does fix part of the issue with the free-cam.

I'll add smoothing for the firstperson charactercam, too, though that's one extra slider and IIRC mouse is hidden in firstperson too, so I'll probably make it show in character 3rdperson only.
EDIT: I ended up making the smoothing common to both. CVars got renamed.

> When I first got in, I was confused if it was working at all, the camera was rotating so slow I thought it was just broken. At even 30 FOV it takes me multiple swipes to turn 10 degrees. I think it'd be good to revert the default speed to what it was before, or at least a good approximate which feels like around 7? 

The original speed was a bug actually, delta time was factored in twice.
I'll add an adjustable FOV-to-speed ratio for both the 1stperson and freecam (2 more sliders... I think I'll use expanding sections, or maybe tabs)
EDIT: I forgot I set mine to 10.0 ... changing default to 8.

> For the Topmenubar sliders, I'd imagine the `gfx_freecam_mouse_speed` could be clamped from 0.1 > 10.0 maybe? 'gfx_freecam_mouse_smooth' seems to permanently break the camera moving at all (WASD too) until I close and re-open the game if I set it to a whole number

Exactly,

> When closing and loading the game again, it seems the mouse_speed goes crazy and the camera will freak out for half a second or so. Something similar happens if the mouse_speed is set to a high number like 40.

That's a bug, happenned to me once but then I forgot.
EDIT: Uninitialized variable, shame on me. 😑  Fixed.

> I admit that I'm a little confused by what `gfx_freecam_mouse_smooth` is supposed to do as a setting? Moving the camera with 0.05 vs 0.10, I don't see any obvious differences. I was originally expecting the camera to kinda lag behind a little, like the opposite of follow-through, so the mouse-moved feel of the camera was less obvious. It's a similar effect that Beam's photo mode or general smoothed free-cam, which comes to mind.

Hmm, we need some builtin hints for the cvars and limits. The meaning is a ratio between current and previous speed, 0.0 being all current speed (no smoothing), 0.99 being nearly all last speed, using 1.0 would just prevent any movement.

> Also, I don't know if it's just me, but it feels like the WASD speed has randomized on me a couple times, even though the "mouse speed" setting doesn't seem to and shouldn't be changing that.

There may be a glitch there... I'll take a look.

Danny-Werewolf:
**Just trying this. The way the smoothing seems to work at 0.99 now is awesome, and is pretty much what I was hoping for. It really changes how vehicles are spectated and I love it.**

Is there any way to be able to tie the smoothing to the tracking camera as well? Either as a toggle to apply the `mouse_smoothing` slider, or it's own slider. That's actually what I originally had in mind when testing the node-fix stuff from earlier, and with the original issue post.

All I could ask is if it was more effectively different at lower values? I'm sure there's an actual term for it, maybe range? But it's hard to tell much difference from 0.00 to 0.60. I think it'd be perfect if how 0.99 feels currently became like, around 0.60 with it getting even more smooth/latent until a new 0.99.

Now that I think about the WASD speed more, it's so much faster than it was prior. Free-cam without the "Ctrl" or "Shift" modifiers is now about as fast as the camera was with Ctrl + Shift. I'm not sure if this is also to do with the double delta-time error you mentioned, but it's now not possible to get finer and slower movement in free-cam, which is very helpful with both setting your camera somewhere specific and even working on vehicles. However possible, I feel like that should be reverted; I'm not sure when you'd ever need to move the camera as fast as it does with Ctrl + Shift on this build lmao.

ohlidalp:
@Danny-Werewolf I wish everybody was so quick and elaborate with their feedback like you are 🫶 

> Is there any way to be able to tie the smoothing to the tracking camera as well? Either as a toggle to apply the `mouse_smoothing` slider, or it's own slider. That's actually what I originally had in mind when testing the node-fix stuff from earlier, and with the original issue post.

Yes - TBD: 
- [ ] smoothen the fixed-tracking camera "lookat" target coords using the same algorithm, with separate settings.

> ...The way the smoothing seems to work at 0.99 now is awesome...
All I could ask is if it was more effectively different at lower values? I'm sure there's an actual term for it, maybe range? But it's hard to tell much difference from 0.00 to 0.60. I think it'd be perfect if how 0.99 feels currently became like, around 0.60 with it getting even more smooth/latent until a new 0.99.

I have only this one algorithm up my sleeve: 
`    smoothspeed = ((1.0-smoothfactor)*curspeed) + (smoothfactor*prevspeed)` 
IIRC it's called 'first order filtering' because it only goes one step to the past. I really should do my homework on these, but not in this PR.

> Now that I think about the WASD speed more, it's so much faster than it was prior. Free-cam without the "Ctrl" or "Shift" modifiers is now about as fast as the camera was with Ctrl + Shift. I'm not sure if this is also to do with the double delta-time error you mentioned, but it's now not possible to get finer and slower movement in free-cam, which is very helpful with both setting your camera somewhere specific and even working on vehicles. However possible, I feel like that should be reverted; I'm not sure when you'd ever need to move the camera as fast as it does with Ctrl + Shift on this build lmao.

Yes it's the double-deltatime but also a different bug: the original speed constants were mislabeled, ROTATE_SPEED was used for translation and TRANS_SPEED for rotation. While fixing that I also tweaked the values without much care.
- [ ] add cvars for the base ROT/TRANSL freecam speed *and* match them more closely to originals.
- [ ] add cvars for the Ctrl/Alt/Shift factors *and* tweak defaults to mimic original behavior more closely.

I'm not going to add sliders for all these in this PR, instead I suggest you create a script for that using this helper:
```
void RoR::DrawCVarFloatSlider(string cvarName, string label, float v_min, float v_max)
{
    CVarClass@ cvar = console.cVarGet(cvarName);
    float val = cvar.getFloat();
    if (ImGui::SliderFloat(label, val, v_min, v_max, "%.2f"))
    {
        cvar.setVal(val);
    }
}

/* Example:
    DrawCVarFloatSlider("io_mouselook_speed", "Mouse-look speed"), 1.f, 25.f);
*/
```

MarkROR:
Some comments from me:

- You used to be able to adjust the outside view camera by right-clicking anywhere and dragging. This is now not possible. Still possible in interior-cam view. The lack of this makes it impossible to do a "turntable" shot (2nd shown [here](https://youtu.be/XKkFRe_m46w)) 

- The **speed of the free camera** when moving via **WASD** keys and using CTRL, ALT and SHIFT modifier keys **is very fast now**. This makes it impossible to do a slow slide/dolly push-pull/tracking shots e.g. (first shot [here](https://youtu.be/XKkFRe_m46w) free camera approaching with **W** key, slowed down with **ALT** key)

- The top menu camera options appear only either the free camera or fixed camera is activated, I understand the need to keep the top menu not cluttered, but this had me confused for a while (thinking the settings were missing)

Overall, very excited about these new camera smoothing options

ohlidalp:
@MarkROR 
* the orbit camera RMB- adjustment should work the same, if it doesn't I broke it by accident.
* I'll adjust the freecam speed, thanks for feedback.
* making intuitive and non-cluttered UI is hard :)

PR #3351, opened 2025-10-06
:video_game: Dashboard system improvements

EMN-CSharp:
This aims to make the MyGUI dashboard system more flexible by allowing modders to specify their own dashboard inputs in their truckfiles, modify them using AngelScript, and use multiple animations. I've also added support for multiple texture dashboards rendered on different RTTTextures.
### Definition of custom inputs
I've added a new section `customdashboardvalues` to the truckfile parser. Each entry consists of the input name and its data type:
```
customdashboardvalues
custom1 bool
custom2 float
custom3 int
custom4 string
```
### Using custom inputs and animations
These inputs are specified just like built-in inputs. Multiple animations can now be set to a MyGUI widget by adding a number to its associated properties (in order):
```xml
<!-- First animation -->
<UserString key="anim" value="textstring"/>
<UserString key="link" value="Custom_Text"/>
<!-- Second animation -->
<UserString key="anim2" value="translate"/>
<UserString key="vmin2" value="0"/>
<UserString key="vmax2" value="1"/>
<UserString key="min2" value="0"/>
<UserString key="max2" value="100"/>
<UserString key="link2" value="Custom_TextTranslateY"/>
<UserString key="direction2" value="down"/>
```
This change is fully compatible with single animations or links present in current dashboard layouts.

Multiple rotate/scale/translate animations can be set, whereas only one of these can be used at the time: `series`, `textstring`, `textformat`, `lamp`.
### Modifying custom inputs
I've also added AngelScript bindings to the `DashBoardManager`. Here's a short example of how it works:
```angelscript
BeamClass@ truck = null;
DashBoardManagerClass@ dash = null;

// Built-in values
int DashLinkID_airspeed;

// Custom values
int DashLinkID_BankAngle;
int DashLinkID_CustomAirspeed;
int DashLinkID_RadioAltimeter;

// Get the dash manager and link IDs at initialization
void Initialize() {
    @truck = game.getCurrentTruck();
    if (@truck != null) {
        // Returns the DashBoardManager of the actor.
        @dash = truck.getDashboardManager();
        DashLinkID_airspeed = dash.getLinkIDForName("airspeed");
        DashLinkID_CustomAirspeed = dash.getLinkIDForName("Custom_Airspeed");
        DashLinkID_RadioAltimeter = dash.getLinkIDForName("Custom_RadioAltimeter");
    }
}

void frameStep(float dt) {
    if (@dash != null) {
        float airspeed = dash.getNumeric(DashLinkID_airspeed);
[...]

PR #3455, opened 2026-09-05
🐛 More dashboard system-related fixes

EMN-CSharp:
This PR fixes the following issues:

- Dashboard scripts didn't build when calling `getLinkIDForName()` or `setString()` with literal strings as parameters (e.g. `getLinkIDForName("cruisecontrol_active")`). I suspect the issue was triggered by the recent AngelScript update, but in any case scripts now build successfully.
- Scale animations for MyGUI dashboards didn't work properly (I inadvertently broke that in #3351, oops).
- Added missing input source: `cruisecontrol_active`. This input source was referenced by many dashboards (even built-ins) and was never included for some reason.
- Renderdash textures were pointlessly loaded (and not freed) for actors other than trucks.

### DIFF
--- a/source/main/gui/DashBoardManager.cpp
+++ b/source/main/gui/DashBoardManager.cpp
@@ -857,22 +857,20 @@ void DashBoard::update(float dt)
                 float scale = (val - animation.vmin) * (animation.wmax - animation.wmin) / (animation.vmax - animation.vmin) + animation.wmin;
                 if (animation.direction == DIRECTION_UP)
                 {
-                    controls[i].widget->setPosition(controls[i].initialPosition.left, controls[i].initialPosition.top - scale);
+                    finalVerticalTranslation -= scale;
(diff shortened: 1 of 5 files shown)


CuriousMike56 (approved):

ohlidalp:
@EMN-CSharp I have a feeling that https://github.com/RigsOfRods/rigs-of-rods/commit/e81b37c9d881ecd0dfb1fd31df6dfc57b32d05bc will have to be reverted. I've tried to make the game 'smarter' like this before, but it backfired because mods use all features indiscriminately (which is not a bad thing!). Recent example: 5046a03c891ed29d58dcff7bcb6608d42fac1a85.

EMN-CSharp:
So far I haven't had any issues that could have been triggered by that change. Prop anims still work as usual for the mods I tested.
Are there any mods that use many features at the same time? I'd like to test those to make sure.
 
The only elements I moved to change the spawn flow are aircraft engines and screwprops, which only depend on nodes, so that shouldn't be an issue.

